Issue #1, opened 2017-06-06
Ability to query the public CVE-search instance and return vulnerabilities


tdruez:
Given a `package identifier` input, return if there is a known vulnerability for it.

Package identifiers:
- name
- name+version

Behind the scenes:
1. Query the CVE-search API and try to find a match for the package 
2. Return results

Step 1. would after that be replaced by a local query, to the local db, where the aggregated and correlated vulnerability data would be populated from the scrapers, but let's not store anything for now, simply get the data on demand.

pombredanne:
Merged. Closing

PR #2, opened 2017-06-09
#1 Query cve-search' database for package vulnerabilities

kartiksibal:


test_api_data.py
```diff
@@ -0,0 +1,50 @@
...
+'''
+
+def test_output_cve_id():
+	
+	##BUG##
+	api.urlopen = Mock()
+	api.urlopen.return_value = test_data
+	api.output_cve_id()
+
+	assert api.output_cve_id.data["data"]["item"] == "CVE-2007-1004"
```
pombredanne: You want a line return an the end of every files.

test_api_data.py
```diff
@@ -0,0 +1,50 @@
...
+                "cpe:2.3:a:mozilla:firefox:2.0:rc3"
+            ],
+            "vulnerable_configuration_cpe_2_2": [
+                "cpe:/a:mozilla:firefox:2.0:rc3"
+            ]}]}
+'''
+
+def test_output_cve_id():
+	
+	##BUG##
```
pombredanne: What is this comment about? and why an empty line before?

api_data.py
```diff
@@ -0,0 +1,34 @@
...
+import json
+from urllib import urlopen 
+
+def output_cve_id(): 
+	"""Takes as input, a package name, package version. 
+	Queries cve-search' dataset for any reported 
+	vulnerabilities of the requested package. If 
+	vulnerability exists, outputs cve-id(s).
+	"""
+	package_name = raw_input('Enter package name: ')
```
pombredanne: You should rather use args instead and not expect user input at all.

api_data.py
```diff
@@ -0,0 +1,34 @@
...
+		package_ver = raw_input('Enter package version: ')
+		url = 'https://cve.circl.lu/api/search/' + package_name + package_ver
+	
+	else:
+		url = 'https://cve.circl.lu/api/search/' + package_name
+	
+	raw_data = urlopen(url).read()
+	data = json.loads(raw_data)
+
+	if len(data) > 0:
```
pombredanne: you could use instead `if data:` to the same effect

api_data.py
```diff
@@ -0,0 +1,34 @@
...
+		print 'Vulnerabilties Found:\n'
```
pombredanne: You should rather have a function that return data and a main that accepts command line args and deal with printing.

api_data.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+
+import json
+from urllib import urlopen 
+
+def output_cve_id(): 
+	"""Takes as input, a package name, package version.
```
pombredanne: Use `Take` rather than `Takes` also start the docstring on the next line.

test_api_data.py
```diff
@@ -0,0 +1,50 @@
+import api_data as api
+
+from mock import Mock
```
pombredanne: If you use packages for tests, then they should be at least part of some tests requirements file.

kartiksibal: I'll add that to the requirements.txt.

test_api_data.py
kartiksibal: Gotcha

api_data.py
kartiksibal: Gotcha

api_data.py
kartiksibal: Yes, I was developing it into, command line based args and flags. 
Will try to get it done in the next commit.

api_data.py
kartiksibal: You are fast. 😛

api_data.py
kartiksibal: Yes, as stated in the previous comment. Working on that.

api_data.py
```diff
@@ -0,0 +1,35 @@
...
+	if data:
+		print ('Vulnerabilties Found:\n')
+		
+		for item in data['data']:
+			print (item['id'])
+	else:
+		print ('No vulnerabilites found')
+
+if __name__ == '__main__':
+	output_cve_id()
```
kartiksibal: @pombredanne I had added a line after the EOF, I think i deleted it un-consiously. Will be more alert next time

tdruez: @kartiksibal as I mentioned, run `pycodestyle` on your code before committing, it will let you know about those type of issues.

api_data.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+
+"""Py version check"""
+if sys.version_info[0] == 3:
+    from urllib.request import urlopen
+else: 
+	from urllib import urlopen 
+
+def output_cve_id(type=None, name=None, version=None): 
+	"""Takes as input, a package name, package version.
```
kartiksibal: Shoot!

api_data.py
```diff
@@ -1,34 +1,35 @@
...
 	"""
-	package_name = raw_input('Enter package name: ')
-	user_choice = raw_input('Do you have a package version? (Y/N): ')
-	
-	if user_choice == 'Y' or user_choice == 'y':
-		package_ver = raw_input('Enter package version: ')
-		url = 'https://cve.circl.lu/api/search/' + package_name + package_ver
-	
+	if version:
+		url = 'https://cve.circl.lu/api/search/' + name + version
```
kartiksibal: @pombredanne The name is a basic necessity. I was thinking of maybe keeping the whole function in a try-catch block, since if the name is none. We have nothing to search?

api_data.py
```diff
@@ -0,0 +1,35 @@
...
+else: 
+	from urllib import urlopen 
+
+def output_cve_id(type=None, name=None, version=None): 
+	"""Take as input, a package name, package version. 
+	Queries cve-search' dataset for any reported 
+	vulnerabilities of the requested package. If 
+	vulnerability exists, outputs cve-id(s).
+	"""
+	if version:
```
pombredanne: This code will crash. Write some tests... and you will see.

api_data.py
```diff
@@ -0,0 +1,35 @@
...
+		url = ('https://cve.circl.lu/api/search/{}/{}').format(name,version)
+	else:
+		url = ('https://cve.circl.lu/api/search/{}').format(name)
+	
+	raw_data = urlopen(url).read()
+	data = json.loads(raw_data)
+
+	if data:
```
pombredanne: Just return data, do not print it there. Print after calling the function

api_data.py
kartiksibal: @pombredanne As mentioned above, i wasn't sure if going with, keeping the function in a try-except block an effective way to counter this?

api_data.py
```diff
@@ -10,15 +10,15 @@
...
 def output_cve_id(type=None, name=None, version=None): 
-	"""Takes as input, a package name, package version. 
+	"""Take as input, a package name, package version. 
 	Queries cve-search' dataset for any reported 
 	vulnerabilities of the requested package. If 
 	vulnerability exists, outputs cve-id(s).
 	"""
 	if version:
-		url = 'https://cve.circl.lu/api/search/' + name + version
+		url = ('https://cve.circl.lu/api/search/{}/{}').format(name,version)
```
tdruez: Here the parenthesis around the string are not needed: `'{}'.format()`
Since we are using Python 3.6+, feel free to use the format string as in:
`f'https://cve.circl.lu/api/search/{name}/{version}'`

kartiksibal: @tdruez Pro tip! 
Thanks 😄

api_data.py
pombredanne: @kartiksibal nope, you should test the values you get:
Did you get a name and version? --> query for name+version
Did you get only a name? --> query for name
Did you get only a version --> query for version

Eventually you could check if this can be searched more specifically in CPEs?

kartiksibal: @pombredanne A query for just version without a package name isn't possible right?
I'll fix this in the next commit.

api_data.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+
+import json
+import sys
+from urllib.request import urlopen
+
+def output_cve_id(type=None, name=None, version=None):
+    """Take as input, a package name, package version.
```
pombredanne: Minor style point: start the doc on the next line as in
```
"""
Take as input, a package name....
```
Also it is best to start with what you return rather than what you accept

api_data.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
pombredanne: You are missing a license header ;)

api_data.py
```diff
@@ -0,0 +1,32 @@
...
+        url = 'https://cve.circl.lu/api/search/{name}'
+
+    raw_data = urlopen(url).read()
+
+    data = json.loads(raw_data)
+
+    if data:
+        for item in data['data']:
+            return item['id']
+    else:
```
pombredanne: else, return None is not needed, this is the default in Python

api_data.py
```diff
@@ -0,0 +1,32 @@
...
+from urllib.request import urlopen
+
+def output_cve_id(type=None, name=None, version=None):
+    """Take as input, a package name, package version.
+    Queries cve-search' dataset for any reported
+    vulnerabilities of the requested package. If
+    vulnerability exists, outputs cve-id(s).
+    """
+    if not name:
+        return None
```
pombredanne: return is enough, None is the default

test_api_data.py
```diff
@@ -0,0 +1,49 @@
...
+            "vulnerable_configuration": [
+                "cpe:2.3:a:mozilla:firefox:2.0:rc3"
+            ],
+            "vulnerable_configuration_cpe_2_2": [
+                "cpe:/a:mozilla:firefox:2.0:rc3"
+            ]}]}
+'''
+
+def test_output_cve_id():
+    #BUG:
```
pombredanne: What is this comment?

tdruez:
@kartiksibal please fix the remaining issues reported by @pombredanne so we can merge this PR.
Also, make sure to run `pycodestyle` and fix all styling issue as well.

api_data.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+
+import json
+import sys
+from urllib.request import urlopen
+
+def output_cve_id(type=None, name=None, version=None):
+    """Take as input, a package name, package version.
```
kartiksibal: I had a talk with @pombredanne regarding this. 
This is the way he suggested it to be. 
We can discuss this further?  :-)

api_data.py
```diff
@@ -0,0 +1,39 @@
...
+    Queries cve-search' dataset for any reported
+    vulnerabilities of the requested package. If
+    vulnerability exists, outputs cve-id(s).
+    """
+    id = []
```
kartiksibal: Gotcha!

api_data.py
```diff
@@ -0,0 +1,39 @@
...
+    data = json.loads(raw_data)
+
+    if data and name and not version:
+        for item in data['data']:
+            id.append(item['id'])
+
+        return id
+
+    if data and version and name:
+        for item in data:
```
kartiksibal: @tdruez The JSON output changes if a version is added to the URL. Hence, this added part.

api_data.py
```diff
@@ -0,0 +1,39 @@
...
+            id.append(item['id'])          
+
+        return id
+
+if __name__ == '__main__':
```
kartiksibal: @tdruez I agree with you. But don't you think the api code has a use case in it's own too?
If someone doesn't want to use the django app and just get data?

api_data.py
kartiksibal: Done

api_data.py
tdruez: Right now, in that state it has absolutely no value. There's a usecase for using vulnerablecode as a script, but we'll get to it later, with a proper cli.

api_data.py
tdruez: As you said, the variable content changes, but the iteration logic stay the same.
Thus, you need to set the proper variable but use the same loop.

api_data.py
```diff
@@ -0,0 +1,64 @@
...
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import json
+from urllib.request import urlopen
+
+def data_cve_circl(name, version=None):
```
tdruez: > Surround top-level function and class definitions with two blank lines.

http://pep8.org/#blank-lines

api_data.py
```diff
@@ -0,0 +1,64 @@
...
+
+def extract_fields(data, fields_names, version=False):
+    """
+    Extracts requested data fields using data generated by
+    cve-search' api. Takes as input data, fields requested
+    """
+    extracted_data = []
+
+    if version:
+        data = data
```
tdruez: This line has no value. The following is enough:
```
if not version:
    data = data['data']
```

api_data.py
```diff
@@ -0,0 +1,64 @@
...
+    else:
+        data = data['data']
+
+    for item in data:
+        results = {}
```
tdruez: This should be replaced by a dict comprehension:
```
{key: value for (key, value) in iterable}
```

api_data.py
tdruez: Also, I think this section of the code belongs to `data_cve_circl()`.
The extraction logic should not care about the `version`.

api_data.py
kartiksibal: @tdruez 
```
results = {}
    
    {results[name]: item[name] for name in fields_names for item in data}
```

It is kind of erratic but I am getting a KeyError. I am unable to debug the issue, could you guide?

Thanks

api_data.py
kartiksibal: Done!

api_data.py
tdruez: Try this instead:
```
results = {name: item[name] for name in fields_names}
extracted_data.append(results)
```

kartiksibal: @tdruez You missed ``for item in data`` and oddly, it is storing just a single entry? i think we should let go off the dict. comprehension because we need a try get block incase of a ``KeyError`` which tends to pop up if a key is missing somewhere in the data. This shouldn't happen, since the data obviously follows a pattern. But, there is something fishy in the backend.

tdruez: Of course the `for item in data` is still required to build the list. My point is about the dict.
Anyway, keep it the way it works for now, this is not a required change, just nice to have.

api_data.py
```diff
@@ -0,0 +1,67 @@
...
+def data_cve_circl(name, version=None):
+    """
+    Outputs cve-ids, if any, related to a package.
+    Take as input, a package name, type, package version and
+    query cve-search' dataset for any reported vulnerabilities
+    """
+    url = f'https://cve.circl.lu/api/search/{name}'
+
+    if version:
+        url += f'/{version}'
```
pombredanne: This `f` is Python 3-only, right? I guess this is OK as we said we would only support Python3, right?

api_data.py
```diff
@@ -0,0 +1,67 @@
...
+    raw_data = urlopen(url).read()
+    data = json.loads(raw_data)
+
+    return data
+
+
+def extract_fields(data, fields_names, version=False):
+    """
+    Extracts requested data fields using data generated by
+    cve-search' api. Takes as input data, fields requested
```
pombredanne: you are missing a quote.

api_data.py
```diff
@@ -0,0 +1,67 @@
...
+
+    raw_data = urlopen(url).read()
+    data = json.loads(raw_data)
+
+    return data
+
+
+def extract_fields(data, fields_names, version=False):
+    """
+    Extracts requested data fields using data generated by
```
pombredanne: Please use singular order-style `Extract` and not `Extracts` and start your doc with what is returned rather than what it does.

test_api_data.py
pombredanne: I guess you did not remove this?

test_api_data.py
```diff
@@ -0,0 +1,49 @@
+import api_data as api
```
pombredanne: You are also missing a license header

test_api_data.py
kartiksibal: @pombredanne I haven't been able to find a workaround. it'd be great if you took a look.

api_data.py
kartiksibal: Yes, that was an update in 3.6 and yes as you said earlier to support 3.6+. So, this can stay.

test_api_data.py
pombredanne: @kartiksibal the think I am not comfy with is the `#BUG` comment which is meaningless... either you explain in a full narrative comment what the bug is or you remove the  `#BUG` line ;)

Issue #3, opened 2017-06-16
django based app, which combines DRF api & cve-search instance.


kartiksibal:
To Do:

1. Initialise a django based app. ✅ 
2. Using the cve-search api code return CVE-IDS. ✅ 
3. Output the data in JSON format. ❎ 

__EDIT:__

Additions in second point:

1. CVSS Score ❎ 
2. Summary ❎ 
3. URL ❎

kartiksibal:
@tdruez @pombredanne Please guide any changes that might be necessary in the todo, 

thanks

pombredanne:
sure!

pombredanne:
Merged. Closing. This is a basic API that does not use DFR for now,

PR #4, opened 2017-06-18
#3 Use django app to query cve-search and display cve-ids

kartiksibal:


app/vulncode_app/views.py
```diff
@@ -0,0 +1,22 @@
...
+
+def product(request, product):
+    """
+    Queries the cve-search api with just
+    a product name.
+    """
+    data = api.output_cve_id(name=product)
+    return HttpResponse(data)
+
+def product_ver(request, product, ver):
```
tdruez: Why using `version` in the api module and `ver` here?
Always prefer to use easy to read and unambiguous variable name.
Also be consistent with your naming within the codebase,  if the content is the same, use the same name.
`def product_ver(request, product, ver):`
->
`def product_version(request, product, version):`

app/vulncode_app/views.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from django.http import HttpResponse
+from django.shortcuts import render
+
+import vulncode_app.api_data as api
+
+def product(request, product):
```
tdruez: You are using `package` in the api_data module and `product` here, any particular reason?

app/vulncode_app/views.py
kartiksibal: I had primarily done it to sort of create a little differentiation, b/w what's being passed and what it is being checked to. But, your point is more valid. 😄

app/vulncode_app/views.py
kartiksibal: Somewhat in contrast to cve-search. I'd switch 'em to package.

app/vulncode_app/views.py
tdruez: `sort of create a little differentiation` should be avoided, consistency first! :)

kartiksibal: @tdruez Gotcha! 😉

app/vulncode_app/views.py
```diff
@@ -1,26 +1,27 @@
...
     Queries the cve-search api with a package
     name and version.
     """
-    api.output_cve_id(name=name, version=version)
+    raw_data = api.data_cve_circl(name=name, version=version)
+    fields_names = ['id', 'summary', 'cvss']
+    extracted_data = api.extract_fields(raw_data, fields_names, version=True)
 
-    return JsonResponse({'cve_id':api.ids,'summary':api.summary,'cvss':api.cvss})
+    return HttpResponse(json.dumps(extracted_data))
```
kartiksibal: @pombredanne I had changed the data structure of the data, from lists to dictionary and this is sort of what I had implemented.

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,3 @@
+{
```
pombredanne: Do not commit the .cache dir: it should be removed and the git ignored.

app/app/settings.py
```diff
@@ -0,0 +1,120 @@
+"""
```
pombredanne: Missing license header

PR #13, opened 2017-07-05
Creates models for the DB

kartiksibal:


app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
-# Create your models here.
+
+class Vulnerability(models.Model):
+	vulnerability_id = models.AutoField(primary_key=True)
+	summary = models.TextField(max_length=50)
+	cvss = models.FloatField(max_length=50)
+
+
+class VulnerabilityReference(models.Model):
+	vulnerability_id = models.ForeignKey('Vulnerability')
```
pombredanne: when you use a FK here, do not use _id. What you get is an object, not an ID. Also reference the model as a class here, not as a string. This applies to all the models here.

app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
+	source = models.CharField(max_length=50)
+	reference_id = models.CharField(max_length=50)
+	url = models.URLField(max_length=50)
```
pombredanne: When did you last see a URL that's 50 char long?

app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
+
+class Vulnerability(models.Model):
+	vulnerability_id = models.AutoField(primary_key=True)
+	summary = models.TextField(max_length=50)
+	cvss = models.FloatField(max_length=50)
+
+
+class VulnerabilityReference(models.Model):
+	vulnerability_id = models.ForeignKey('Vulnerability')
+	source = models.CharField(max_length=50)
```
pombredanne: Please add a help to each field that documents them.

app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
+	package_id = models.ForeignKey('Package')
+
+
+class ResolvedPackage(models.Model):
+	vulnerability_id = models.ForeignKey('Vulnerability')
+	package_id = models.ForeignKey('Package')
+
+
+class Package(models.Model):
+	package_id = models.AutoField(primary_key=True)
```
pombredanne: This is not needed. We can use the implicit id instead.
We also need some unique together on (platform, name, version)

app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
+from __future__ import unicode_literals
 from django.db import models
 
-# Create your models here.
+
+class Vulnerability(models.Model):
+	vulnerability_id = models.AutoField(primary_key=True)
```
pombredanne: I wonder if we  are going to need a vulnerability ID that's not an auto field I think.

kartiksibal: @pombredanne I had the same dilemma, my thought process was to primarily diff. package id from vulnerability id.

app/vulncode_app/models.py
kartiksibal: @pombredanne I read it here: https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey

Could you please explain a bit, as to why in the documentation it is referenced as a string?

app/vulncode_app/models.py
kartiksibal: @pombredanne I didn't understand your tone here. Too long or short? 😛

app/vulncode_app/models.py
tdruez: https://docs.djangoproject.com/en/1.11/ref/models/fields/#database-representation
> Behind the scenes, Django appends "_id" to the field name to create its database column name. In the above example, the database table for the Car model will have a manufacturer_id column. (You can change this explicitly by specifying db_column) However, your code should never have to deal with the database column name, unless you write custom SQL. You’ll always deal with the field names of your model object.

Simple use: `vulnerability = models.ForeignKey('Vulnerability')`

app/vulncode_app/models.py
tdruez: 50 is way to short, 1024 should be plenty.

app/vulncode_app/models.py
tdruez: @pombredanne is referring to https://docs.djangoproject.com/en/1.11/ref/models/fields/#help-text
This is a great way to document the purpose of each field. This will also be re-used in the admin and API documentation.

app/vulncode_app/models.py
tdruez: You can remove this line. See https://docs.djangoproject.com/en/1.11/ref/models/fields/#primary-key
> If you don’t specify primary_key=True for any field in your model, Django will automatically add an AutoField to hold the primary key

This apply to all models.

app/vulncode_app/models.py
```diff
@@ -1,6 +1,63 @@
...
+class ResolvedPackage(models.Model):
+	vulnerability_id = models.ForeignKey('Vulnerability')
+	package_id = models.ForeignKey('Package')
+
+
+class Package(models.Model):
+	package_id = models.AutoField(primary_key=True)
+	platform = models.CharField(max_length=50)
+	name = models.CharField(max_length=50)
+	version = models.FloatField(max_length=50)
```
tdruez: You should use a CharField for all `version` type field.
FloatField is not suitable for something like "1.2.3"

app/vulncode_app/models.py
kartiksibal: @tdruez I did read that. My primary motive was to diff. vulnerability id from package id. But, I'll change that. 👍

app/vulncode_app/models.py
kartiksibal: @tdruez This is great. 👍

app/vulncode_app/models.py
kartiksibal: @tdruez Right, same thought. 😛

app/vulncode_app/models.py
kartiksibal: @tdruez Pro Tip 👍  😉

Issue #7, opened 2017-06-29
Scrape Ubuntu advisories


pombredanne:
and store in models

Issue #6, opened 2017-06-29
Scrape Debian security advisories


pombredanne:
The goal is to collect and store these in our models

tdruez:
From https://github.com/nexB/vulnerablecode/pull/14#issuecomment-315236716

@kartiksibal the current data structure for the returned CVEs is not easily usable, instead of the current 3 lists of strings, we should return a single list of CVEs dictionaries.

Instead of:
```
(
    ['CVE-2016-5416', 'CVE-2012-6655'],  
    ['389-ds-base', 'accountsservice'],  
    ['not yet assigned', 'not yet assigned'],
)
```

Return:
```
[
    {
        'cve_id': 'CVE-2016-5416', 
        'package_name': '389-ds-base', 
        'vulnerability_status': 'not yet assigned',
    },
    {
        'cve_id': 'CVE-2012-6655', 
        'package_name': 'accountsservice', 
        'vulnerability_status': 'not yet assigned',
    },
]
```
It will be much easier to use that data structure to then create model instances (save the data in the DB).

PR #14, opened 2017-07-07
#6 Debian & #7 Ubuntu Scrapers

kartiksibal:
What it does:

1. Gets the job done.

Things remaining:

1. Add test cases.
2. Remove boilerplate.


test_scrapers.py
```diff
@@ -0,0 +1,71 @@
...
+    extracted_data = ubuntu.extracted_data_ubuntu(test_data)
+
+    assert extracted_data == (['CVE-2002-2439'], ['High'], ['gcc-4.4'])
+
+
+def test_debian_data():
+    # Fix Me: The test data doesn't accurately depict the
+    # actual debian website on which, the code under testing
+    # is written
+    from scraper import debian
```
kartiksibal: @tdruez @pombredanne The main debian website (https://security-tracker.debian.org/tracker/), has several child datasets eg: https://security-tracker.debian.org/tracker/status/release/unstable. 

The scraping code, firsts visits the main website. Extracts the child links, and then visits the child links to extract data. I couldn't form a test data, that could depict this behaviour. 

What do you suggest?

scraper/debian.py
```diff
@@ -0,0 +1,81 @@
...
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import bs4 as bs
+import re
+from urllib.request import urlopen
+
+
+def debian_data():
+    cve_id = []
```
tdruez: Those variables are out of their scope.

scraper/debian.py
```diff
@@ -0,0 +1,81 @@
...
+    # Extract links of child datasets
+    for tag in soup.find_all('a'):
+        href = tag.get('href')
+
+        if re.findall('^/track+.*', href):
+            links.append(href)
+
+    for child_links in range(6):
+        # Extract package info from all the child datasets
+        child_url = urlopen("https://security-tracker.debian.org"
```
tdruez: This part should be refactored into its own function.

test_scrapers.py
tdruez: > firsts visits the main website, extracts the child links, and then visits the child links to extract data.

Then I suggest those 3 logics should be in 3 separate functions, this would make thing much easier to test as well.

tdruez: Why is the import in the function instead of on top of the module?

test_scrapers.py
```diff
@@ -0,0 +1,71 @@
...
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import bs4 as bs
+from mock import Mock
```
tdruez: Remove un-used imports.

scraper/debian.py
```diff
@@ -0,0 +1,94 @@
...
+            id = re.findall('(?<=/tracker/).*', href)
+            cve_id.append(id[0])
+
+        if re.search('/tracker/source-package/(.+)', href):
+            pkg = re.findall('(?<=/tracker/source-package/).*', href)
+            package_name.append(pkg[0])
+
+        # if package name is empty, use the previous package name
+        if href == '/tracker/source-package/':
+            package_name.append(pkg)
```
tdruez: Warning: The `pkg` may be referenced before assignment.
IE: if the previous `if` condition did not check, pkg will not be available.

tdruez:
@kartiksibal the current data structure for the returned CVEs is not easily usable, instead of the current 3 lists of strings, we should return a single list of CVEs dictionaries.

Instead of:
```
(
    ['CVE-2016-5416', 'CVE-2012-6655'],  
    ['389-ds-base', 'accountsservice'],  
    ['not yet assigned', 'not yet assigned'],
)
```

Return:
```
[
    {
        'cve_id': 'CVE-2016-5416', 
        'package_name': '389-ds-base', 
        'vulnerability_status': 'not yet assigned',
    },
    {
        'cve_id': 'CVE-2012-6655', 
        'package_name': 'accountsservice', 
        'vulnerability_status': 'not yet assigned',
    },
]
```
It will be much easier to use that data structure to then create model instances (save the data in the DB).

Issue #16, opened 2017-07-20
Collect Debian vulnerability data


kartiksibal:
Debian security page is at : https://security-tracker.debian.org/tracker/
The main data source is a json file at:  https://security-tracker.debian.org/tracker/data/json

tdruez:
We are moving on using the JSON dataset available at https://security-tracker.debian.org/tracker/data/json to collect the full scope of data needed in vulnerablecode.

@kartiksibal please describe here the pieces that need to be extracted from that dataset and how you plan to structure the results.

kartiksibal:
@tdruez Our aim here is to:

1. extract package name
2. extract the subsequent cve-ids
3. the cve-ids further contain data.

We need references, summary(s) and package details.
Which we require to establish strong vulnerability-package ties.
this data will then be cross-referenced with out cve-search' data.

The same will be done for Ubuntu, the only difference being it's in `XML` and debian is in `JSON`

Now, to scrape JSON we'll need the package names and cve-ids linked to it.

As you can see in the sample data it is `{key: value {key:value}}`

**The problem:**
We need the package names and it's subsequent cve-ids, to be able to extract further information

**The solution:**
All the package names are in this format `{'package name': "CVE-XYZ-123": {`
We can place a capture group in the package name to extract it and then further create a `key:value` pair of `packgaename: cve-id-1, cve-id-2`
then place it in the JSON like:
```
data[packagename][CVE-ID]["releases"]["jessie"])
```

tdruez:
The dataset input is https://security-tracker.debian.org/tracker/data/json
A sample is available at https://pastebin.com/HmagZBm6

The scraper job is to collect and transform that input into the following format:
```
{
  "mimetex": [
      "CVE-1234-1234": {
          "summary": "xyz",
          "repositories": {"jessie": "1.74-1"},
          "urgency": "high",
          "fixed_version":  "1.50-1.1"
        }
    ]
}
```

pombredanne:
@tdruez :+1:

pombredanne:
@kartiksibal I am not even sure  that any transformation is needed in this json?
why not just iterate this and update/insert data in the models?

pombredanne:
@kartiksibal JSON does not need to be "scraped"... this is structured data. You just iterate and transform it as needed.... for instance something such as 
```Python
for package_name, vulnerabilities  in debian.items():
    for vulnerability, details in vulnerabilities.items():
        scope = details.get('scope')
[...]

PR #17, opened 2017-07-20
Refactor scraper logic and datastructure #16

tdruez:
- All data is scraped in 1 loop
- Remove the need for regex
- Return a dict datastructure
- Enhanced the tests using input as collected on the source

Signed-off-by: Thomas Druez <tdruez@nexb.com>


kartiksibal:
@tdruez Looks great! 👍 
We'll get to Debian in a little while? @pombredanne

scraper/debian.py
```diff
@@ -22,59 +22,26 @@
...
 
-    for tag in soup.find_all('td'):
-        if 'medium' in tag or 'low' in tag or 'not yet assigned' in tag:
-            vulnerability_status.append(tag.text)
-        elif tag.find_all('span', {'class': 'red'}) and tag.text == 'high**' or tag.text == 'high':
-            vulnerability_status.append(tag.text)
+    fields_names = ['status', 'urgency', 'fixed_version']
 
-    return cve_id, package_name, vulnerability_status
+    return [{name: version_detail.get(name) for name in fields_names}
```
kartiksibal: @pombredanne @tdruez This will get the three mentioned fields, but obviously not `Package Name` & `Vulnerability`. I can't seem to return data in `{Dicitonary[list{dicitonary}]}` which is `{package_name [cve-ids{cve_id_data}]}`

scraper/debian.py
```diff
@@ -22,59 +22,26 @@
...
-    """
-    soup = bs4.BeautifulSoup(html, 'lxml')
-    tracker_links = soup.findAll('a', href=re.compile('^/track+.*'))
-    return [link.get('href') for link in tracker_links]
-
-
-def extract_cves_from_tracker(html):
+def debian_extract_data():
     """
     Return all CVEs extracted from the given `html` input.
```
pombredanne: The docstring is no longer up to date

scraper/debian.py
```diff
@@ -22,59 +22,26 @@
...
-            cve_id.append(id[0])
-
-        if re.search('/tracker/source-package/(.+)', href):
-            pkg = re.findall('(?<=/tracker/source-package/).*', href)
-            package_name.append(pkg[0])
-
-        # if package name is empty, use the previous package name
-        if href == '/tracker/source-package/':
-            package_name.append(pkg)
+    test_input = urlopen("https://security-tracker.debian.org/tracker/data/json").read()
```
pombredanne: you need to load this with the `json` module...
e.g. once you have read the content, call `debian_vulnerabilities = json.loads(test_input)`

scraper/debian.py
pombredanne: Do not ever use complicated and complex comprehensions with more than a single level.
Use regular loops instead.

pombredanne: @kartiksibal FWIW, @tdruez recommendation to use a dict comprehension was for when you have a more simple, single-level iteration... not multiple level nested loops

scraper/debian.py
```diff
@@ -21,74 +21,40 @@
...
-
-    return cves
+    final_data = []
+
+    for package_name, vulnerabilities in data.items():
+        for vulnerability, details in vulnerabilities.items():
+            for distro, version_detail in details.get('releases', {}).items():
+                if distro == 'jessie':
+                    final_data.append({
+                        "package_name": package_name,
```
pombredanne: Please use single quotes except for doc strings and use all lower case keys.

tests/test_scrapers.py
```diff
@@ -21,105 +21,81 @@
...
+
+                {
+                    'fixed_version': None,
+                    'package_name': 'git-repair',
+                    'status': 'open',
+                    'urgency': 'unimportant',
+                    'vulnerability_id': 'TEMP-0807341-84E914'
+                },
+
+                {
```
tdruez: Collecting this have no value.
If the 'jessie' data is not available we should not create an entry in the output.

PR #25, opened 2017-07-27
Scrapes data from scrapers and dump in the database

kartiksibal:
* Also adds description in Debian's scraper logic.


data_dump.py
```diff
@@ -0,0 +1,62 @@
...
+    for i, v in enumerate(extract_data):
+        vulnerability = Vulnerability(summary=extract_data[i].get('description'))
+        vulnerability_reference = VulnerabilityReference(
+                                              reference_id=extract_data[i].get('vulnerability_id'))
+        package = ImpactedPackage(name=extract_data[i].get('package'),
+                                  version=extract_data[i].get('fixed_version'))
+
+    vulnerability.save()
+    vulnerability_reference.save()
+    package.save()
```
pombredanne: the save() need to be in the loop...
Otherwise you are only saving the latest item. Add some tests that show that

data_dump.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def ubuntu_dump():
+
```
pombredanne: Add some docstring

data_dump.py
```diff
@@ -0,0 +1,62 @@
...
+from vulncode_app.models import ResolvedPackage
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+
+from scraper import debian
+from scraper import ubuntu
+
+
+def debian_dump():
+
```
pombredanne: Add docstring

data_dump.py
```diff
@@ -0,0 +1,62 @@
...
+    json_data = debian.json_data()
+    extracted_data = debian.extract_data(json_data)
+
+    for i, v in enumerate(extract_data):
```
pombredanne: Why do you enumerate here at all? this code makes no sense at all.

pombredanne: You still need to process the comment I amde above ;)

kartiksibal: @pombredanne I have removed enumerate

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    json_data = debian.json_data()
+    extract_data = debian.extract_data(json_data)
+
+    return extract_data
+
+
+def debian_dump(extract_data):
+    """
+    Dump data scraped from Debian' security tracker.
```
pombredanne: Save might be a better name than dump ;)

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,77 @@
...
+    for data in extract_data:
+        vulnerability = Vulnerability(summary=data.get('description'))
+        vulnerability_reference = VulnerabilityReference(reference_id=data.get('vulnerability_id'))
+        package = Package(name=data.get('package_name'), version=data.get('fixed_version'))
+
+        vulnerability.save()
+        vulnerability_reference.save()
+        package.save()
+
+        return vulnerability, vulnerability_reference, package
```
pombredanne: Why do you need to return anything?  Things are saved in the DB.
Also if you return here.... this will return at the first iteration. skipping all the other records.

app/vulncode_app/test_data_dump.py
```diff
@@ -0,0 +1,51 @@
...
+        Scrape data from Debian' main tracker, dump it
+        in the database and verify entries.
+        """
+        with open("tests/test_data/debian.json") as f:
+            test_data = json.loads(f.read())
+
+        extract_data = debian.extract_data(test_data)
+        data_dump = debian_dump(extract_data)
+
+        self.assertEqual(data_dump[0].summary, extract_data[0].get('description'))
```
pombredanne: Your tests should instead do a query in the DB. In particular you should start by asserting that  the number of records you think you inserted based on the contents of `tests/test_data/debian.json` is correct.
Then test the values of all records

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,63 @@
...
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+
+
+class TestVulnerability(TestCase):
+    def test_vulnerability(self):
+        summary_create = Vulnerability.objects.create(summary="Affected package xyz")
+        summary_get = Vulnerability.objects.get(pk=summary_create.pk)
+
+        self.assertEqual(str(summary_create), "Affected package xyz")
```
pombredanne: Do not convert to a string. Instead test the attributes values here and in your other tests

kartiksibal: @pombredanne Do you think we should be keeping `test_models`. Won't, `test_data_dump` suffice?

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    data = ubuntu.scrape_cves()
+    return data
+
+
+def ubuntu_dump():
+    """
+    Dump data scraped from Ubuntu's security tracker.
+    """
+    for data in extract_data:
```
pombredanne: You did not run the tests ;) where could extract_data come from?

app/vulncode_app/test_data_dump.py
```diff
@@ -0,0 +1,59 @@
...
+        Scrape data from Debian' main tracker, dump it
+        in the database and verify entries.
+        """
+        with open("tests/test_data/debian.json") as f:
+            test_data = json.loads(f.read())
+
+        extract_data = debian.extract_data(test_data)
+        data_dump = debian_dump(extract_data)
+
+        for i in range(3):
```
pombredanne: Why limit yourself to 3?

app/vulncode_app/test_data_dump.py
```diff
@@ -0,0 +1,59 @@
...
+            self.assertEqual(3, len(Vulnerability.objects.all()))
+            self.assertEqual(3, len(VulnerabilityReference.objects.all()))
+            self.assertEqual(3, len(Package.objects.all()))
+            self.assertEqual(extract_data[i].get('description'),
+                             Vulnerability.objects.get(pk=i+1).summary)
+            self.assertEqual(extract_data[i].get('vulnerability_id'),
+                             VulnerabilityReference.objects.get(pk=i+1).reference_id)
+            self.assertEqual(extract_data[i].get('package_name'),
+                             Package.objects.get(pk=i+1).name)
```
pombredanne: You should do a query, not hope that the PK will be incremented linearly

app/vulncode_app/test_data_dump.py
pombredanne: But why doing three iterations?

app/vulncode_app/test_data_dump.py
```diff
@@ -0,0 +1,59 @@
...
+        in the database and verify entries.
+        """
+        with open("tests/test_data/debian.json") as f:
+            test_data = json.loads(f.read())
+
+        extract_data = debian.extract_data(test_data)
+        data_dump = debian_dump(extract_data)
+
+        for i in range(3):
+            self.assertEqual(3, len(Vulnerability.objects.all()))
```
pombredanne: Why is the query in the range loop?

pombredanne: Also you should not assert on all() but instead with a filter that select what you expect there.

app/vulncode_app/test_models.py
pombredanne: nope, keep these tests

pombredanne: and do not convert to str

kartiksibal: @pombredanne Will update. 👍

app/vulncode_app/test_data_dump.py
kartiksibal: @pombredanne The query shouldn't be. I'll update that. I'll factor in the second point.

app/vulncode_app/data_dump.py
kartiksibal: @pombredanne It is being passed as a parameter in the method?
https://github.com/nexB/vulnerablecode/blob/data_dump/app/vulncode_app/data_dump.py#L44

app/vulncode_app/test_data_dump.py
kartiksibal: @pombredanne I was iterating over `extract_data` and incrementing `pk`. To match the data.

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,75 @@
...
+from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import ImpactedPackage
+from vulncode_app.models import ResolvedPackage
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+
+from scraper import debian, ubuntu
+
+
+def debian_data():
```
tdruez: This should be in the scraper/debian.py module, for consistency with https://github.com/nexB/vulnerablecode/blob/develop/scraper/ubuntu.py#L50

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,75 @@
...
+    extract_data = debian.extract_data(json_data)
+
+    return extract_data
+
+
+def debian_dump(extract_data):
+    """
+    Save data scraped from Debian' security tracker.
+    """
+    for data in extract_data:
```
tdruez: You could use `ModelClass.objects.create()` instead of `save()` for simplicity.

tdruez: Also, I doubt the following objects creation works, the required fields are not properly set.

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,75 @@
...
+        vulnerability = Vulnerability(summary=data.get('description'))
+        vulnerability_reference = VulnerabilityReference(reference_id=data.get('vulnerability_id'))
+        package = Package(name=data.get('package_name'), version=data.get('fixed_version'))
+
+        vulnerability.save()
+        vulnerability_reference.save()
+        package.save()
+
+
+def ubuntu_data():
```
tdruez: This is un-needed, just call `ubuntu.scrape_cves()` in the `ubuntu_dump()` function.

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TestVulnerabilityReference(TestCase):
+    def test_vulnerability_reference(self):
+        ref_id_create = VulnerabilityReference.objects.create(reference_id="CVE-2017-8564")
+        ref_id_get = VulnerabilityReference.objects.get(pk=ref_id_create.pk)
```
tdruez: Un-used variable.

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,63 @@
...
+
+        self.assertEqual(str(ref_id_create), "CVE-2017-8564")
+
+
+class TestPackage(TestCase):
+    def test_package(self):
+        package_name_create = Package.objects.create(name="Firefox")
+        package_name_get = Package.objects.get(pk=package_name_create.pk)
```
tdruez: Un-used variable.

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,63 @@
...
+
+        self.assertEqual(str(package_name_create), "Firefox")
+
+
+class TestPackageReference(TestCase):
+    def test_package_reference(self):
+        platform_create = PackageReference.objects.create(platform="Maven")
+        platform_get = PackageReference.objects.get(pk=platform_create.pk)
```
tdruez: Un-used variable.

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,63 @@
...
+from vulncode_app.models import ImpactedPackage
+from vulncode_app.models import ResolvedPackage
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+
+
+class TestVulnerability(TestCase):
+    def test_vulnerability(self):
+        summary_create = Vulnerability.objects.create(summary="Affected package xyz")
+        summary_get = Vulnerability.objects.get(pk=summary_create.pk)
```
tdruez: Un-used variable.

kartiksibal: @tdruez `test_models.py` was just a trial method to test models, initially. I haven't updated it. I was under the impression that `test_data_dump` will suffice.

Will update this now. 👍

app/vulncode_app/data_dump.py
kartiksibal: @tdruez Valid point. 👍

app/vulncode_app/data_dump.py
kartiksibal: @tdruez Could you elaborate this?

app/vulncode_app/data_dump.py
kartiksibal: @tdruez gotcha 👍

app/vulncode_app/data_dump.py
pombredanne: I cannot see a parameter in https://github.com/nexB/vulnerablecode/pull/25/files#diff-102b650ccd1cc2cb42f5359f5101e8e1R66
`def ubuntu_dump():` .... there is no args: hence I can only conclude that the code has  never be run or tested ;)

app/vulncode_app/test_data_dump.py
pombredanne: This is not a robust approach for testing. Instead use a query, this is an iterable that you can then handle. Never expect that the pk have a specific values.

kartiksibal: @pombredanne Got it. 👍

app/vulncode_app/test_data_dump.py
kartiksibal: Updating. 👍

app/vulncode_app/test_data_dump.py
kartiksibal: @pombredanne Since, we are testing the number of `entries`. Isn't `.all()` a requirement here?

app/vulncode_app/data_dump.py
tdruez: For example: `Vulnerability.objects.create(summary="Affected package xyz")`
-> `django.db.utils.IntegrityError: NOT NULL constraint failed: vulncode_app_vulnerability.cvss`
Since the cvss is not a nullable field, and since you do not set a value for it, your save() cannot work

kartiksibal: @tdruez Yes, I get your point. I'll add the updated models along the next commit.

app/vulncode_app/test_data_dump.py
pombredanne: My main point was not to make that test three times in a range loop.
As far all is concerned, that is fine to count, but then use `count()` not `all()`

kartiksibal: Gotcha 👍 
I have addressed your first point.

app/vulncode_app/models.py
```diff
@@ -26,15 +26,21 @@
 
 
 class Vulnerability(models.Model):
-    summary = models.TextField(max_length=50, help_text="Summary of the vulnerability")
-    cvss = models.FloatField(max_length=50, help_text="CVSS Score")
+    summary = models.TextField(max_length=50, help_text="Summary of the vulnerability", null=True)
```
tdruez: What's the use case for creating Vulnerability without a summary?

app/vulncode_app/models.py
```diff
@@ -26,15 +26,21 @@
...
+    def __str__(self):
+        return self.summary
 
 
 class VulnerabilityReference(models.Model):
-    vulnerability = models.ForeignKey('Vulnerability')
-    source = models.CharField(max_length=50, help_text="Source's name eg:NVD")
-    reference_id = models.CharField(max_length=50, help_text="Reference ID, eg:CVE-ID")
-    url = models.URLField(max_length=1024, help_text="URL of Vulnerability data")
+    vulnerability = models.ForeignKey('Vulnerability', null=True)
```
tdruez: What's the use case for creating VulnerabilityReference  without a Vulnerability reference?

app/vulncode_app/models.py
```diff
@@ -48,18 +54,24 @@ class ResolvedPackage(models.Model):
 
 
 class Package(models.Model):
-    platform = models.CharField(max_length=50, help_text="Package platform eg:maven")
-    name = models.CharField(max_length=50, help_text="Package name")
-    version = models.CharField(max_length=50, help_text="Pacakge version")
+    platform = models.CharField(max_length=50, help_text="Package platform eg:maven", null=True)
+    name = models.CharField(max_length=50, help_text="Package name", null=True)
```
tdruez: What's the use case for creating Package without a name?

app/vulncode_app/models.py
```diff
@@ -48,18 +54,24 @@ class ResolvedPackage(models.Model):
...
+    version = models.CharField(max_length=50, help_text="Pacakge version", null=True)
+
+    def __str__(self):
+        return self.name
 
 
 class PackageReference(models.Model):
-    package = models.ForeignKey('Package')
+    package = models.ForeignKey('Package', null=True)
```
tdruez: What's the use case for creating PackageReference without a Package reference

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,57 @@
...
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from vulncode_app.models import Vulnerability
```
tdruez: Remove un-used imports

app/vulncode_app/data_dump.py
```diff
@@ -0,0 +1,57 @@
...
+from scraper import debian, ubuntu
+
+
+def debian_dump(extract_data):
+    """
+    Save data scraped from Debian' security tracker.
+    """
+    for data in extract_data:
+        vulnerability = Vulnerability.objects.create(summary=data.get('description', ''))
+        vulnerability_ref = VulnerabilityReference.objects.create(
```
tdruez: No need to create a variable if you do not need to use the object.

app/vulncode_app/models.py
```diff
@@ -26,15 +26,18 @@
...
 
 
 class VulnerabilityReference(models.Model):
     vulnerability = models.ForeignKey('Vulnerability')
-    source = models.CharField(max_length=50, help_text="Source's name eg:NVD")
-    reference_id = models.CharField(max_length=50, help_text="Reference ID, eg:CVE-ID")
-    url = models.URLField(max_length=1024, help_text="URL of Vulnerability data")
+    source = models.CharField(max_length=50, help_text="Source's name eg:NVD", blank=True)
+    reference_id = models.CharField(max_length=50, help_text="Reference ID, eg:CVE-ID", blank=True)
+    url = models.URLField(max_length=1024, help_text="URL of Vulnerability data", null=True)
```
tdruez: `URLField` is a subclass of `CharField`, use `blank` instead.

app/vulncode_app/models.py
```diff
@@ -48,18 +51,26 @@ class ResolvedPackage(models.Model):
 
 
 class Package(models.Model):
-    platform = models.CharField(max_length=50, help_text="Package platform eg:maven")
-    name = models.CharField(max_length=50, help_text="Package name")
-    version = models.CharField(max_length=50, help_text="Pacakge version")
+    platform = models.CharField(max_length=50, help_text="Package platform eg:maven", blank=True)
+    name = models.CharField(max_length=50, help_text="Package name", blank=True)
+    version = models.CharField(max_length=50, help_text="Pacakge version", blank=True)
```
tdruez: Typo in "Pacakge"

app/vulncode_app/models.py
```diff
@@ -26,15 +26,18 @@
 
 
 class Vulnerability(models.Model):
-    summary = models.TextField(max_length=50, help_text="Summary of the vulnerability")
-    cvss = models.FloatField(max_length=50, help_text="CVSS Score")
+    summary = models.TextField(max_length=50, help_text="Summary of the vulnerability", blank=True)
```
tdruez: Why using a TextField and not a CharField here?

app/vulncode_app/test_models.py
```diff
@@ -0,0 +1,85 @@
...
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from django.test import TestCase
+
+from vulncode_app.models import Vulnerability
```
tdruez: Again, un-used import and un-used variable in that file.

.travis.yml
```diff
@@ -7,9 +7,13 @@ install:
 
 before_script:
   - pycodestyle --exclude=migrations,settings.py,lib,tests --max-line-length=100 .
+  - cd app/
+  - python3 manage.py makemigrations
```
pombredanne: Remove this. Migrations should be committed

.travis.yml
```diff
@@ -7,9 +7,13 @@ install:
...
+  - python3 manage.py migrate
 
 script:
   - python3.6 -m pytest -v tests/
+  - python3.6 manage.py test
```
pombredanne: We should use one or the other for running tests.... do not run tests twice

README.md
```diff
@@ -28,18 +29,28 @@ pycodestyle --exclude=migrations,settings.py,lib,tests --max-line-length=100 .
 cd app/
 python3.6 -m pytest -v tests/
```
pombredanne: Here too, we should only use one way to run tests. Not two

app/vulncode_app/models.py
```diff
@@ -30,9 +29,11 @@ class Vulnerability(models.Model):
     A software vulnerability with minimal information.
     Identifiers are stored as VulnerabilityReference.
     """
-    summary = models.TextField(max_length=1024,
-                               help_text="Summary of the vulnerability")
-    cvss = models.FloatField(help_text="CVSS Score")
+    summary = models.CharField(max_length=50, help_text="Summary of the vulnerability", blank=True)
```
pombredanne: Please use single quotes for all strings  (unless this is a docstring or  it contains a single quote itself)
This applies here and below

app/vulncode_app/test_data_dump.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.assertTrue(Vulnerability.objects.get(
+                        summary='Multiple stack-based buffer overflows in mimetex.cgi in mimeTeX'))
+
+        self.assertTrue(Vulnerability.objects.get(
+                        summary='Multiple unspecified vulnerabilities in mimeTeX.'))
+
+        self.assertTrue(VulnerabilityReference.objects.get(reference_id="CVE-2009-2458"))
+
+        self.assertTrue(VulnerabilityReference.objects.get(reference_id="CVE-2009-2459"))
```
pombredanne: Single quotes for strings throughout.

app/vulncode_app/urls.py
```diff
@@ -22,10 +22,11 @@
...
 
 from django.conf.urls import url
+
 from . import views
 
-urlpatterns = [
 
+urlpatterns = [
     url(r'(?P<name>[a-z]+)/(?P<version>[0-9]+)', views.package_version, name="package_version"),
     url(r'^(?P<name>[a-z]+)', views.package, name="package"),
```
pombredanne: Single quotes

Issue #8, opened 2017-06-29
Create JSON API that accepts a scancode package name and version


pombredanne:
.... and return essential data about the found vulnerability if any with
- vulnerabilities references
- packages references

pombredanne:
I suggest this instead:
```
[
{ name: foo,
  version: 12.3,
  vulnerabilities: [
     {summary:...
      references: [
          {url: ......, other fields},
          {url: ......, other fields}
       ]
    },
    { other vulnerability}
  ]
 },
 { other package name and version...}
]
```

pombredanne:
and for each package version that is impacted if we have have resolved versions, these should be listed too as a list

tdruez:
From DRF documentation on relations http://www.django-rest-framework.org/api-guide/relations/#nested-relationships

Let me give you an example to serialize the related `PackageReference` objects within the `PackageSerializer`

```
from rest_framework import serializers

from vulncode_app.models import Package
from vulncode_app.models import PackageReference


class PackageReferenceSerializer(serializers.ModelSerializer):
    class Meta:
        model = PackageReference
        fields = ('repository', 'platform', 'name', 'version')


class PackageSerializer(serializers.ModelSerializer):
    package_references = PackageReferenceSerializer(source='packagereference_set', many=True)

    class Meta:
        model = Package
        fields = ('name', 'version', 'package_references')
```

```
>>> import json
>>> package = Package.objects.latest('pk')
>>> print(json.dumps(PackageSerializer(package).data, indent=2))
{
  "name": "gcc-4.6",
  "version": "",
  "package_references": [
    {
      "repository": "repo",
      "platform": "platform",
      "name": "name",
      "version": "version"
    },
    {
      "repository": "repo",
      "platform": "platform",
      "name": "name",
      "version": "version"
    }
  ]
}
```

@kartiksibal You can apply the same serialization logic to include `vulnerabilities` on the `PackageSerializer`.

kartiksibal:
@pombredanne @tdruez 
```
[
    {
        "package": {
            "name": "mimetex",
            "version": "1.50-1.1"
        },
        "vulnerability": {
            "summary": "Multiple stack-based buffer overflows in mimetex.cgi in mimeTeX, when downloaded before 20090713, allow remote attackers to execute arbitrary code via a TeX file with long (1) picture, (2) circle, or (3) input tags.",
            "reference": [
                {
                    "reference_id": "CVE-2009-1382"
                }
            ]
        }
    },
    {
[...]

PR #26, opened 2017-08-14
JSON API #8

kartiksibal:
Signed-off-by: Kartik sibal <kartiksibal@gmail.com>


app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+
+from vulncode_app.models import Vulnerability
+from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import Package
+
+
+class VulnerabilitySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Vulnerability
+        exclude = ('id', 'cvss')
```
tdruez: Avoid using `exclude`, use `fields` instead, see https://docs.djangoproject.com/en/1.11/topics/forms/modelforms/#selecting-the-fields-to-use

> It is strongly recommended that you explicitly set all fields that should be edited in the form using the fields attribute. Failure to do so can easily lead to security problems when a form unexpectedly allows a user to set certain fields, especially when new fields are added to a model.

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class VulnerabilityReferenceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = VulnerabilityReference
+        exclude = ('id', 'source', 'url', 'vulnerability')
```
tdruez: Similar comment, use `fields`.

app/vulncode_app/data_dump.py
```diff
@@ -23,6 +23,7 @@
 
 from vulncode_app.models import Vulnerability
 from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import ImpactedPackage
```
pombredanne: Imports should be sorted

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,65 @@
...
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from rest_framework import serializers
+
+from vulncode_app.models import Vulnerability
```
pombredanne: Import should be sorted

app/vulncode_app/urls.py
```diff
@@ -22,11 +22,21 @@
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 from django.conf.urls import url
-
-from . import views
+from rest_framework.urlpatterns import format_suffix_patterns
```
pombredanne: Empty line needed after this

app/vulncode_app/urls.py
```diff
@@ -22,11 +22,21 @@
...
+from vulncode_app import views
 
 
 urlpatterns = [
-    url(r'(?P<name>[a-z]+)/(?P<version>[0-9]+)', views.package_version, name='package_version'),
-    url(r'^(?P<name>[a-z]+)', views.package, name='package'),
+    url(r'^cve-search/(?P<name>[a-z]+)/(?P<version>[0-9]+)',
```
pombredanne: Why do you call this `cve-search`?

app/vulncode_app/views.py
```diff
@@ -24,10 +24,45 @@
...
+    def get(self, request, package_name):
+        vulnerability = []
+        extract = []
+        summary = []
+        reference_id = []
+        version = []
+
+        pk = Package.objects.filter(name=package_name)
+
+        for i, v in enumerate(pk):
```
tdruez: This is the wrong approach and should be handled by the proper serializers.

pombredanne: @kartiksibal rather than `for i, v in enumerate(pk):` just iterate on the packages:
```
packages = Package.objects.filter(name=package_name)
for package in packages:
  ....
```

app/vulncode_app/views.py
```diff
@@ -24,10 +24,42 @@
...
 
+class VulnerabilityData(APIView):
+    def get(self, request, package_name):
+        vulnerability = []
+        extract = []
+        data = []
+
+        pk = Package.objects.filter(name=package_name)
+
+        for i, v in enumerate(pk):
```
pombredanne: you are still enumerating pk. Instead loop on packages:
`for package in Package.objects.filter(name=package_name): ....`

app/vulncode_app/models.py
```diff
@@ -50,13 +50,16 @@ class VulnerabilityReference(models.Model):
...
+        return self.source
+
 
 class ImpactedPackage(models.Model):
     """
     Relates a vulnerability to package(s) impacted by it.
     """
     vulnerability = models.ForeignKey('Vulnerability')
-    package = models.ForeignKey('Package')
+    package_fk = models.ForeignKey('Package')
```
kartiksibal: @pombredanne It was necessary to rename this since the reverse query name for `Package.impacted_package` clashes with `ImpactedPackage.package`

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+class ImpactedPackageSerializer(serializers.ModelSerializer):
+    vulnerability = VulnerabilitySerializer()
+
+    class Meta:
+        model = ImpactedPackage
+        fields = ('vulnerability',)
+
+
+class PackageSerializer(serializers.ModelSerializer):
+    vulnerability = ImpactedPackageSerializer(source='impactedpackage_set', many=True)
```
pombredanne: This should be vulnerabilities, plural IMHO

app/vulncode_app/data_dump.py
```diff
@@ -21,9 +21,10 @@
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
-from vulncode_app.models import Vulnerability
-from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import ImpactedPackage
```
pombredanne: Sorts import

app/vulncode_app/data_dump.py
```diff
@@ -38,10 +39,18 @@ def debian_dump(extract_data):
...
             reference_id=data.get('vulnerability_id', ''),
         )
-        Package.objects.create(
+        package = Package.objects.create(
             name=data.get('package_name', ''),
             version=data.get('fixed_version', ''),
         )
+        impacted_package = ImpactedPackage.objects.create(
+            vulnerability=vulnerability,
+            package_fk=package
```
pombredanne: Do not call this package_fk in the model, but package.

app/vulncode_app/data_dump.py
```diff
@@ -38,10 +39,18 @@ def debian_dump(extract_data):
...
+        )
+        package.impacted_package.add(
```
pombredanne: Why do you add this? You created it alright in line 46?

app/vulncode_app/data_dump.py
```diff
@@ -56,6 +65,10 @@ def ubuntu_dump(html):
             vulnerability=vulnerability,
             reference_id=data.get('cve_id'),
         )
-        Package.objects.create(
+        package = Package.objects.create(
             name=data.get('package_name'),
         )
+        ImpactedPackage.objects.create(
+            vulnerability=vulnerability,
+            package_fk=package
```
pombredanne: Same as above: use package, not package_fk as an attribute name in the model

app/vulncode_app/models.py
```diff
@@ -50,13 +50,16 @@ class VulnerabilityReference(models.Model):
     class Meta:
         unique_together = ('vulnerability', 'source', 'reference_id')
```
pombredanne: Would not the URL also be part of the unique together? just a question, not sure. We can quite likely have multiple URLs for a source that refer to the same reference_id.

app/vulncode_app/models.py
pombredanne: What do you mean by clash? can you be more specific? at the db level? in code? where? provide an example with some stack/error trace.
In any case, package_fk is not a good name.
If a name change is really need to avoid some collision, then this would be the ManyToMany field name in Package that would need to be changed instead. 

Or maybe what you are really returning in you API as a primary model is not a Package... but an ImpactedPackage instead which would be actually more natural, what do you think?

app/vulncode_app/models.py
```diff
@@ -73,6 +76,7 @@ class Package(models.Model):
     A software package with minimal identifying information.
     Other identifiers are stored as PackageReference.
     """
+    impacted_package = models.ManyToManyField('ImpactedPackage')
```
pombredanne: Is this really a ManyToMany here? Can an ImpactedPackage be resolved by many packages?
if so the name should be plural `impacted_packages`... since this will resolve to many not one. Hence you likely name collision source?

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import Vulnerability
+
+
+class PackageReferenceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = PackageReference
+        fields = ('name', 'version', 'repository', 'paltform', )
```
pombredanne: paltform has a typo in the field name

app/vulncode_app/test_api.py
```diff
@@ -0,0 +1,156 @@
...
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import json
```
pombredanne: Sort your imports

app/vulncode_app/views.py
```diff
@@ -24,10 +24,23 @@
 import json
 
 from django.http import HttpResponse
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework import status
 
+from vulncode_app.serializers import PackageSerializer
```
pombredanne: Sort your imports

app/vulncode_app/views.py
```diff
@@ -24,10 +24,23 @@
...
+from vulncode_app.models import Package
 from vulncode_app import api_data
 
 
+class VulnerabilityData(APIView):
+    def get(self, request, package_name):
+        pk = Package.objects.filter(name=package_name)
+        response = PackageSerializer(pk, many=True).data
+
```
pombredanne: Empty line not needed here

app/vulncode_app/views.py
```diff
@@ -24,10 +24,23 @@
...
+        return Response(response)
+
```
pombredanne: Only one line between class methods

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from rest_framework import serializers
+
+from vulncode_app.models import ImpactedPackage
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+from vulncode_app.models import VulnerabilityReference
+from vulncode_app.models import Vulnerability
```
pombredanne: Sort your imports, this comes before VulnerabilityReference

app/vulncode_app/models.py
```diff
@@ -73,6 +76,7 @@ class Package(models.Model):
...
     platform = models.CharField(max_length=50, help_text='Package platform eg:maven', blank=True)
     name = models.CharField(max_length=50, help_text='Package name', blank=True)
     version = models.CharField(max_length=50, help_text='Package version', blank=True)
```
pombredanne: in the line below 
```
def __str__(self):
    return self.name
```
we likely need more than just name ... , but this is minor

app/vulncode_app/models.py
kartiksibal: Error Trace:
screen shot 2017-08-22 at 8 25 41 pm

kartiksibal: And I'll take another look into it, before lodging a ticket.

app/vulncode_app/test_api.py
```diff
@@ -0,0 +1,156 @@
...
+
+from django.test import TestCase
+from rest_framework.response import Response
+from collections import OrderedDict
```
pombredanne: e.g. this is stdlib and should not be in this block

app/vulncode_app/models.py
pombredanne: You also likely need a unique together on these two fields to avoid dupe entries

pombredanne:
Also you likely need to run makemigrations and commit the migrations for the model updates

app/vulncode_app/models.py
kartiksibal: done

app/vulncode_app/data_dump.py
kartiksibal: done

app/vulncode_app/data_dump.py
kartiksibal: done

app/vulncode_app/models.py
kartiksibal: I think more than one URL can point at a particular `reference_id`

app/vulncode_app/serializers.py
kartiksibal: Done

app/vulncode_app/serializers.py
kartiksibal: done

app/vulncode_app/views.py
kartiksibal: done

app/vulncode_app/views.py
kartiksibal: done

app/vulncode_app/views.py
kartiksibal: Not class methods.

app/vulncode_app/test_api.py
kartiksibal: done

app/vulncode_app/test_api.py
kartiksibal: done

app/vulncode_app/models.py
pombredanne: ok, then you need to update the unique together then?

pombredanne: And you may want to check other models that may need updated unique together too

app/vulncode_app/data_dump.py
```diff
@@ -38,10 +39,14 @@ def debian_dump(extract_data):
             vulnerability=vulnerability,
             reference_id=data.get('vulnerability_id', ''),
         )
-        Package.objects.create(
+        package = Package.objects.create(
             name=data.get('package_name', ''),
             version=data.get('fixed_version', ''),
```
pombredanne: Based on our chat, does it ever make sense to create package without name and version or both empty values?
Also what about the platform field? Same for Ubuntu

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+from vulncode_app.models import Package
+from vulncode_app.models import PackageReference
+from vulncode_app.models import Vulnerability
+from vulncode_app.models import VulnerabilityReference
+
+
+class PackageReferenceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = PackageReference
+        fields = ('name', 'version', 'repository', 'platform', )
```
pombredanne: Can you keep the field order from the model?

app/vulncode_app/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class VulnerabilityReferenceSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = VulnerabilityReference
+        fields = ('reference_id', 'source', 'url')
```
pombredanne: Here and elsewhere, please keep the model fields order

app/vulncode_app/test_api.py
```diff
@@ -0,0 +1,165 @@
...
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import json
+
```
pombredanne: no empty lines between a block of imports from the same kind (here stdlib)

app/vulncode_app/test_api.py
```diff
@@ -0,0 +1,165 @@
...
+                       ('version', '1.50-1.1'),
+                       ('vulnerabilities',
+                        [OrderedDict([
+                            ('vulnerability',
+                             OrderedDict([
+                               (
+                                 'summary',
+                                 'Multiple stack-based buffer overflows in mimetex.cgi in mimeTeX'
+                               ),
+                               ('reference',
```
pombredanne: Should not this plural "references"?

kartiksibal: atm, every `reference` contains a single `reference id`. So, I think not.

app/vulncode_app/test_api.py
kartiksibal: done

app/vulncode_app/serializers.py
kartiksibal: done

app/vulncode_app/serializers.py
kartiksibal: done

app/vulncode_app/test_api.py
pombredanne: My point is that there are multiple references, hence the list should be plural

tdruez:
@kartiksibal The API code is in much better shape now, thanks to using proper Serializers ;)

I've made some cleanup in https://github.com/nexB/vulnerablecode/pull/26/commits/8158a84e0b000cdc14625261f13a757496e2884e

Some refinements are still needed from you before the merge:

1. Add some documentation on how to use the API in the README.md. You can create a new section at the end of the file.
2. All the test_* files should be located in a `tests` directory
3. `data` is a terrible name for the API URL, please use a proper name. Since this API endpoint returns Packages, I would suggest `/api/packages/`

Avoid using `data` for naming things in general.

@pombredanne could you review the shape of the returned package data structure?

app/vulncode_app/data_dump.py
kartiksibal: Refer: https://github.com/nexB/vulnerablecode/issues/29

app/vulncode_app/tests/test_api.py
```diff
@@ -0,0 +1,158 @@
...
+class TestResponse(TestCase):
+    def test_debian_response(self):
+        with open('tests/test_data/debian.json') as f:
+            test_data = json.loads(f.read())
+
+        extract_data = debian.extract_vulnerabilities(test_data)
+        debian_dump(extract_data)
+        response = self.client.get('/vulncode_app/api/mimetex', format='json')
+
+        expected = [{
```
pombredanne: I would expect the returned payload to have some "header" data (e.g. some tool version, what was the query made, number of results returned, ... And to have the list of packages returned under the packages: element

app/vulncode_app/tests/test_api.py
```diff
@@ -0,0 +1,158 @@
...
+            "name": "mimetex",
```
pombredanne: May be we are missing other Package fields at the same level as this such as the "platform" ?

app/vulncode_app/tests/test_api.py
```diff
@@ -0,0 +1,158 @@
...
+            "version": "1.50-1.1",
+            "vulnerabilities": [{
+                "vulnerability": {
+                    "summary": "Multiple stack-based buffer overflows in mimetex.cgi in mimeTeX",
+                    "reference": [{
```
pombredanne: Should this be references (plural) and not reference, since this is also a list of mappings?

app/vulncode_app/tests/test_api.py
```diff
@@ -0,0 +1,158 @@
...
+
+        extract_data = debian.extract_vulnerabilities(test_data)
+        debian_dump(extract_data)
+        response = self.client.get('/vulncode_app/api/mimetex', format='json')
+
+        expected = [{
+            "name": "mimetex",
+            "version": "1.50-1.1",
+            "vulnerabilities": [{
+                "vulnerability": {
```
pombredanne: IMHO we do not need vulnerabilities/vulnerability
Only vulnerabilities as a list of mappings, each being a vulnerability since we only list these there. No need for an extra level of nesting and a "vulnerability" nested mapping?

app/vulncode_app/tests/test_api.py
tdruez: Create as https://github.com/nexB/vulnerablecode/issues/30

vulnerabilities/urls.py
```diff
@@ -23,10 +23,20 @@
...
 urlpatterns = [
-    url(r'(?P<name>[a-z]+)/(?P<version>[0-9]+)', views.package_version, name='package_version'),
-    url(r'^(?P<name>[a-z]+)', views.package, name='package'),
+    url(r'^cve-search/(?P<name>[a-z]+)/(?P<version>[0-9]+)',
+        views.package_version,
+        name='package_version'),
+    url(r'^cve-search/(?P<name>[a-z]+)',
+        views.package,
+        name='package'),
+    url(r'^api/(?P<package_name>[a-z]+)',
```
tdruez: This pattern `[a-z]+` is not good enough to catch real package names, for example: `ruby1.9.1` and `gcc-4.6`. Use `(?P<package_name>.+)` instead.

app/vulncode_app/tests/test_api.py
kartiksibal: done

app/vulncode_app/tests/test_api.py
kartiksibal: done

PR #45, opened 2019-04-07
Add admin

singh1114:
- Add admin to see the packages being added to the database.


vulnerabilities/models.py
```diff
@@ -29,7 +29,7 @@ class Vulnerability(models.Model):
     A software vulnerability with minimal information.
     Identifiers are stored as VulnerabilityReference.
     """
-    summary = models.CharField(max_length=100, help_text='Summary of the vulnerability', blank=True)
+    summary = models.TextField(help_text='Summary of the vulnerability', blank=True)
```
singh1114: Some of the summaries in the Debian package were more than 2000 characters.

vulnerablecode/urls.py
```diff
@@ -33,6 +34,7 @@
 
 
 urlpatterns = [
-    url(r'^vulnerabilities/', include('vulnerabilities.urls')),
-    url(r'^api/', include(api_router.urls)),
+    re_path(r'^admin/', admin.site.urls),
+    re_path(r'^vulnerabilities/', include('vulnerabilities.urls')),
+    re_path(r'^api/', include(api_router.urls))
```
singh1114: Recommended way in newer versions.

PR #48, opened 2019-07-11
add ResolvedReference

puneeth2001:
Added the Resolved reference which was specified.


puneeth2001:
@lohani2280 Can you check if the changes are correct, In the mean time I will fix the builds

Issue #28, opened 2017-08-21
Evolve logics to deal with duplicates in data.


kartiksibal:
Example of:` GET /vulncode_app/data/prototypejs`
```
{
    "name": "prototypejs",
    "version": [
        "0",
        "1.6.0.2-1"
    ],
    "vulnerabilities": [
        {
            "summary": "Unspecified vulnerability in Prototype JavaScript framework (prototypejs) before 1.6.0.2 allows attackers to make \"cross-site ajax requests\" via unknown vectors.",
            "reference_id": "CVE-2008-7220"
        },
        {
            "summary": "Unspecified vulnerability in Prototype JavaScript framework (prototypejs) before 1.6.0.2 allows attackers to make \"cross-site ajax requests\" via unknown vectors.",
            "reference_id": "CVE-2008-7220"
        },
        {
            "summary": "Unspecified vulnerability in Prototype JavaScript framework (prototypejs) before 1.6.0.2 allows attackers to make \"cross-site ajax requests\" via unknown vectors.",
            "reference_id": "CVE-2008-7220"
        },
        {
            "summary": "The Prototype (prototypejs) framework before 1.5.1 RC3 exchanges data using JavaScript Object Notation (JSON) without an associated protection scheme, which allows remote attackers to obtain the data via a web page that retrieves the data through a URL in the SRC attribute of a SCRIPT element and captures the data using other JavaScript code, aka \"JavaScript Hijacking.\"",
            "reference_id": "CVE-2007-2383"
        }
    ]
}
```

sbs2001:
The new models and import process fix this

PR #58, opened 2019-09-24
Use get or create to reduce the number of objects created

### LINKED ISSUE #40: Replace objects.create with objects.update_or_create.
Replace objects.create with objects.update_or_create along with new tests regarding this.

singh1114:
Fixes #40

Signed-off-by: Ranvir Singh <ranvir.singh1114@gmail.com>

### DIFF
--- a/vulnerabilities/data_dump.py
+++ b/vulnerabilities/data_dump.py
@@ -34,39 +34,39 @@ def debian_dump(extract_data):
     Save data scraped from Debian' security tracker.
     """
     for data in extract_data:
-        vulnerability = Vulnerability.objects.create(
+        vulnerability, _ = Vulnerability.objects.get_or_create(
             summary=data.get('description', ''),
         )
-        VulnerabilityReference.objects.create(
+        VulnerabilityReference.objects.get_or_create(
             vulnerability=vulnerability,
             reference_id=data.get('vulnerability_id', ''),
         )
 
         pkg_name = data.get('package_name', '')
-        package = Package.objects.create(
+        package, _ = Package.objects.get_or_create(
             name=pkg_name,
             version=data.get('version', ''),
         )
 
         if data['status'] == 'open':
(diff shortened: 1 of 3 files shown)


singh1114:
@pombredanne Did that!

singh1114:
> This is looking good otherwise, though do you think there could be a unit test that would show that there were issues with a bare create before?

I think we already had tests which specified the number of objects being created. Reducing the count in the tests itself will account for the same thing.

haikoschol:
@singh1114 I still don't see how this PR fixes #40. Did you mean to reference #28, which is about duplicate data?

Either way, I think the import process needs to be redesigned (and called "import" instead of "data dump", but let's not bikeshed about that just now :). Apart from issues #28 and #40, the current process is also not scalable, already way too slow, and it will only get worse by using either `get_or_create` or `create_or_update`.

We should distinguish between importing historical data once, when an instance of vulnerablecode is provisioned, and continuously importing updates from vulnerability sources. Otherwise the time for the latter process will continue to grow with the amount of data.

singh1114:
@haikoschol Sorry meant to refer #28 for the duplicate data. #40 doesn't seem to be correct and is a duplicate to #28 

Agreed with your view of differentiating both the procedures.

For improving the speed of the dump, we can use `bulk_create` (and it would be good to call it dump if we use that but handling duplicates in that case will be more difficult).

Anyway, we will have to make sure the data in the DB is unique.

haikoschol:
@singh1114 I agree that we should use `bulk_create`, at least for the initial import, but probably for updates as well.

However, I don't think that #40 is a duplicate of #28. I suggest we continue the discussion what #40 is about directly on the issue instead of this PR.

pombredanne:
At this stage and based on the discussions here and the changes that were applied, I think it is best to close and continue the discussion in #40

PR #125, opened 2019-11-04
Add unique CVE ID field to Vulnerability

### LINKED ISSUE #20: Collect vulnerabilities from arch linux
The data is available at https://security.archlinux.org/

haikoschol:
This change adds the field cve_id to the Vulnerability model and based
on that, improves the data import for Arch Linux.

The improvements made have been discussed in issue #20:

- For each CVE in a given AVG, exactly one Vulnerability is stored

- For each CVE, one VulnerabilityReference to its page on
  security.archlinux.org is stored

- Each ASA mentioned in an AVG is stored as a VulnerabilityReference

Since there is no production deployment of vulnerablecode yet, I took
the opportunity of changing the models to remove all migrations and
create a new one that creates the whole schema.

Since the cve_id field on Vulnerability has a unique constraint set, I
needed to make some changes to the import code that belong to issue #28.
I kept them minimal however so #28 is still open and needs to be
addressed later.

closes #20

Signed-off-by: Haiko Schol <hs@haikoschol.com>


vulnerabilities/data_dump.py
```diff
@@ -34,15 +34,11 @@ def debian_dump(extract_data, base_release='jessie'):
...
     """
     for data in extract_data:
-        vulnerability = Vulnerability.objects.create(
-            summary=data.get('description', ''),
-        )
-        VulnerabilityReference.objects.create(
-            vulnerability=vulnerability,
-            reference_id=data.get('vulnerability_id', ''),
+        vulnerability, _ = Vulnerability.objects.get_or_create(
+            cve_id=data['cve_id'],
```
pombredanne: What if there is no cve_id in the data? will this crash? And in this case we would likely still want to create a new Vulnerability?

vulnerabilities/data_dump.py
```diff
@@ -105,90 +97,96 @@ def archlinux_dump(extract_data):
...
-                version=affected_version
-            )
-            ImpactedPackage.objects.create(
-                vulnerability=vulnerability,
-                package=package_affected
-            )
-            PackageReference.objects.create(
-                package=package_affected,
-                repository=f'https://security.archlinux.org/package/{package_name}',
+                version=avg['affected'],
```
pombredanne: Is affected always a single version?

vulnerabilities/scraper/npm.py
```diff
@@ -47,30 +47,27 @@ def remove_spaces(x):
...
+    data = json.load(urlopen(package_url))
+    return [v for v in data.get('versions', {})]
 
 
-def extract_version(package_name, aff_version_range, fixed_version_range):
+def extract_versions(package_name, aff_version_range, fixed_version_range):
     """
-    Seperate list of Affected version and fixed version from all version
-    using the range specified
+    Seperate list of affected versions and fixed versions from all version
```
pombredanne: nitpicking: `Separate list of affected versions and fixed versions from all versions`

vulnerabilities/scraper/npm.py
```diff
@@ -47,30 +47,27 @@ def remove_spaces(x):
...
+    using the ranges specified
     """
-
+    # FIXME This skips unfixed vulnerabilities
```
pombredanne: Good catch... this warrants a ticket

vulnerabilities/scraper/npm.py
pombredanne: Also unrelated, could it make sense to adopt "affect/fixed" rather "impacted/resolved" ? is this a better wording and more explicit?

vulnerabilities/tests/test_api.py
```diff
@@ -121,8 +121,7 @@ def test_ubuntu_response(self):
         self.assertEqual(1, len(result['vulnerabilities']))
 
         vuln = result['vulnerabilities'][0]
-        self.assertEqual(1, len(vuln['references']))
-        self.assertEqual('CVE-2012-3386', vuln['references'][0]['reference_id'])
+        self.assertEqual(0, len(vuln['references']))
```
pombredanne: Note: if we decide to use pytest and its dnajo extension, we could start using plain asserts everywhere.

vulnerabilities/tests/test_data_dump.py
```diff
@@ -59,24 +59,15 @@ def test_Vulnerability(self):
...
     def test_VulnerabilityReference(self):
         """
-        Check that all vulnerability references from the test data are stored in the database
+        Check that no vulnerability references were found in the test data
         """
-        self.assertEqual(3, VulnerabilityReference.objects.count())
-        self.assertTrue(VulnerabilityReference.objects.get(reference_id='CVE-2009-1382'))
-        self.assertTrue(VulnerabilityReference.objects.get(reference_id='CVE-2009-2459'))
-        self.assertTrue(VulnerabilityReference.objects.get(reference_id='CVE-2014-8242'))
+        self.assertEqual(0, VulnerabilityReference.objects.count())
```
pombredanne: Is this right? or needed at all? I just realize that the previous code was making assumptions about data across tests.

vulnerabilities/tests/test_data_dump.py
```diff
@@ -99,22 +90,22 @@ def test_ImpactedPackage(self):
         Check that all impacted packages from the test data are stored in the database
         """
         impacted_pkgs = ImpactedPackage.objects.all()
-        impacted_pkg = impacted_pkgs[0]
 
-        self.assertEqual(1, len(impacted_pkgs))
-        self.assertEqual('librsync', impacted_pkg.package.name)
-        self.assertEqual('0.9.7-10', impacted_pkg.package.version)
+        self.assertEqual(1, impacted_pkgs.count())
```
pombredanne: I am always puzzled by tests that make assertions about things done in the setup or elsewhere... this was not your making, but if we could find a way to have tests that are really unit and do not depend on each other and on global setup() that would be better IMHO... what's your take? @tdruez ping too

vulnerabilities/tests/test_data_dump.py
```diff
@@ -154,52 +149,81 @@ def test_Vulnerability(self):
...
             self.assertEqual('pacman', pkg.type)
             self.assertEqual('archlinux', pkg.namespace)
 
     def test_PackageReference(self):
         """
-        Check that all package references from the test data are stored in the database
+        Check that no package references were found in the test data
         """
-        self.assertEqual(8, PackageReference.objects.count())
+        self.assertEqual(0, PackageReference.objects.count())
```
pombredanne: What is the purpose of this test?

vulnerabilities/tests/test_npm.py
```diff
@@ -23,7 +23,7 @@
 
 
 from django.test import TestCase
-from vulnerabilities.scraper.npm import remove_spaces, get_all_version, extract_data
+from vulnerabilities.scraper.npm import remove_spaces, get_all_versions, extract_data
```
pombredanne: I personally prefer having each import on its own line.
Also  separate imports in four sorted group with an empty line between each: future import, standard lib, thirdparty, own modules. Would this work for you too?

vulnerabilities/scraper/npm.py
haikoschol: Right. I meant to create one for it but apparently forgot. :-/ Done that now: https://github.com/nexB/vulnerablecode/issues/126

vulnerabilities/scraper/npm.py
haikoschol: I think either is fine. We should make sure to be consistent in the data we store, API responses, documentation, etc. but I'm fine sticking with impacted/resolved. If you are in favor of affected/fixed we should change it now though, since this PR already resets migrations.

vulnerabilities/data_dump.py
haikoschol: This is for the Debian import, where it doesn't look like there will be any entries without CVE ID. At least not in the feed we are currently consuming. The CVE ID is used as a dictionary key:
```
{
  "librsync": {
    "CVE-2014-8242": {
      "scope": "remote",
      "debianbug": 776246,
      "description": "librsync before 1.0.0 uses a truncated MD4 checksum to match blocks",
[...]
```

vulnerabilities/data_dump.py
haikoschol: It was a string in all the records I looked at so I'm pretty sure, but haven't tried to verify it any further.

In general I think it's not really a problem if the import crashes due to unexpected schemas. In fact, I'd argue we want it to crash as soon as possible instead of writing lots of empty values into the database due to overly defensive code.
We just need to make sure to be notified about any crashes that happen in cron jobs and such.

vulnerabilities/tests/test_api.py
haikoschol: Right, I had that thought as well. Wrote it down in a ticket now: https://github.com/nexB/vulnerablecode/issues/127

vulnerabilities/tests/test_data_dump.py
haikoschol: It is correct, although it's not ideal for the test, that for this data no vulnerability references are created any more. :)

I'm not sure what you mean with "assumptions about data across tests" though. The way the tests work now with `setUpTestData()` is that the data is loaded once and then used for assertions in all test methods. Since all of the tests are read-only, it should not  make a difference in outcome while being a bit more efficient. Should be easy to switch to plain `setUp()`/`tearDown()` though if necessary.

This test is for the debian import and since this PR is for an issue about Arch, I wanted to keep the changes minimal. But I just checked and [Debian has pages for CVEs](https://security-tracker.debian.org/tracker/CVE-2009-1382) just like Arch, so we could add those as vulnerability references.

vulnerabilities/tests/test_data_dump.py
haikoschol: I might actually have changed that in an earlier PR from having the import called in every test, which was tedious to change. If we use pytest we also have the option to write tests as simple functions instead of classes. If we do that I would still prefer to reuse setup code where it makes sense, either by calling a setup function in the test or using pytest decorators. Maybe that way it is more explicit?

vulnerabilities/tests/test_data_dump.py
haikoschol: We had a discussion on the chat a while ago about the purpose of `PackageReference` after we added the PURL mixin. I don't think any decisions/tickets came out of it, but until we decide what to do with it, this just checks that the ArchLinux import does not currently create any package references. For example we might eventually ask ourselves "What are we using PackageReference for?" this will answer it for Arch.

vulnerabilities/tests/test_npm.py
haikoschol: yes

vulnerabilities/tests/test_data_dump.py
pombredanne: > Should be easy to switch to plain setUp()/tearDown() though if necessary.

My point there was I personally find that the use of setUp()/tearDown() and similar to be confusing and making the tests less readable and less "unit" by bringing in some global state . But that's minor and nothing to fix just now.

PR #133, opened 2019-12-14
Change Django  Tests to pytests

sbs2001:

I have used [this  pytest extension](https://github.com/pytest-dev/pytest-django). We are able to run all the tests simply by running `pytest` in the root directory.

I am currently working on fixing the  following warnings(I could use some help here). 

```
(venv) shivam@shivam-Latitude-D630:~/coding/opensource/vulnerablecode$ pytest
===================================================================== test session starts =====================================================================
platform linux -- Python 3.6.9, pytest-5.3.2, py-1.8.0, pluggy-0.13.1
Django settings: vulnerablecode.settings (from ini file)
rootdir: /home/shivam/coding/opensource/vulnerablecode, inifile: pytest.ini
plugins: django-3.7.0
collected 32 items                                                                                                                                            

vulnerabilities/tests/test_api.py ......                                                                                                                [ 18%]
vulnerabilities/tests/test_data_dump.py ............                                                                                                    [ 56%]
vulnerabilities/tests/test_models.py ....                                                                                                               [ 68%]
vulnerabilities/tests/test_api_data.py ..                                                                                                               [ 75%]
vulnerabilities/tests/test_import_cli.py ...                                                                                                            [ 84%]
vulnerabilities/tests/test_npm.py ...                                                                                                                   [ 93%]
vulnerabilities/tests/test_scrapers.py ..                                                                                                               [100%]

====================================================================== warnings summary =======================================================================
vulnerabilities/tests/test_api.py::test_debian_query_by_name
vulnerabilities/tests/test_api.py::test_debian_query_by_invalid_package_url
vulnerabilities/tests/test_api.py::test_debian_query_by_package_url
vulnerabilities/tests/test_api.py::test_debian_query_by_package_url_without_namespace
vulnerabilities/tests/test_api.py::test_ubuntu_response
  /home/shivam/coding/opensource/vulnerablecode/venv/lib/python3.6/site-packages/whitenoise/base.py:116: UserWarning: No directory at: /home/shivam/coding/opensource/vulnerablecode/staticfiles/
    warnings.warn(u"No directory at: {}".format(root))

vulnerabilities/tests/test_api.py::test_debian_query_by_name
vulnerabilities/tests/test_api.py::test_ubuntu_response
  /home/shivam/coding/opensource/vulnerablecode/venv/lib/python3.6/site-packages/django_filters/rest_framework/backends.py:31: MigrationNotice: `PackageViewSet.filter_fields` attribute should be renamed `filterset_fields`. See: https://django-filter.readthedocs.io/en/master/guide/migration.html
    filterset_class = self.get_filterset_class(view, queryset)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================== 32 passed, 7 warnings in 88.52s (0:01:28) ==========================================================
```


haikoschol:
@sbs2001 Thanks for this PR!

Since all tests are now being discovered and run with one command, can you also remove [the old call to pytest from `.travis.yml`](https://github.com/nexB/vulnerablecode/pull/133/files#diff-354f30a63fb0907d4ad57269548329e3L19)?

I think the warnings are unrelated to your changes. My guess is that pytest shows them by default whereas the Django test runner doesn't. Two dependencies (whitenoise and Django REST Framework) use features of Django that will be removed in a future version. It is possible that there are more recent versions of those apps that address these warnings. You can try changing their versions in `requirements.txt` to the latest to check.

In the chat you wrote:

> I am bit shaky about whether I used the fixtures and marks effectively.

I think you use them correctly. However, the marks are only used in `test_models.py` which does not actually test any behaviour of vulnerablecode but just basic Django ORM operations. I don't think having these tests provides any benefit so I deleted the file in my pending PR #132.

And regarding the fixtures; Having them in a separate file increases the distance to the code using them and arguably makes it harder to read the tests. @pombredanne commented elsewhere that he prefers tests to be self-contained if possible. Personally I don't mind it and prefer to avoid duplication if that is the tradeoff to be made. Maybe @pombredanne wants to chime in on this? Otherwise, if we do leave it as is, would you mind renaming the file from `conftest.py` to `fixtures.py`?

I didn't expect you to change all the existing tests from `TestCase` subclasses to simple functions. But since you did I'm fine with keeping them that way. However by removing the extra namespace of the class, you created a few name clashes, effectively removing some tests. If you look at the CI log on your PR, it says 27 passed after running `pytest`. Before, running `manage.py test`, it reported 28 tests. The four tests that needed to be run separately before are now included in running `pytest`, so the total number of tests should be 32.

vulnerabilities/tests/test_data_dump.py
```diff
@@ -23,207 +23,180 @@
...
+    assert 'deb' == pkg.type
+    assert 'ubuntu' == pkg.namespace
+
+
+CVE_IDS = ('CVE-2018-11362', 'CVE-2018-11361', 'CVE-2018-11360',
+           'CVE-2018-11359', 'CVE-2018-11358', 'CVE-2018-11357',
+           'CVE-2018-11356', 'CVE-2018-11355', 'CVE-2018-11354')
+
+
+def test_Vulnerability(setArchLinuxData):
```
haikoschol: This one overwrites the previous `test_Vulnerability` for Debian, right? I would suggest just keeping the `TestCase` based implementation of the tests in this file. But feel free to resolve it differently, as long as no tests are accidentally removed. :)

vulnerabilities/tests/test_api.py
```diff
@@ -23,129 +23,103 @@
...
-from django.test import TestCase
+import pytest
 
 from vulnerabilities.api import PackageSerializer
 from vulnerabilities.data_dump import debian_dump
 from vulnerabilities.data_dump import ubuntu_dump
 from vulnerabilities.models import Package
 from vulnerabilities.scraper import debian
 from vulnerabilities.scraper import ubuntu
+from django.test import Client
```
haikoschol: Would you mind moving this import between the ones from the stdlib and the ones from `vulnerablecode`?

sbs2001:
@haikoschol I have corrected the code.  I changed the names to avoid name  clashes. The warnings are not fixed though. Thanks!

haikoschol:
> The warnings are not fixed though.

We can look into that after merging.

> Regarding conftest.py , after researching albeit I don't think we can change that to fixtures.py .

I didn't realize that the filename has special meaning for pytest. In that case we'll leave it as is.

vulnerabilities/tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+    Package.objects.create(
+        name='mimetex',
+        version='1.50-1.1',
+        type='deb',
+        namespace='ubuntu'
+    )
+
+
+@pytest.fixture
+def setdDebianData(db):
```
haikoschol: This name is *very* easy to confuse with `setDebianData`. :) I think we don't need both functions though. The extra `Package` object that is created in `setDebianData` is only required in one test (`test_debian_query_by_package_url_without_namespace`) to verify that the namespace part of package URLs is optional.

I suggest you delete `setDebianData`, rename `setdDebianData` to `setDebianData` and add the creation of the Ubuntu package to the beginning of `test_debian_query_by_package_url_without_namespace`. That makes it clear that this test data is only needed there. If you do this, you'll also need to adapt some assertions in the other tests that use the current version of `setDebianData`. For example in `test_debian_query_by_name`, four records would be expected in the response instead of five.

sbs2001: I had used that for debugging purpose and forgot to remove it. I have removed it now.Sorry

Issue #126, opened 2019-11-06
NPM import skips unfixed vulnerabilities


haikoschol:
The [function that resolves version ranges](https://github.com/nexB/vulnerablecode/blob/58e3ff375f8cc40cd457aa8130a5ac5264e29cd7/vulnerabilities/scraper/npm.py#L68) for affected and fixed packages returns empty lists if either of the ranges is empty. For unfixed vulnerabilities, the fixed version range is usually empty, hence no data for those will be added.

PR #138, opened 2019-12-21
Fix npm.py bug

sbs2001:

Possibly fixes https://github.com/nexB/vulnerablecode/issues/126 

Also implements the following logic and prevents the behaviour given below.

> Unaffected version is that version which  is in the fixed_version_range or which is absent in the aff_version_range

Before patch:

```
>>> import npm
>>> z=npm.extract_versions('ldapauth',"<=2.2.4","<0.0.0")
>>> z
(['1.0.0', '1.0.1', '1.0.2', '2.0.0', '2.1.0', '2.2.0', '2.2.1', '2.2.2', '2.2.3', '2.2.4'], [])
```
After patch:

```
>>> import npm
>>> z=npm.extract_versions('ldapauth',"<=2.2.4","<0.0.0")
>>> z
(['1.0.0', '1.0.1', '1.0.2', '2.0.0', '2.1.0', '2.2.0', '2.2.1', '2.2.2', '2.2.3', '2.2.4'], ['2.3.0', '2.3.1'])
```
Here are some details regarding [ldapauth package](https://registry.npmjs.org/ldapauth).

Link to ldapauth advisory [here](https://registry.npmjs.org/-/npm/v1/security/advisories?page=1) (check for id 19)

### DIFF
--- a/vulnerabilities/scraper/npm.py
+++ b/vulnerabilities/scraper/npm.py
@@ -58,25 +58,22 @@ def get_all_versions(package_name):
 
 def extract_versions(package_name, aff_version_range, fixed_version_range):
     """
-    Seperate list of affected versions and fixed versions from all versions
-    using the ranges specified
+    Seperate list of affected versions and unaffected versions from all versions
+    using the ranges specified.
     """
-    # FIXME This skips unfixed vulnerabilities
-    if aff_version_range == '' or fixed_version_range == '':
-        return ([], [])
-
     aff_spec = semantic_version.NpmSpec(remove_spaces(aff_version_range))
     fix_spec = semantic_version.NpmSpec(remove_spaces(fixed_version_range))
     all_ver = get_all_versions(package_name)
     aff_ver = []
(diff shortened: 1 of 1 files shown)


vulnerabilities/scraper/npm.py
```diff
@@ -58,25 +58,24 @@ def get_all_versions(package_name):
...
-    if aff_version_range == '' or fixed_version_range == '':
-        return ([], [])
-
     aff_spec = semantic_version.NpmSpec(remove_spaces(aff_version_range))
     fix_spec = semantic_version.NpmSpec(remove_spaces(fixed_version_range))
     all_ver = get_all_versions(package_name)
     aff_ver = []
     fix_ver = []
+    """
+    Unaffected version is that version which  is in the fixed_version_range
```
haikoschol: Please use `#` for comments instead of a multi-line string.

sbs2001: Ok done!

Issue #86, opened 2019-09-26
Collect Rubygems advisories from https://rubysec.com/


pombredanne:
See https://rubysec.com/

sbs2001:
[...]

PR #143, opened 2020-01-17
Add ruby gems advisory importer

sbs2001:

Fixes https://github.com/nexB/vulnerablecode/issues/86


sbs2001:
I will soon write some tests for this importer

haikoschol:
@sbs2001 Thanks for this PR! A few general nitpicks (other comments inline):
- please use `snake_case` instead of `camelCase` for variable names
- order imports alphabetically in three blocks: standard library, third-party libraries, modules inside vulnerablecode
- leave out parentheses in `if` and `for` if they are not necessary (i.e. not a multi-line statement)

sbs2001:
> @sbs2001 Thanks for this PR! A few general nitpicks (other comments inline):
> 
>     * please use `snake_case` instead of `camelCase` for variable names
> 
>     * order imports alphabetically in three blocks: standard library, third-party libraries, modules inside vulnerablecode
> 
>     * leave out parentheses in `if` and `for` if they are not necessary (i.e. not a multi-line statement)

@haikoschol  Thanks for the review! I have made the suggested changes.

vulnerabilities/scraper/ruby.py
```diff
@@ -0,0 +1,104 @@
...
+                    for spec in specs:
+                        if version in spec:
+                            unaffected_versions.add(version)
+                            break
+            affected_versions = all_versions - unaffected_versions
+            vulnerability_to_package_map.append({
+                'package_name': package_name,
+                'summary': summary,
+                'cve_id': vulnerability_id,
+                'fixed_versions': unaffected_versions,
```
haikoschol: I think this came up elsewhere before: Fixed (or patched) and unaffected packages are not the same thing. The rubysec yaml schema includes both:
```
$ grep -A 2 _versions gems/nokogiri/CVE-2015-5312.yml
unaffected_versions:
  - "< 1.6.0"

patched_versions:
  - ">= 1.6.7.1"
```
With the current data model of vulnerablecode, unaffected versions would just be packages that have neither a corresponding `ImpactedPackage` nor a `ResolvedPackage`. I don't know off the top of my head whether we make this distinction elsewhere, but this is one reason why we need a better architecture for importers: To have one place where those decisions are made.

vulnerabilities/scraper/ruby.py
```diff
@@ -0,0 +1,104 @@
...
+            vulnerability = saneyaml.load(yaml_file)
+            package_name = vulnerability.get(
+                'engine', vulnerability.get('gem'))
+            summary = vulnerability.get('description', '')
+            if 'cve' in vulnerability:
+                vulnerability_id = 'CVE-{}'.format(vulnerability['cve'])
+                if vulnerability_id in ids:
+                    continue
+                ids.add(vulnerability_id)
+            else:
```
haikoschol: Eventually we will have to include vulnerabilities without a CVE ID as well, since that is meant to be one of the defining features of vulnerablecode.

vulnerabilities/data_dump.py
```diff
@@ -190,3 +190,44 @@ def npm_dump(extract_data):
                     vulnerability=vulnerability,
                     package=package_fixed
                 )
+
+
+def ruby_dump(extract_data):
+    for package_data in extract_data:
+
+        if package_data['severity']:
```
haikoschol: This can be written much shorter and without repetition by using `dict.get()`, since `cvss` is a nullable field in the model:
```
    vulnerability, _ = Vulnerability.objects.get_or_create(
        cve_id=package_data['cve_id'],
        summary=package_data['summary'],
        cvss=package_data.get('severity')
    )
```
But see my other comment below about skipping severity for now.

vulnerabilities/data_dump.py
```diff
@@ -190,3 +190,44 @@ def npm_dump(extract_data):
...
+                summary=package_data['summary']
+            )
+
+        VulnerabilityReference.objects.get_or_create(
+            vulnerability=vulnerability,
+            url=package_data['advisory']
+        )
+
+        for version in package_data['affected_versions']:
+            affected_package = Package.objects.create(
```
haikoschol: We also need `type="gem"` here to enable querying by "Package URL".

However, this is only correct for actual Ruby gems. The data we are storing here also contains vulnerabilities in Ruby implementations (MRI (aka CRuby), JRuby and Rubinius (aka rbx)). Those are not gems and therefore not distributed via rubygems.org, as far as I can tell. Arguably those are not even packages. They are the stuff that goes into packages distributed by Debian, Fedora, etc. This issue has come up in https://github.com/package-url/purl-spec/issues/13 a long time ago and I don't think it will be solved before we merge this PR. So for the time being, I suggest we skip security advisories for Ruby implementations.

vulnerabilities/data_dump.py
```diff
@@ -190,3 +190,44 @@ def npm_dump(extract_data):
...
+                name=package_data['package_name'],
+                version=version
+            )
+            ImpactedPackage.objects.create(
+                vulnerability=vulnerability,
+                package=affected_package
+            )
+
+        for version in package_data['fixed_versions']:
+            unaffected_package = Package.objects.create(
```
haikoschol: Same here: Add `type="gem"`.

vulnerabilities/scraper/ruby.py
```diff
@@ -0,0 +1,104 @@
...
+    path_to_zip, _ = urllib.request.urlretrieve(
+        RUBYCVE_LINK, os.path.join(
+            DOWNLOAD_PATH, 'ruby.zip'))
+    ZipFile(path_to_zip).extractall(DOWNLOAD_PATH)
+    os.remove(path_to_zip)
+
+
+def path_of_yaml_of_all_packages():
+    gem_path = os.path.join(DOWNLOAD_PATH, 'ruby-advisory-db-master', 'gems')
+    rubies_path = os.path.join(
```
haikoschol: See [my previous comment](https://github.com/nexB/vulnerablecode/pull/143/files#diff-b585e6dc3a5cc26d34cbf5ca4b03dcd5R216) about this: I think for now we want to skip the "rubies" directory.

vulnerabilities/scraper/ruby.py
```diff
@@ -0,0 +1,104 @@
...
+    vulnerability_to_package_map = []
+    for vulnerability_path in path_of_yaml_of_all_packages():
+        with open(vulnerability_path) as yamlFile:
+            vulnerability = saneyaml.load(yamlFile)
+            package_name = vulnerability.get(
+                'engine', vulnerability.get('gem'))
+            summary = vulnerability.get('description', '')
+            if 'cve' in vulnerability:
+                vulnerability_id = 'CVE-{}'.format(vulnerability['cve'])
+                if vulnerability_id in ids:
```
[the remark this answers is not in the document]
sbs2001: There are duplicates of same CVE's , while  doing the data dump django gives error  due to violating some constraint defined in `models.py` so I had to implement the sets to skip duplicates.

vulnerabilities/scraper/ruby.py
```diff
@@ -0,0 +1,104 @@
...
+                continue
+            yield RangeSpecifier(spec)
+    else:
+        return None
+
+
+def import_vulnerabilities():
+    get_rubycve_db()
+    ids = set()
+    vulnerability_to_package_map = []
```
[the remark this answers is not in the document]
sbs2001: It's a list of dictionaries, each dictionary mapping a vulnerability to affected package. Any better names you can think of ? Naming is tough

vulnerabilities/scraper/ruby.py
sbs2001: Exactly, that would need some refactoring for all the current importers.

vulnerabilities/scraper/ruby.py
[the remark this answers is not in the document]
sbs2001: I have changed it `vulnerability_package_dicts` for now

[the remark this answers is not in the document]
haikoschol: Yes, naming is hard. In general I think it's preferable to avoid describing the data structure in the variable name. But let's not get hung up on this here and stick with the current name.

vulnerabilities/scraper/ruby.py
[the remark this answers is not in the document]
haikoschol: Sorry, I missed that the `continue` skips processing the vulnerability entirely. For that the check is necessary, but then the question is how those duplicates occur. One vulnerability can affect more than one package. An example for this is [CVE-2015-1840, which affects jquery-rails and jquery-ujs](https://www.honeybadger.io/blog/understanding-the-rails-jquery-csrf-vulnerability-cve-2015-1840/).
A quick fix for this is to call `Vulnerability.objects.get_or_create()` instead of `create()` when storing the data. This is probably not a long-term solution, but it's good enough for now.

Issue #92, opened 2019-09-26
Collect Rust Cargo


pombredanne:
https://github.com/RustSec/advisory-db

PR #148, opened 2020-02-10
Add rust security importer

sbs2001:
This is the format of data obtained from  https://github.com/RustSec/advisory-db
`{'package_name': 'yaml-rust', 'vuln_id': 'RUSTSEC-2018-0006', 'fixed_versions': {'0.4.1', '0.4.3', '0.4.2'}, 'affected_versions': {'0.3.5', '0.3.0', '0.2.2', '0.2.1', '0.3.2', '0.3.3', '0.4.0', '0.2.0', '0.3.4'}, 'advisory': 'https://github.com/chyh1990/yaml-rust/pull/109'}`

Fixes https://github.com/nexB/vulnerablecode/issues/92

### DIFF
--- a/vulnerabilities/scraper/rust.py
+++ b/vulnerabilities/scraper/rust.py
@@ -1,11 +1,13 @@
+from io import BytesIO
 import json
+
+from dephell_specifier import RangeSpecifier
 import pytoml as toml
 from urllib.request import urlopen
 import urllib.request
 from urllib.error import HTTPError
-from io import BytesIO
 from zipfile import ZipFile
-from dephell_specifier import RangeSpecifier
+
 
 RUSTSEC_DB_URL = 'https://github.com/RustSec/advisory-db/archive/master.zip'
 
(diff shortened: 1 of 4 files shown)


vulnerabilities/scraper/rust.py
```diff
@@ -0,0 +1,90 @@
...
+def rust_crate_advisories(url, prefix='advisory-db-master/crates/'):
+    with urlopen(url) as response:
+        with ZipFile(BytesIO(response.read())) as zf:
+            for path in zf.namelist():
+                if path.startswith(prefix) and path.endswith('.toml'):
+                    yield toml.load(zf.open(path))
+
+
+def all_versions_of_crate(crate):
+    info_url = "https://crates.io//api/v1/crates/{}".format(crate)
```
haikoschol: nit: There's an extra `/` after `.io`.

vulnerabilities/scraper/rust.py
```diff
@@ -0,0 +1,90 @@
...
+    info = json.load(urlopen(info_url))
```
haikoschol: I can't find any mention in the documentation but I'm pretty sure the [`HTTPResponse`](https://docs.python.org/3/library/http.client.html#httpresponse-objects) object returned by `urlopen()` needs to be closed to avoid a resource leak. This is the reason for `with urlopen(url) as response:` in `rust_crate_advisories()`.

vulnerabilities/scraper/rust.py
sbs2001: Fixed , btw any idea why it worked while using wrong url?

haikoschol: I think URLs with multiple consecutive slashes in the path component are valid (see [this StackOverflow thread](https://stackoverflow.com/questions/20523318/is-a-url-with-in-the-path-section-valid)). That's why I included the "nit" prefix. :)

Issue #123, opened 2019-10-22
Manage vulnerability sources in database


haikoschol:
Add a model for available data sources where we can keep metadata such as the license under which the vulnerability data is published. This would make it easier to maintain this kind of information than having it in the code. When we have importers for standardized formats like OVAL it would also allow to add additional data sources that follow this standard by adding a new entry to this DB table.

An implementation of this would benefit from a bit of refactoring to some kind of base class for vulnerability sources.

haikoschol:
Some quick notes:
My initial idea was to have a base class called `Importer` or something like that, which defines an interface and contains common logic from which implementations for various data souces inherit. 

An alternative would be to have only one `Importer` which is a Django model and contains fields like
- `name` (e.g. "ubuntu", "npm", etc.)
- `enabled`
- `last_run`
- one or two fields that contain names of classes which contain the data source specific code. That code can be roughly categorized as "fetching data" and "processing data".
- possibly a JSON field with configuration for the data source specific code 

Decoupling the fetching from the processing seems to help with using fixtures for testing, but I'm not sure there is enough to do for the fetching to warrant a separate class. That's not very important though as long as the interface consumed by the `Importer` separates these steps clearly. One might expect only providing the data newer than a given timestamp (to be used with `Importer.last_run`) to be part of the "fetching" step as well, but in many cases that will require parsing the data format. 

The processing step would take whatever the format of the data source is (OVAL XML, one huge JSON, one YAML document per advisory, etc.) and transform it into a common intermediate format that is then turned into model objects and stored in the database by the `Importer`.
The intermediate format also helps with testing (no need to run a full import and then assert that expected rows have been written to the database) and it allows us to keep iterating on the data model without having to constantly update all importers (we might need a new name for the classes that implement the data source specific logic if there is one common class called `Importer` as well).

All this should be done in a "streaming" fashion, probably using generators and somehow reasonably efficiently written to the DB (probably using [bulk inserts](https://www.caktusgroup.com/blog/2019/01/09/django-bulk-inserts/)).

Requirements for the new and improved import process are:
- Preventing duplicates in the database
- Ability to deal with updated information (i.e. a preliminary security advisory is updated with more details)
- Reasonable efficiency & performance (i.e. using bulk inserts instead of making a few SELECT queries to check for existing data and possibly timestamps on it before issuing a single INSERT)

haikoschol:
closed by #152

PR #152, opened 2020-03-01
Manage vulnerability sources in database

haikoschol:
This is an implementation of #123.

The change adds a new model `Importer` that stores metadata as well as the name of a class (a `vulnerabilities.data_source.DataSource`) that fetches and transforms vulnerability data into a common, DB schema agnostic format. This format is encoded in the two data classes `vulnerabilities.data_source.Package` and `vulnerabilities.data_source.VulnerabilityInfo`.

Running an `Importer` and storing the results in the database is done by `ImportRunner`.

The unused and redundant model `PackageReference` has been removed.

Another minor, but overdue change is renaming the `scraper` module to `importers`.

@Reviewers:

It might be helpful to start with looking at how the `import` management command uses the above described classes and then jump into the rest of the code from there.


vulnerabilities/models.py
```diff
@@ -123,3 +128,36 @@ class PackageReference(models.Model):
...
+    def make_data_source(self, cutoff_date=None) -> DataSource:
+        """
+        Return a configured and ready to use instance of
+        this importers data source implementation.
+
+        cutoff_date - timestamp of the oldest data to include in the import (default: self.last_run)
+        """
+        cd = cutoff_date or self.last_run
+        importers_module = importlib.import_module('vulnerabilities.importers')
+        cls = getattr(importers_module, self.data_source)
```
sbs2001:
```
 cls = getattr(importers_module, self.data_source)
 return cls(cutoff_date=cd, config=self.data_source_cfg)
```
 `cls` is a python keyword ,IMHO to avoid confusion, if this variable can be renamed please consider renaming it.

haikoschol: It's not a keyword, but a name conventionally used in one specific case and this is not it. I'll rename it to `klass`.

vulnerabilities/import_runner.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    vulnerabilities = set()
+
+    for vuln_info in batch:
+        vuln: models.Vulnerability
+
+        if vuln_info.cve_id:
+            vuln, created = models.Vulnerability.objects.get_or_create(cve_id=vuln_info.cve_id)
+            if created:
+                vuln.summary = vuln_info.summary
```
sbs2001: We need some way  to determine, which `summary` is better and then decide to update the summary or leave it as it is.

haikoschol: I'm not optimistic there is one. But even if there was a way, if all we have is the summary, we can't detect that a new vulnerability is actually an update to an existing vulnerability.

sbs2001: I was thinking of moving `summary` from `Vulnerability` model to the `ImpactedPackage` model . So the context won't be lost.

haikoschol: In theory it makes sense that the same vulnerability could be described in different terms, depending on the affected package. But in practice, do any of the source we ingest so far do this? Or would we end up copying the same description to every `ImpactedPackage`? That wouldn't really be an issue though.

Fundamentally the questions are:
* What is the definition of "vulnerability"?
* What is the minimum number of attributes (and what are those) for identifying a vulnerability?

vulnerabilities/import_runner.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def _get_or_create_vulnerability(advisory: Advisory) -> Tuple[models.Vulnerability, bool]:
+    query_kwargs = {}
+
+    if advisory.cve_id:
+        query_kwargs['cve_id'] = advisory.cve_id
+    else:
+        query_kwargs = {'summary': advisory.summary}
+    vuln, created = models.Vulnerability.objects.get_or_create(**query_kwargs)
```
sbs2001, line 135: Question: Is there any reason, for not considering the case where `cve_id` and `summary` both are available, in such case shouldn' those be added to `query_kwargs`?

 By the way, this is really great work.

haikoschol: Thanks for the kind words! It sure is taking long enough. :)

Including both in the query would only work if we add a "unique together" constraint for `cve_id` and `summary`. As it is now, if we have the vulnerability with a different summary and include both in the call to `get_or_create()`, we get this:
```
In [54]: Vulnerability.objects.count()                                                                                                      
Out[54]: 0

In [55]: Vulnerability.objects.create(cve_id='abc', summary='remote root many bad')                                                         
Out[55]: <Vulnerability: remote root many bad>

In [56]: Vulnerability.objects.get_or_create(cve_id='abc', summary='other summary for the same vuln')                                       
---------------------------------------------------------------------------
DoesNotExist                              Traceback (most recent call last)
~/code/vulnerablecode/venv/lib/python3.8/site-packages/django/db/models/query.py in get_or_create(self, defaults, **kwargs)
    558         try:
--> 559             return self.get(**kwargs), False
    560         except self.model.DoesNotExist:

~/code/vulnerablecode/venv/lib/python3.8/site-packages/django/db/models/query.py in get(self, *args, **kwargs)
    414         if not num:
--> 415             raise self.model.DoesNotExist(
    416                 "%s matching query does not exist." %

DoesNotExist: Vulnerability matching query does not exist.

During handling of the above exception, another exception occurred:

UniqueViolation                           Traceback (most recent call last)
~/code/vulnerablecode/venv/lib/python3.8/site-packages/django/db/backends/utils.py in _execute(self, sql, params, *ignored_wrapper_args)
     85             else:
---> 86                 return self.cursor.execute(sql, params)
     87 

UniqueViolation: duplicate key value violates unique constraint "vulnerabilities_vulnerability_cve_id_key"
DETAIL:  Key (cve_id)=(abc) already exists.


The above exception was the direct cause of the following exception:

IntegrityError                            Traceback (most recent call last)
<ipython-input-56-196121bcc8ce> in <module>
----> 1 Vulnerability.objects.get_or_create(cve_id='abc', summary='other summary for the same vuln')

~/code/vulnerablecode/venv/lib/python3.8/site-packages/django/db/models/manager.py in manager_method(self, *args, **kwargs)
     80         def create_method(name, method):
     81             def manager_method(self, *args, **kwargs):
---> 82                 return getattr(self.get_queryset(), name)(*args, **kwargs)
     83             manager_method.__name__ = method.__name__
     84             manager_method.__doc__ = method.__doc__

--------------- preposterously long traceback snipped ----------------------

~/code/vulnerablecode/venv/lib/python3.8/site-packages/django/db/backends/utils.py in _execute(self, sql, params, *ignored_wrapper_args)
     84                 return self.cursor.execute(sql)
     85             else:
---> 86                 return self.cursor.execute(sql, params)
     87 
     88     def _executemany(self, sql, param_list, *ignored_wrapper_args):

IntegrityError: duplicate key value violates unique constraint "vulnerabilities_vulnerability_cve_id_key"
DETAIL:  Key (cve_id)=(abc) already exists.
```

NB: As you see further below in the code I made the completely arbitrary decision to update the database with the summary from the currently imported data source in the case they differ.

vulnerabilities/data_source.py
```diff
@@ -0,0 +1,262 @@
...
+    def error(self, msg: str) -> None:
+        """
+        Helper method for raising InvalidConfigError with the class name in the message.
+        """
+        raise InvalidConfigError(f'{type(self).__name__}: {msg}')
+
+
+class GitDataSource(DataSource):
+    repository_url: Optional[str] = None
+    branch: str = 'master'
```
sbs2001, line 130: This will probably give an error if the user doesn't specify the branch to checkout and the repo doesn't have a `master` branch. 

I think `branch` should not have any default value.  The pygit2's docs at https://www.pygit2.org/repository.html#pygit2.clone_repository mentions  regarding `checkout_branch`  
`Branch to checkout after the clone. The default is to use the remote’s default branch.`

haikoschol: Thanks for pointing that out! You are right, the correct behaviour is to use the repository's default branch. I just realized that today as well and already changed it.

vulnerabilities/data_source.py
```diff
@@ -0,0 +1,371 @@
...
+                previous_commit = commit
+                continue
+
+            for d in commit.tree.diff_to_tree(previous_commit.tree).deltas:
+                if not _include_file(d.new_file.path, subdir, recursive, file_ext) or d.is_binary:
+                    continue
+
+                # TODO
+                # Just filtering on the two status values for "added" and "modified" is too simplistic.
+                # This does not cover file renames, copies & deletions.
```
sbs2001, line 282: This can be solved by checking the commit status with  `pygit2.GIT_DELTA_RENAMED` ,`pygit2.GIT_DELTA_COPIED` , and `pygit2.GIT_DELTA_DELETED`  https://github.com/libgit2/pygit2/blob/d4e328de64a1e872f6a9f708443d8e7bb81c97a6/src/pygit2.c#L561 .

I guess the pain part is actually making changes to the DB according the commit status.

vulnerabilities/data_source.py
```diff
@@ -0,0 +1,371 @@
...
+            # These really should be declared in DataSourceConfiguration above but that would prevent DataSource
+            # subclasses from declaring mandatory parameters (i.e. positional arguments)
+            setattr(self.config, 'last_run_date', last_run_date)
+            setattr(self.config, 'cutoff_date', cutoff_date)
+        except Exception as e:
+            raise InvalidConfigurationError(str(e))
+
+        self.validate_configuration()
+
+    def __enter__(self):
```
sbs2001, line 112: nit: `__enter__ ` and `__exit__` are already defined in the `ContextManager` abstract class, which is inherited by this class(DataSource) . These can be safely removed.

sbs2001: `ContextManager` is basically https://github.com/python/cpython/blob/eff870b618ca6f6b7a60a271f15af7e54b8a1b97/Lib/contextlib.py#L15 as mentioned in https://docs.python.org/3/library/typing.html#typing.ContextManager

vulnerabilities/data_source.py
haikoschol: Correct. I saw those constants, but the question is what all the scenarios are in which they need to be considered. For example, in [RustSec/advisory-db](https://github.com/RustSec/advisory-db) files are (sometimes?) added with a temporary name `RUSTSEC-0000-0000.toml`.

vulnerabilities/data_source.py
haikoschol: > nit: `__enter__ ` and `__exit__` are already defined in the `ContextManager` abstract class, which is inherited by this class(DataSource) . These can be safely removed.

Right. First I had some code in there but then moved it elsewhere. I'll remove it.

vulnerabilities/models.py
```diff
@@ -87,39 +90,38 @@ class Meta:
...
-    repository = models.CharField(
+    name = models.CharField(max_length=100, unique=True, help_text='Name of the importer')
+    license = models.CharField(max_length=100, blank=True, help_text='License of the vulnerability data')
+    last_run = models.DateTimeField(null=True, help_text='UTC Timestamp of the last run')
+
+    data_source = models.CharField(
         max_length=100,
-        help_text='Repository URL eg:http://central.maven.org',
-        blank=True,
+        help_text='Class name of the data source implementation importable from vulnerabilities.importers',
```
sbs2001, line 109: Just for clarity, so eventually  our current  `importers`(rust,npm etc ) will be inheriting from the `DataSource` class or `GitDataSource` (for git-based importers)?

vulnerabilities/data_source.py
sbs2001: Interesting case, could that be handled by passing some special config?

vulnerabilities/data_source.py
```diff
@@ -0,0 +1,371 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.config.remove_working_directory:
+            shutil.rmtree(self.config.working_directory)
+
+    def added_advisories(self) -> Set[Advisory]:
+        raise NotImplementedError
+
+    def updated_advisories(self) -> Set[Advisory]:
+        raise NotImplementedError
+
```
sbs2001, line 197: How about adding `deleted_advisories` abstract method ? Implementing this method for `GitDataSource`(and other future data sources), will solve the below TODO partially. `ImportRunner`  will also need to have the logic to deal with  deleted advisories.

vulnerabilities/data_source.py
haikoschol: What do you mean with special config? I don't know how it could be handled, that's why I added a `TODO` comment instead of writing the code for it. :)

vulnerabilities/data_source.py
haikoschol: Turns out if I remove them, some tests fail because inheriting from `DataSource`, which inherits from `ContextManager` does not give `MockDataSource` these methods. ¯\_(ツ)_/¯

vulnerabilities/models.py
haikoschol: yes

vulnerabilities/data_source.py
haikoschol: That would be easy enough to do, but I don't know how it would be used by classes inheriting form `DataSource` and I don't want to add methods speculatively. We can iterate and improve on it later. Let me just finish this bloody PR with as little extra fuss as possible. :)

sbs2001: Totally fine with me.

vulnerabilities/data_source.py
sbs2001: As this is a corner case, we could handle it later as you said. 
What I had in mind, is to pass in JSON data like `{'ignore':['RUSTSEC-0000-0000']}` in  `data_source_cfg` while registering the Importer model for rust.

As this `data_source_cfg` while be passed in to instantiate the corresponding `data_source` in the Importer model's `make_data_source` method, thus this  `data_source` has access to its `data_source_cfg`.

Now the data source for rust will most probably inherit from `GitDataSource` , we could then override/implement ` added_advisories` and `updated_advisories` such that it doesn't include vulnerabilities which have ids present in it's `data_source_cfg` 's  'ignore' list .

I hope I make sense :)

haikoschol: Ah, now I understood what you mean with special config. Yes, that should work. I just finished the Rust importer based on `GitDataSource` and just filter the files in the code there (feedback welcome on that as well of course). But it could certainly be generalized like that if we need it in many importers.

vulnerabilities/importers/rust.py
```diff
@@ -1,99 +1,176 @@
...
+
+    return unaffected, affected
+
+
+class CratesAPI:
+    def __init__(self, cache=None):
+        self.versions_cache = cache or {}
+
+    def all_versions_of_crate(self, crate: str) -> Set[str]:
+        if crate in self.versions_cache:
```
sbs2001, line 164: I haven't finished reviewing this in detail. So I might be missing some things here . 
This API provides last modification date, how about using it , to update the cache.

I don't know what difference it would make, but will using a single  `requests.Session()`  instead of the current solution to make  all the API calls increase the performance,  https://2.python-requests.org/en/v2.8.1/user/advanced/#session-objects ?
Actually we could make huge gains in performance at the cost of increased complexity using asynchronous API calls, we won't dive into that.  

Btw @pombredanne  mentioned in the chat about having  some code for obtaining data from APIs provided by  package managers.

vulnerabilities/importers/safety_db.py
```diff
@@ -0,0 +1,170 @@
...
+            except HTTPError:  # PyPi does not have data about this package
+                pass
+
+            self.cache[package_name] = releases
+
+        return self.cache[package_name]
+
+
+@dataclasses.dataclass
+class DebianConfiguration(DataSourceConfiguration):
```
sbs2001, line 86: This seems rather out of place and unused. WIP ?

haikoschol: yep :)

vulnerabilities/migrations/0003_alpine_importer.py
```diff
@@ -0,0 +1,55 @@
...
+    Importer = apps.get_model('vulnerabilities', 'Importer')
+    qs = Importer.objects.filter(name='alpine')
+    if qs:
+        qs[0].delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('vulnerabilities', '0002_rust_importer'),
```
sbs2001, line 50: Not a migration expert here, but why is the `000X_ABCD_importer` a dependency to each other in all these migration scripts, any purpose?

sbs2001:
Don't we need a migration script to have an Importer for `safety_db` , since you have implemented new importer for it ?

vulnerabilities/migrations/0003_alpine_importer.py
haikoschol: I tried avoiding that but the Django DB migration implementation insists on having a simple parent/child relationship. I guess otherwise it gets tricky/ambiguous with rollbacks and such. Would have been nicer to have all these "importer adding migrations" independent of each other.

haikoschol:
> Don't we need a migration script to have an Importer for `safety_db` , since you have implemented new importer for it ?

We do and it seems I missed that. Good catch!

> Bump the Python 3.6 to Python 3.8 in the README :)

Also good catch! :)

PR #153, opened 2020-03-03
Add  importer for safetydb

sbs2001:

This data is imported from https://github.com/pyupio/safety-db .

### DIFF
--- a/vulnerabilities/scraper/safety_db.py
+++ b/vulnerabilities/scraper/safety_db.py
@@ -30,13 +30,12 @@ def import_vulnerabilities():
     vulnerability_package_dicts = []
     data = get_data_file()
     cves = set()
-    for package in data:
-        package_name = package
+    for package_name in data:
         all_package_versions = set(get_all_versions_of_package(package_name))
         if len(all_package_versions) == 0:
             # PyPi does not have data about this package, we skip these
             continue
-        for advisory in data[package]:
+        for advisory in data[package_name]:
             description = advisory['advisory']
             cve_id = advisory.get('cve')
             vuln_id = advisory['id']
--- a/vulnerabilities/scraper/safety_db.py
+++ b/vulnerabilities/scraper/safety_db.py
@@ -15,13 +15,15 @@ def get_data_file():
 
 def get_all_versions_of_package(package_name):
     url = "https://pypi.org/pypi/{}/json".format(package_name)
+    releases = set()
     try:
         with urlopen(url) as response:
             json_file = json.load(response)
     except HTTPError:  # PyPi does not have data about this package, we skip these
-        return set()
+        return releases
     for release in json_file['releases']:
-        yield release
+        releases.add(release)
+    return releases
 
 
 def import_vulnerabilities():
@@ -36,7 +38,7 @@ def import_vulnerabilities():
             continue
         for advisory in data[package]:
             description = advisory['advisory']
-            cve_id = advisory['cve'] if advisory['cve'] else None
+            cve_id = advisory.get('cve')
             vuln_id = advisory['id']
             vuln_version_ranges = advisory['specs']
             for vuln_version_range in vuln_version_ranges:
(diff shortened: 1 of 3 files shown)


pombredanne:
@Jwomers ping!
Do you have any intention to have an open source license with the NC for pyup data? https://github.com/pyupio/safety-db/blob/b873cceee96c04c4e9df93d17087592e94290482/LICENSE.txt
@sbs2001 thank ++ for this. This is useful to have but we cannot aggregate the data if this is precludes commercial use (whatever non commercial means is vague enough) . We can still provide the tool, but we will NOT run it until this is resolved.

vulnerabilities/scraper/safety_db.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_all_versions_of_package(package_name):
+    url = "https://pypi.org/pypi/{}/json".format(package_name)
+    try:
+        with urlopen(url) as response:
+            json_file = json.load(response)
+    except HTTPError:  # PyPi does not have data about this package, we skip these
+        return set()
+    for release in json_file['releases']:
+        yield release
```
haikoschol: This function should either be a generator or a regular function returning a set. And in this case it should return a set since you turn the result into a set immediately anyway at the callsite in `import_vulnerabilties()`.

vulnerabilities/scraper/safety_db.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def import_vulnerabilities():
+    vulnerability_package_dicts = []
+    data = get_data_file()
+    cves = set()
+    for package in data:
+        package_name = package
```
haikoschol: Why not call the variable `package_name` right away (in the for loop) if you think this is more descriptive?

vulnerabilities/scraper/safety_db.py
```diff
@@ -0,0 +1,60 @@
...
+        all_package_versions = set(get_all_versions_of_package(package_name))
+        if len(all_package_versions) == 0:
+            # PyPi does not have data about this package, we skip these
+            continue
+        for advisory in data[package]:
+            description = advisory['advisory']
+            cve_id = advisory['cve'] if advisory['cve'] else None
```
haikoschol: Should be `advisory.get('cve')`, since it will crash if the key `cve` is not present in the dictionary.

vulnerabilities/scraper/safety_db.py
```diff
@@ -0,0 +1,60 @@
...
+            vuln_id = advisory['id']
+            vuln_version_ranges = advisory['specs']
+            for vuln_version_range in vuln_version_ranges:
+                version_range = RangeSpecifier(vuln_version_range)
+                affected_versions = set()
+                for version in all_package_versions:
+                    if version in version_range:
+                        affected_versions.add(version)
+            unaffected_versions = all_package_versions - affected_versions
+            if cve_id not in cves:
```
haikoschol: If two packages are affected by the same CVE, doesn't this check cause the second one to be skipped entirely? And didn't this issue come up before somewhere? It seems familiar...

vulnerabilities/scraper/safety_db.py
sbs2001: Sorry for the repeated mistake

vulnerabilities/scraper/safety_db.py
sbs2001: You are correct, I remember vaguely, that the code was violating constraints which was temporarily fixed by removing the `summary`. We need some concrete way to deal with this. Some pointers to duplicate(not exactly) data https://github.com/pyupio/safety-db/blob/b873cceee96c04c4e9df93d17087592e94290482/data/insecure_full.json#L247  and   https://github.com/pyupio/safety-db/blob/b873cceee96c04c4e9df93d17087592e94290482/data/insecure_full.json#L256

haikoschol: Let's merge it as is and refine it later.

Issue #162, opened 2020-03-13
NPM importer is down


sbs2001:
```
>>> npm.scrape_vulnerabilities()

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspace/vulnerablecode/vulnerabilities/scraper/npm.py", line 120, in scrape_vulnerabilities
    response = json.load(urlopen(cururl))
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 641, in http_response
    'http', request, response, code, msg, hdrs)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 503: Service Unavailable
```

sbs2001:
Apparently, the importer is stuck on first page i.e it doesn't move on to next pages. It is stuck on 'https://registry.npmjs.org/-/npm/v1/security/advisories?page=0'.

haikoschol:
> Apparently, the importer is stuck on first page i.e it doesn't move on to next pages. It is stuck on 'https://registry.npmjs.org/-/npm/v1/security/advisories?page=0'.

How did you come to this conclusion? In the traceback you posted the NPM registry was unavailable. Did you try again and then conclude that it was stuck on the first page?

The URL to the next page is still included in the JSON returned from the registry:
```
$ curl -s "https://registry.npmjs.org/-/npm/v1/security/advisories?page=0" | jq .urls.next
"/-/npm/v1/security/advisories?page=1"
```

And I just ran the NPM import successfully on my machine. Looks like this was just a temporary outage of the registry.

sbs2001:
[...]

PR #163, opened 2020-03-13
Fix npm importer

sbs2001:
This fixes https://github.com/nexB/vulnerablecode/issues/162

Also made some changes to npm importer, like replacing semantic_version with dephell_specifier. The data_dump is modified such that now we are able to construct purls from the data obtained from the npm importer.

### DIFF
--- a/vulnerabilities/scraper/npm.py
+++ b/vulnerabilities/scraper/npm.py
@@ -57,7 +57,7 @@ def extract_versions(package_name, aff_version_range, fixed_version_range):
     aff_spec = RangeSpecifier(aff_version_range)
     fix_spec = RangeSpecifier(fixed_version_range)
     all_ver = get_all_versions(package_name)
-    if all_ver == []:
+    if not all_ver:
         # NPM registry has no data regarding this package, we skip these
         return ([], [])
     aff_ver = []
@@ -89,7 +89,7 @@ def extract_data(JSON):
             obj.get('vulnerable_versions', ''),
             obj.get('patched_versions', '')
         )
-        if affected_versions == fixed_versions == []:
+        if not affected_versions and not fixed_versions:
             continue
             # NPM registry has no data regarding this package finally we skip these
 
--- a/vulnerabilities/scraper/npm.py
+++ b/vulnerabilities/scraper/npm.py
@@ -31,6 +31,7 @@ from itertools import count
 NPM_URL = 'https://registry.npmjs.org{}'
 PAGE = '/-/npm/v1/security/advisories?'
 
+
 def get_all_versions(package_name):
     """
     Returns all versions available for a module
@@ -42,8 +43,9 @@ def get_all_versions(package_name):
             data = json.load(response)
     except HTTPError:
         return []
-        #NPM registry has no data regarding this package, we skip these
+        # NPM registry has no data regarding this package, we skip these
     return [v for v in data.get('versions', {})]
+
 
 def extract_versions(package_name, aff_version_range, fixed_version_range):
     """
@@ -53,9 +55,9 @@ def extract_versions(package_name, aff_version_range, fixed_version_range):
     aff_spec = RangeSpecifier(aff_version_range)
     fix_spec = RangeSpecifier(fixed_version_range)
     all_ver = get_all_versions(package_name)
-    if all_ver == [] :
-        #NPM registry has no data regarding this package, we skip these
-        return ([],[])
+    if all_ver == []:
+        # NPM registry has no data regarding this package, we skip these
+        return ([], [])
     aff_ver = []
     fix_ver = []
     # Unaffected version is that version which  is in the fixed_version_range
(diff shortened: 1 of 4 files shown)


vulnerabilities/scraper/npm.py
```diff
@@ -113,12 +107,15 @@ def scrape_vulnerabilities():
...
     Extract JSON From NPM registry
     """
-    nextpage = PAGE
     package_vulnerabilities = []
-    while nextpage:
-        cururl = NPM_URL.format(nextpage)
-        response = json.load(urlopen(cururl))
-        package_vulnerabilities.extend(extract_data(response))
-        next_page = response.get('urls', {}).get('next')
+    for page_num in count():
```
haikoschol: As I mentioned [on the ticket](https://github.com/nexB/vulnerablecode/issues/162#issuecomment-599063793), I don't think the error you encountered is fixed by this change. It was just a temporary issue of the registry backend.
IMHO the previous approach of following the URL to the next page included in the JSON response is preferrable so I propose we stick with that.

haikoschol: Turns out it was not an issue of the registry and there has been a bug in the NPM importer all along. I would still like to stick with the current approach and just fix the bug: rename `next_page` to `nextpage` (or the other way around).

sbs2001: It's not the only issue being fixed ,  for eg  if I change `next_page`   to `nextpage` (that was my initial approach btw)  one of the issue was this

```
>>> npm.scrape_vulnerabilities()
  File "/workspace/vulnerablecode/vulnerabilities/scraper/npm.py", line 122, in scrape_vulnerabilities
    package_vulnerabilities.extend(extract_data(response))
  File "/workspace/vulnerablecode/vulnerabilities/scraper/npm.py", line 97, in extract_data
    obj.get('patched_versions', '')
  File "/workspace/vulnerablecode/vulnerabilities/scraper/npm.py", line 67, in extract_versions
    all_ver = get_all_versions(package_name)
  File "/workspace/vulnerablecode/vulnerabilities/scraper/npm.py", line 55, in get_all_versions
    with urlopen(package_url) as response:
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 641, in http_response
    'http', request, response, code, msg, hdrs)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/home/gitpod/.pyenv/versions/3.7.6/lib/python3.7/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```
This is due to a package called 'panellum' whose data the registry doesnt have.
It's located at  https://registry.npmjs.org/-/npm/v1/security/advisories?page=127

haikoschol: But that issue has a different cause and is handled in `get_all_versions()`.

I do want to merge that fix and the change to using `RangeSpecifier`. All I'm asking is that you fix this specific bug that caused an endless loop in a different way. The argument for that is, since the response from NPM includes a URL to the next page we should use that instead of making assumptions about the URL (even if they seem obviously correct).

sbs2001: Yeah, that approach is better

vulnerabilities/scraper/npm.py
```diff
@@ -62,20 +52,21 @@ def extract_versions(package_name, aff_version_range, fixed_version_range):
     Seperate list of affected versions and unaffected versions from all versions
     using the ranges specified.
     """
-    aff_spec = semantic_version.NpmSpec(remove_spaces(aff_version_range))
-    fix_spec = semantic_version.NpmSpec(remove_spaces(fixed_version_range))
+    aff_spec = RangeSpecifier(aff_version_range)
+    fix_spec = RangeSpecifier(fixed_version_range)
     all_ver = get_all_versions(package_name)
+    if all_ver == []:
```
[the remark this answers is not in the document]
sbs2001: Are you sure about this?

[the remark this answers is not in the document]
haikoschol: Yes, unless I'm missing some broader context where the difference between an empty list and `None` matters.

```
In [1]: foo = []                                                                                         

In [2]: if not foo: print('bar')                                                                         
bar
```

sbs2001:
Changes are done

PR #166, opened 2020-03-18
Allow skipping internet dependent tests, fix npm dump

sbs2001:
This is done by adding pytest markers on the tests which require internet connections , thus they can be skipped if necessary easily.

Simply run `DJANGO_DEV=1 pytest  -m "not webtest"`


sbs2001:
npm dump was missing 'type' field for unaffected packages , this is fixed

Issue #62, opened 2019-09-25
Collect SUSE


haikoschol:
https://www.suse.com/support/security/cvrf/

pombredanne:
Here are a few candidate CVRF libraries and pointers that may be of some help:
- @mprpic https://github.com/mprpic/advisory-parser (which has quite a few interesting formats beyond CVRF)
- @mschiffm https://github.com/mschiffm/cvrfparse
- https://github.com/oasis-open/csaf-parser
- https://github.com/sdrees/cvrf-util
- https://github.com/efrenfuentes/cvrfparse

And articles or specs:
  - https://blogs.cisco.com/security/tools-of-the-trade-cvrfparse
  - https://blogs.cisco.com/security/the-missing-manual-cvrf-1-1-part-1-of-2
  - https://blogs.cisco.com/security/the-missing-manual-cvrf-1-1-part-2-of-2
  - https://blogs.cisco.com/security/cvrf-1-2-2
  - https://github.com/CVRF
  - https://github.com/oasis-tcs/csaf 
  - https://github.com/oasis-open/csaf-documentation
  - https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html

pombredanne:
Also, not sure this is all up to date but we have https://support.novell.com/security/oval/suse.linux.enterprise.server.11.xml

sbs2001:
This is  the total data which I am able to get 

```
Note_Notes :::: Title:Vulnerability Description|Type:General|Ordinal:1|{http://www.w3.org/XML/1998/namespace}lang:en|Systems with microprocessors utilizing speculative execution and indirect branch prediction may allow unauthorized disclosure of information to an attacker with local user access via a side-channel analysis.
ProductID :::: openSUSE Leap 42.3:ucode-intel-debuginfo-20170707-13.1
CVE_Vulnerability :::: CVE-2017-5715
Status_ProductStatuses :::: Type:Fixed
Threat_Threats :::: Type:Impact|Description:critical
BaseScore_ScoreSet :::: 4.7
Vector_ScoreSet :::: AV:L/AC:M/Au:N/C:C/I:N/A:N
Remediation_Remediations :::: Type:Vendor Fix|URL:http://lists.opensuse.org/opensuse-security-announce/2018-01/msg00015.html
Description_Remediation :::: {http://www.w3.org/XML/1998/namespace}lang:en|Please Install the update.
[...]

Issue #44, opened 2019-04-07
Collect Red Hat RHSA


pombredanne:
https://www.redhat.com/security/data/oval/
http://www.redhat.com/security/data/oval/com.redhat.rhsa-all.xml
https://www.redhat.com/security/data/metrics/

See also the API such as at https://access.redhat.com/hydra/rest/securitydata/cvrf.json
https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/index and https://access.redhat.com/articles/221883

haikoschol:
Or maybe CVRF? https://www.redhat.com/security/data/cvrf/

pombredanne:
There is also this http://www.redhat.com/security/data/oval/com.redhat.rhsa-all.xml

haikoschol:
Comparing OVAL and CVRF entries for the same RHSA, it seems that CVRF has all the information we need in an easier to consume structure. However, the XML version of it has more data than the JSON linked above.

pombredanne:
Actually we could also start with the simpler https://www.redhat.com/security/data/metrics/rpm-to-cve.xml

sbs2001:
I didn't see this mentioned, so here it is https://access.redhat.com/hydra/rest/securitydata/cve.json

Anyways the larger  issue with extracting data from any of the above mentioned sources (and CVRF advisories in general) is that there is no explicit mention of package name and package version . Rather these are combined and hence we will need to split these combined names into package name and version to construct purl.

pombredanne:
I am reopening as we are missing Oval and/or CVRF data for completeness. Or at least we need to double check 
https://www.redhat.com/security/data/oval/v2/
The license is not clear though

pombredanne:
Based on the feedback of a RedHat buddy:
> The license for what's under https://www.redhat.com/security/data/oval/v2/ is CC BY 4.0 as
explained here: https://access.redhat.com/security/data

Issue #41, opened 2019-04-04
Add support for Microsoft vulnerabilities  and CVRF


pombredanne:
MSFT releases their vulnerabilities using this CVRF format https://www.icasi.org/cvrf/

See https://github.com/Microsoft/MSRC-Microsoft-Security-Updates-API for details 

@mschiffm https://github.com/mschiffm/cvrfparse is a library to likely handle this alright

pombredanne:
See also the updated pointers for CVRF in https://github.com/nexB/vulnerablecode/issues/62#issuecomment-535420939

sbs2001:
These are closed source vulnerabilities, isn't that out of scope of vulnerablecode ?

haikoschol:
Microsoft ships quite a bit of OSS code these days. Having said that, I don't know whether this feed makes sense for VulnerableCode. Someone needs to do the research on it. :)

sbs2001:
The data from the CVRF is basically https://portal.msrc.microsoft.com/en-us/security-guidance , majority of the vulnerabilities are of closed source Microsoft products like IE, Paint ,Windows etc. The only value these have are about .net vulnerabilities.

pombredanne:
There is value in these after all. See ttps://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2022-41032
It has unique information on affected NuGet versions that are not available elsewhere.

PR #173, opened 2020-04-15
[WIP]Add cvrf  parser and import opensuse cvrf advisories

sbs2001:
The cvrf parser is obtained by forking https://github.com/oasis-open/csaf-parser and trimming it down . 

TODOs:

- [ ] Trim the parser further. This includes removing the cli funtionality of parser entirely,rather  than giving it fixed inputs and removing support for other schemas.

img

- [ ] Handle the licensing. Should I add LICENSE from https://github.com/oasis-open/csaf-parser/blob/master/LICENSE to the parser directory or do something else? , @pombredanne  help please

- [ ]  Tests! Here is some sample data which was successfully dumped in db. 

[sampledata.zip](https://github.com/nexB/vulnerablecode/files/4482017/sampledata.zip)

I also tried this parser to get data from Red hat's cvrfs and it works, but that's not in the scope of this PR for _now._

Fixes https://github.com/nexB/vulnerablecode/issues/62, addresses https://github.com/nexB/vulnerablecode/issues/44 and https://github.com/nexB/vulnerablecode/issues/41


sbs2001:
https://colab.research.google.com/drive/1eLKwA5oe6pfCtKBqP672AQmWITmAKG9s here's google colab notebook to see the importer running

sbs2001:
I've realized that OVAL advisories are better than CVRF advisories, because of following reason

 OVAL advisories have a concrete way of denoting a package's name and it's version. On the other hand CVRF advisories publish package with it's name and version in one string. So we have to do some guesswork(which no matter how accurate, is bad) to figure out what is package name and what is package version. 

I would rather prefer to import data from SUSE OVAL advisory than CVRF advisories. I'll leave this open in case we might need to handle CVRF advsiories(in case no other formats are available)

PR #174, opened 2020-04-20
Add schema validator for Alpine linux

sbs2001:

Continuation of implementation schema  validators.


sbs2001:
Black check seems to like double quotes more:)

Issue #59, opened 2019-09-25
Use OVAL export for Ubuntu vulnerabilities source


haikoschol:
The current data source for Ubuntu vulnerability information is [an HTML page](https://people.canonical.com/~ubuntu-security/cve/main.html). Apart from the fact that this is not really meant to be machine-readable, it also does not include version information for affected/fixed packages.
Looking at https://github.com/coreos/clair/issues/804, it seems that the source we want to use are XML files in OVAL format available at https://people.canonical.com/~ubuntu-security/oval/.

mjherzog:
fyi...https://security.stackexchange.com/questions/134269/how-to-read-cve-2016-5696-correctly

haikoschol:
@mjherzog thanks. So I guess you are saying it is possible to determine the affected/fixed versions by following links to vulnerabilities and parsing more HTML. I'd still rather use the OVAL data if possible.

mjherzog:
@haikoschol I was just sharing some background information to help me understand the HTML page. I am not suggesting that we should parse the HTML - OVAL sounds better.

PR #188, opened 2020-06-02
Ubuntu oval rewrite

sbs2001:
Use ubuntu oval files which are machine readable rather than importing data from HTML ubuntu security tracker. Also experiment with storing ETags to avoid doing useless work .

Fixes https://github.com/nexB/vulnerablecode/issues/59

- [x] Implement `OvalDataSource`

- [x] Store and utilize etags to avoid doing work on already imported data. 

- [x] Add tests to ensure etags work as expected


vulnerabilities/data_source.py
```diff
@@ -380,3 +390,127 @@ def _include_file(
...
+        a tuple of file's metadata and it's ET.ElementTree.
+        Subclasses must implement this method.
+
+        Note: Mapping MUST INCLUDE "type" key. Example values of Mapping
+              {"type":"deb","qualifiers":{"distro":"buster"} }
+
+        """
+        raise NotImplementedError
+
+    def added_advisories(self) -> List[Advisory]:
```
haikoschol, line 434: `added_advisories()` *must* only return advisories that have been added since the last run of the importer. All other advisories should be returned by `updated_advisories()`. Is this somehow ensured elsewhere in your implementation?

vulnerabilities/importers/ubuntu.py
```diff
@@ -21,36 +20,112 @@
...
+            )
+        # In case every file is latest, _fetch won't yield anything(due to checking for new etags),
+        # this would return None to added_advisories
+        # which will cause error, hence this
+        # function return an empty list
+        return []
 
-UBUNTU_ROOT_URL = 'https://people.canonical.com/~ubuntu-security/cve/main.html'
+    def set_api(self, packages):
+        asyncio.run(self.pkg_manager_api.load_api(packages))
```
haikoschol, line 81: I'm not sure `asnycio.run()` is the right choice here. The [documentation](https://docs.python.org/3.8/library/asyncio-task.html#asyncio.run) says: "This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once."
But I don't know much about Python asyncio.

vulnerabilities/data_source.py
sbs2001: Oops, I forgot to change it back to `updated_advisories`
No the implementation, doesn't do that. I will change this to `updated_advisories`, I used `added_advisories` for fast testing because it uses bulk inserts.

haikoschol: I see. From a quick glance at the OVAL test data, it seems advisories have an attribute `public_date`. Maybe we can use that to tell added and updated advisories apart. But that can be worked on in a separate PR.

vulnerabilities/importers/ubuntu.py
sbs2001: >It should be used as a main entry point for asyncio programs.

That's what it's precisely doing in this case. 

As for 
>Should ideally only be called once
 
They also mention, 
>This function cannot be called when another asyncio event loop is running in the same thread.

Which IMHO, is fine with the implementation, because `self.pkg_manager_api.load_api(packages)` doesn't lead to creation of other event loops i.e if you follow the function  `self.pkg_manager_api.load_api(packages)`  nowhere it creates an event loop (via for eg. `asyncio.run`).

I wouldn't worry about this.

vulnerabilities/data_source.py
sbs2001: No, they are for a  whole  xml oval document, which results into creation of 1000s of `Advisory` objects. No way to know which `Advisory` is changed.

sbs2001: Oh, I see what you mean,  apparently Ubuntu OVAL files provide `public_date` , seems vendor specific since SUSE OVAL files don't contain this. 

This is good news, I agree with you let's first get this merged and I'll iterate upon it in other small PR.

PR #191, opened 2020-06-06
Make the config of Importer  'updateable'

sbs2001:
This will enable the us to update the config of an Importer after each run.  Essential to utilise Etags, and for https://github.com/nexB/vulnerablecode/pull/188

### DIFF
--- a/vulnerabilities/data_source.py
+++ b/vulnerabilities/data_source.py
@@ -75,7 +75,7 @@ class InvalidConfigurationError(Exception):
 
 @dataclasses.dataclass
 class DataSourceConfiguration:
-    batch_size: int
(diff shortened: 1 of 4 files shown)


vulnerabilities/data_source.py
```diff
@@ -75,7 +75,8 @@ class InvalidConfigurationError(Exception):
 
 @dataclasses.dataclass
 class DataSourceConfiguration:
-    batch_size: int
+    # batch_size: int
```
haikoschol, line 78: please remove the line instead of commenting it out

sbs2001: Done

Issue #74, opened 2019-09-26
Collect vulnerabilities from GitHub


pombredanne:
There are a few levels to consider:
1. GH provides some security feeds for their own GHE appliances
2. the projects hosted can publish their own feeds
3. a project dependencies can have vulnerabilities reported by GH

pombredanne:
See https://github.com/gradle/gradle/security/advisories/GHSA-4cwg-f7qc-6r95 for an example

sbs2001:
GitHub provides a GraphQL based API for these, we could collect this data from this API. As this API requires OAuth token, we need some mechanism to allow the user to specify  the token  (environmental variable). 

An example query at https://developer.github.com/v4/explorer/ to obtain this data would be :
```

query MyQuery {
  securityVulnerabilities(first: 10) {
    nodes {
      package {
        ecosystem
        name
      }
      vulnerableVersionRange
      advisory {
        description
        ghsaId
      }
    }
  }
}

```

sbs2001:
Fixed by https://github.com/nexB/vulnerablecode/pull/204

PR #204, opened 2020-06-15
Add GitHub API importer

sbs2001:
TODOs

- [x] Add Nuget, Maven and Composer package manager's version API

- [x]  Add tests for everything 

I will file another separate PR to move all these VersionAPI classes to one separate module.


vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
```
pombredanne, line 1: Nit: best to remove the date IMHO here or add 2020

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+                if resp.get('message') == 'Bad credentials':
+                    raise GitHubTokenError('Invalid GitHub token')
+
+                end_cursor = resp['data']['securityVulnerabilities']['pageInfo'][
+                    'endCursor'
+                ]
+                end_cursor_exp = 'after: {}'.format('"{}"'.format(end_cursor))
+                api_data[ecosystem].append(resp)
+                print(resp)
```
pombredanne, line 125: This may be a left over debug print?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+                if not resp['data']['securityVulnerabilities']['pageInfo'][
+                    'hasNextPage'
```
pombredanne, line 128: why the line breaks?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if ecosystem == 'MAVEN':
+            self.version_api = MavenVersionAPI()
+
+        elif ecosystem == 'NUGET':
+            self.version_api = NugetVersionAPI()
+
+        elif ecosystem == 'COMPOSER':
+            self.version_api = ComposerVersionAPI()
+
```
pombredanne, line 143: What about this instead?
```
    def set_version_api(self, ecosystem: str) -> None:
        versioners = dict(
            MAVEN=MavenVersionAPI,
            NUGET=NugetVersionAPI,
            COMPOSER=ComposerVersionAPI,
        )
        versioner = self.get_versioner(ecosystem)
        if versioner:
            self.version_api = versioner()
```
Using a mapping for long conditionals is a useful idiom :)  .... and here I expect that list to be growing, right?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+    @staticmethod
+    def process_name(
+        ecosystem: str, pkg_name: str
+    ) -> Optional[Tuple[Optional[str], str]]:
```
pombredanne, line 147: why not put everything on one line?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if ecosystem == 'MAVEN':
+
```
pombredanne, line 150: nit: no blank needed here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+                        )
+        return adv_list
+
+    @staticmethod
+    def categorize_versions(
+        version_range: str, all_versions: Set[str]
+    ) -> Tuple[Set[str], Set[str]]:
+        version_range = RangeSpecifier(version_range)
+        affected_versions = {
+            version for version in all_versions if version in version_range
```
pombredanne, line 224: nit: no need for line wrapping here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if pkg_name in self.cache:
+            return
+
+        artifact_comps = pkg_name.split(':')
+        endpoint = self.artifact_url(artifact_comps)
+        resp = requests.get(endpoint).content
+
+        try:
+
```
pombredanne, line 246: nit: no blank needed here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+            xml_resp = ET.ElementTree(ET.fromstring(resp.decode('utf-8')))
+            self.cache[pkg_name] = self.extract_versions(xml_resp)
+
```
pombredanne, line 249: nit: no blank needed here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        except ET.ParseError:
+            self.cache[pkg_name] = set()
+
+    @staticmethod
+    def artifact_url(artifact_comps: List[str]) -> str:
+
```
pombredanne, line 255: nit: no blank needed here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        group_id, artifact_id = artifact_comps
+        group_url = group_id.replace('.', '/')
+        suffix = group_url + '/' + artifact_id + '/' + 'maven-metadata.xml'
+        endpoint = base_url.format(suffix)
+
+        return endpoint
+
+    @staticmethod
+    def extract_versions(xml_response: ET.ElementTree) -> Set[str]:
+
```
pombredanne, line 266: nit: no blank needed here

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        self.cache[pkg_name.lower()] = self.extract_versions(resp)
+
+    @staticmethod
+    def nuget_url(pkg_name: str) -> str:
+        base_url = 'https://api.nuget.org/v3/registration5-semver1/{}/index.json'
+        return base_url.format(pkg_name.lower())
+
+    @staticmethod
+    def extract_versions(json_resp: dict) -> Set[str]:
+        all_versions = set()
```
pombredanne, line 303: Why a set vs. a list? do you expect duplicates? and would using a set means that any original order may be lost?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+
+        all_versions = set()
+        for child in xml_response.getroot().iter():
+            if child.tag == 'version':
+                all_versions.add(child.text)
+
+        return all_versions
+
+
+class NugetVersionAPI:
```
pombredanne, line 275: Is there some code share between the various "VersionAPI" classes? What about moving the common parts in a super class? or some helper?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        try:
+
+            xml_resp = ET.ElementTree(ET.fromstring(resp.decode('utf-8')))
+            self.cache[pkg_name] = self.extract_versions(xml_resp)
+
+        except ET.ParseError:
+            self.cache[pkg_name] = set()
+
+    @staticmethod
+    def artifact_url(artifact_comps: List[str]) -> str:
```
pombredanne, line 254: You should see if that would be useful to reuse either scancode, or at least pymaven here... may not worth for a single function, but at least this may be worth sharing?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        group_url = group_id.replace('.', '/')
+        suffix = group_url + '/' + artifact_id + '/' + 'maven-metadata.xml'
+        endpoint = base_url.format(suffix)
+
+        return endpoint
+
+    @staticmethod
+    def extract_versions(xml_response: ET.ElementTree) -> Set[str]:
+
+        all_versions = set()
```
pombredanne, line 267: Why a set? Would the order of versions matter?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        base_url = 'https://api.nuget.org/v3/registration5-semver1/{}/index.json'
+        return base_url.format(pkg_name.lower())
+
+    @staticmethod
+    def extract_versions(json_resp: dict) -> Set[str]:
+        all_versions = set()
+        try:
+            for entry in json_resp['items'][0]['items']:
+                all_versions.add(entry['catalogEntry']['version'])
+        # json response for YamlDotNet.Signed triggers this exception
```
pombredanne, line 307: What does this means exactly?What about using a cascade of get() instead to understand exactly why there are these key errors? Can you show me the problematic JSON?

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        except KeyError:
+            return all_versions
```
pombredanne, line 309: A `pass` would be clearer... BUT the reason for swallowing the KeyError is what I would like to understand

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+    def load_to_api(self, pkg_name: str) -> None:
+        if pkg_name in self.cache:
+            return
+        endpoint = self.composer_url(pkg_name)
+        json_resp = requests.get(endpoint).json()
+        self.cache[pkg_name] = self.extract_versions(json_resp, pkg_name)
+
+    @staticmethod
+    def composer_url(pkg_name: str) -> str:
+        vendor, name = pkg_name.split('/')
```
pombredanne, line 330: always prefer a more robust partition() or rpartition():
```
>>> vendor, name = pkg_name.split('/')
NameError: name 'pkg_name' is not defined
>>> vendor, _, name = 'foo'.partition('/')
>>> vendor, name
('foo', '')
>>> vendor, _, name = 'foo/bar'.partition('/')
>>> vendor, name
('foo', 'bar')
```

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,342 @@
...
+        return f'https://repo.packagist.org/p/{vendor}/{name}.json'
+
+    @staticmethod
+    def extract_versions(json_resp: dict, pkg_name: str) -> Set[str]:
+        all_versions = json_resp['packages'][pkg_name].keys()
```
pombredanne, line 335: What has this response data mapping anything to do with JSON at this stage of processing?

vulnerabilities/importers/github.py
sbs2001: No, actually check https://github.com/nexB/vulnerablecode/issues/74#issuecomment-642041958

vulnerabilities/importers/github.py
sbs2001: That will be done in another separate PR, because this is not just limited to  this importer, rather almost every other importer in `vulnerabilities/importers` has one such `VersionAPI` class thingie.

vulnerabilities/importers/github.py
sbs2001: my linter had  pep8 72 character line limit. Apparently we use 100 characters and this could've easily fitted into one line. I will make the changes(to the code and the linter). Thanks

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Your approach is better, I will use that instead

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: >do you expect duplicates?

No

> and would using a set means that any original order may be lost?

Yes, I use sets wherever  order doesn't matter, the rationale behind being I'd rather have guaranteed  unique elements at the cost of  losing the order which I don't care about here. 

If there are any significant performance gains(I will research about this) , I will switch to lists instead.

vulnerabilities/importers/github.py
sbs2001: I'm not sure I understand this correctly, 

Do you mean, like there is function like `artifact_url` in pymaven or scancode and I reuse that here instead ?

vulnerabilities/importers/github.py
sbs2001: Check my comment above, 

> Would the order of versions matter?

No

>Why a set?

Because order doesn't matter ;) .

vulnerabilities/importers/github.py
sbs2001: Renaming this `json_resp` to `resp`

vulnerabilities/importers/github.py
sbs2001: Good catch, removing this

vulnerabilities/importers/github.py
sbs2001: Done

vulnerabilities/importers/github.py
sbs2001: >Can you show me the problematic JSON?

This is raising that exception https://api.nuget.org/v3/registration5-semver1/yamldotnet.signed/index.json 
All others follow this schema instead https://api.nuget.org/v3/registration5-semver1/sustainsys.saml2/index.json 

>What about using a cascade of get() instead to understand exactly why there are these key errors? 

That can be done, but I thought catching an exception would be more 'pythonic' (EAFP).

vulnerabilities/importers/github.py
sbs2001: I have no idea how to handle https://api.nuget.org/v3/registration5-semver1/yamldotnet.signed/index.json which is causing the exception, hence the 'swallowing' of the KeyError

vulnerabilities/importers/github.py
sbs2001: >always prefer a more robust partition() or rpartition():

I didn't knew about this function. Thanks.
Robust in the sense it doesn't raise `ValueError: not enough values to unpack` .

 But in this case it makes more sense to me to use `split` because that would raise an exception and crash out VulnerableCode, instead of  ending up writing garbage in the DB.

singh1114:
LGTM!

vulnerabilities/importers/github.py
pombredanne: yes I think there is one. That's minor though

vulnerabilities/importers/github.py
pombredanne: The performance does not matter here, this is more about ordered and unique vs. unordered
I think that the order may matter though as this is the version orders?

vulnerabilities/importers/github.py
pombredanne: > That can be done, but I thought catching an exception would be more 'pythonic' (EAFP).

Well, the key is readability and clarity first: here when I read the code in `json_resp['items'][0]['items']` I wonder immediately. What if there is no items or this is not a mapping? What is [0]? Is this a list or a mapping with a key `0`? if this is a list, what about the other items in that list? and what is this second `items` about?

Here we could have something instead like that:
```
for item in json_resp.get('items', []):
    for entry in item.get('items', []):
        ...
```
When I read it I get these are mapping that return lists and I have no problem to understand the underlying data structure that I do not know about. Otherwise, I feel I need to do quite a bit of effort to read the code and it assumes that I know about the upstream JSON data structure before hand. So EAFP is fine, but it matters more to think about the readability of the code IMHO and how it will resist to future changes in the API and how that will make it easy to debug or not.



> This is raising that exception https://api.nuget.org/v3/registration5-semver1/yamldotnet.signed/index.json
> All others follow this schema instead https://api.nuget.org/v3/registration5-semver1/sustainsys.saml2/index.json

This is the kind of comment that is needed in the code :)

vulnerabilities/importers/github.py
sbs2001: Order doesn't matter here.

vulnerabilities/importers/github.py
sbs2001: I agree with everything you said. 

I am wondering what's the best answer in this context to the question you asked.

>how it will resist to future changes in the API and how that will make it easy to debug or not.

Because in this case, the end result in such an event, the outcome of both approaches  EAFP(try-except) and LBYL(get cascade)  is same.

pombredanne: 1. you can still keep inside your direct items access as `all_versions.add(entry['catalogEntry']['version'])` so if that part changes if will fail.
2. we should have tests too
3. we could think about a very limited subset of "canary"-like tests that are doing live queries on the web and fail if the API changes by returning different data.

sbs2001:
We will merge this after the PRs dealing with migrations and changes in data models are done with. It's really painful to refactor these.

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,318 @@
...
+
+        try:
+            xml_resp = ET.ElementTree(ET.fromstring(resp.decode("utf-8")))
+            self.cache[pkg_name] = self.extract_versions(xml_resp)
+        except ET.ParseError:
+            self.cache[pkg_name] = set()
+
+    @staticmethod
+    def artifact_url(artifact_comps: List[str]) -> str:
+        base_url = "https://repo.maven.apache.org/maven2/{}"
```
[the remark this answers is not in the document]
pombredanne: that's minor though and could wait

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,318 @@
...
+class GitHubAPIDataSource(DataSource):
+
+    CONFIG_CLASS = GitHubAPIDataSourceConfiguration
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        try:
+            self.gh_token = os.environ["GH_TOKEN"]
+        except KeyError:
+            raise GitHubTokenError("Envirnomental variable GH_TOKEN is missing")
```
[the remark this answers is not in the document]
sbs2001: Fixed

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,318 @@
...
+
+                query_json = {"query": query % (ecosystem, end_cursor_exp)}
+                resp = requests.post(self.config.endpoint, headers=headers, json=query_json).json()
+                print(resp)
+                if resp.get("message") == "Bad credentials":
+                    raise GitHubTokenError("Invalid GitHub token")
+
+                end_cursor = resp["data"]["securityVulnerabilities"]["pageInfo"]["endCursor"]
+                end_cursor_exp = "after: {}".format('"{}"'.format(end_cursor))
+                api_data[ecosystem].append(resp)
```
pombredanne, line 119: In general it is often better to yield rather than to accumulate in a list, FWIW

vulnerabilities/importers/github.py
pombredanne: ok

vulnerabilities/importers/github.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if ecosystem == "NUGET":
+            return None, pkg_name
+
+        if ecosystem == "COMPOSER":
+            vendor, name = pkg_name.split("/")
+            return vendor, name
+
+    def process_response(self) -> List[Advisory]:
+        adv_list = []
```
pombredanne, line 152: Same as above... yielding rather than returning a list is always a good option

PR #205, opened 2020-06-18
OpenSSL importer

### LINKED ISSUE #97: Collect OpenSSL
https://www.openssl.org/news/vulnerabilities.html and https://www.openssl.org/news/vulnerabilities.xml

edoardolanzini:
Importing vulnerability information from [https://www.openssl.org/news/vulnerabilities.xml](OpenSSL).

Fixes #97


vulnerabilities/importers/openssl.py
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
```
sbs2001, line 1: :nit Remove the `2017`, I know we have that same outdated license header everywhere, which need to be updated as well.

vulnerabilities/importers/openssl.py
```diff
@@ -0,0 +1,90 @@
...
+        raw_data = self.fetch()
+        advisories = self.to_advisories(raw_data)
+        return self.batch_advisories(advisories)
+
+    def fetch(self):
+        return requests.get("https://www.openssl.org/news/vulnerabilities.xml").content
+
+    @staticmethod
+    def to_advisories(xml_response: str) -> Set[Advisory]:
+        advisories = list()
```
sbs2001, line 46: :nit Since you've used `[]` everywhere else I'd prefer to have `[]` here as well instead of `list()`

vulnerabilities/importers/openssl.py
```diff
@@ -0,0 +1,90 @@
...
+                for info in element:
+                    if info.tag == 'cve':
+                        cve_id = 'CVE-' + info.attrib.get('name')
+                    if info.tag == 'affects':
+                        # Vulnerable package versions
+                        vuln_pkg_versions.append(info.attrib.get('version'))
+                    if info.tag == 'fixed':
+                        # Fixed package versions
+                        safe_pkg_versions.append(info.attrib.get('version'))
+                        if len(info) > 0:
```
sbs2001, line 66: :nit Use `if info` instead

vulnerabilities/importers/openssl.py
```diff
@@ -0,0 +1,90 @@
...
+                            commit_hash = info[0].attrib.get('hash')
```
sbs2001, line 67: :nit Why use `get` here? IMHO if `commit_hash=None` then the code will crash out below anyway.  If the importer works as it is now, IMHO it'd be fine with    `commit_hash = info[0].attrib['hash']`

vulnerabilities/importers/openssl.py
```diff
@@ -0,0 +1,90 @@
...
+        for element in root:
+            if element.tag == "issue":
+                cve_id = ''
+                summary = ''
+                safe_pkg_versions = []
+                vuln_pkg_versions = []
+                ref_urls = []
+                for info in element:
+                    if info.tag == 'cve':
+                        cve_id = 'CVE-' + info.attrib.get('name')
```
sbs2001, line 59: :nit Do a line break after each of these `if` blocks.

vulnerabilities/tests/test_openssl.py
```diff
@@ -0,0 +1,220 @@
...
+                cve_id='CVE-2019-1551')
+        ]
+        found_data = OpenSSLDataSource.to_advisories(data)
+
+        # Sort them by CVE-ID
+        found_data.sort(key=lambda x: x.cve_id)
+        expected_data.sort(key=lambda x: x.cve_id)
+
+        # Check first advisory
+        assert found_data[0].cve_id == expected_data[0].cve_id
```
sbs2001, line 206: question: Won't these be same as doing a single `assert found_data == expected_data` ? If yes do use  `assert found_data == expected_data`

vulnerabilities/importers/openssl.py
edoardolanzini: Should I change it to 2020 or just remove the year?

sbs2001:
LGMT ! Could you squash the commits and force push ? After squashing we will merge this once we get approval about `pkg_type='openssl'` from @pombredanne . Thanks again.

PR #211, opened 2020-06-20
Code cleanup.

sbs2001:
data_dump.py is finally obsolete, hence removing it.
Ubuntu now uses OVAL hence removing it's old test_data.
To achieve this some changes in pytest's conftest.py were made,
by removing old fixtures which were unused.


pombredanne:
The obvious question that you surely have already checked: is the data really the same?

sbs2001:
I'm pretty sure the old importer collected only a fraction of data, than the new importer. The biggest flaws with the old importer were 
 1. It just collected package names, vulnerabilities that's it, no version ranges, or versions 
 2. It used html
I will come up with a colab notebook, to compare the OVAL vs https://people.canonical.com/~ubuntu-security/cve/main.html scraper

sbs2001:
@pombredanne  , @haikoschol 
Here it is https://colab.research.google.com/drive/1kudxmE8r0_jt9tOEYIlpdbqNXfDhmoRU?usp=sharing , 
the gist of it is  , ( the numbers denote vulnerable packages.)

```
In HTML 340
In Oval 1036
Only in HTML 176
```
@pombredanne  I think it's  still better to delete this outdated code , it anyways is not compatible to new `Importer` and `DataSource` codebase. There's no value in refactoring the html importer into the new codebase because 

While preparing the notebook I found that you can do a  `git clone git://git.launchpad.net/ubuntu-cve-tracker`, I would rather import data from the git repository(which powers https://people.canonical.com/~ubuntu-security/cve/main.html), since we now have `GitDataSource` which is much much superior(it can differentiate between added and changed vulnerabilities). 

Btw that git repository seems to contain many many other goodies, I'd take a loot at it afterwards.

sbs2001:
Check https://git.launchpad.net/ubuntu-cve-tracker/tree/scripts/oval_lib.py#n58, it generates oval files from the same data, IMHO it ignores packages which are  part of eol releases or vulnerabilities which are not yet confirmed/triaged check https://git.launchpad.net/ubuntu-cve-tracker/tree/scripts/generate-oval#n101 . Again this  repo has so many goodies.

haikoschol:
lgtm

PR #214, opened 2020-06-22
Add docs directory and add tutorial about adding new importer

sbs2001:
A start to adding documentation

### DIFF
--- a/docs/how-to-guides/add_new_importer.rst
+++ b/docs/how-to-guides/add_new_importer.rst
@@ -1,0 +1,199 @@
+.. raw:: html
+
+   <h1 align="center">
+
+Adding a new Importer to VulnerableCode
+
+.. raw:: html
+
+   </h1>
+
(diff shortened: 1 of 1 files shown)


haikoschol:
Is all this "raw html" stuff really necessary?

sbs2001:
@haikoschol re
>Is all this "raw html" stuff really necessary?


I used https://github.com/jgm/pandoc to convert the markdown gist to rst, which converted the inline html from the markdown to 'raw html blocks' . 

I can get rid of the blocks, but there seems no rst native way to center things.

haikoschol:
> I used https://github.com/jgm/pandoc to convert the markdown gist to rst, which converted the inline html from the markdown to 'raw html blocks' .
> 
> I can get rid of the blocks, but there seems no rst native way to center things.

I see. Was just curious why it's in there. I don't mind keeping it.

Issue #157, opened 2020-03-05
Change database schema to accommodate different types of severity indicators


sbs2001:
Repaste   from gitter : 

> @sbs2001
> Mar 04 14:53
> Btw, I wanted to start discussion regarding storage of severity indicators: The problem is different advisories provide severity scores in different formats, for example cvss v1,cvss v2,cvss v3,cvss v3.1 or just textual description(High,Low,Medium). I can think of 2 approaches:
 1. Make separate columns for each type of severity indicator for model Vulnerability.
     OR
 
 2. cvss scores of different versions can be interconverted by manipulating the attack vector, this adds some error (upto 5%). This way we will have 2 columns to indicate severity one for cvss score of the latest standard and other for textual representation of severity.
     What are your opinions?

sbs2001:
Another approach I can think of is having 2 columns for cvss's , 1st column will store the version of cvss and 2nd column will store the score

pombredanne:
What about having something  tad different?
I do not want us to be limited to CVSS which is a rather complex system and we know tat there can be several interpretation of them for the same vulnerability.

IMHO we should have instead three data points for each score:
- like for a reference, some source URL of sorts where that score was published
- a type of score (cvss, etc)
- the score value itself.

And since there can be many of these, we make that a model of its own.
This is a tad heavier but this is the correct model IMHO.
It could be toned down to a single JSON field, but that does not help much with queries

pombredanne:
A third an likely cleaner approach is based on the facts that we know that:
1. there are multiple scoring mechanisms aka. "scoring type" (some formal cvssversion, or informal high/medium/low)
2. for a "scoring type" there can be different interpretations from different "sources" leading to different "score"

So IMHO you could get these "type, source, score" in a Score table that has has a foreign key to a Vulnerability much like a Reference of sorts.

Single JSON columns get messy quickly and are best avoided, and limiting ourselves to the CVSS scores is not enough IMHO.

sbs2001:
@pombredanne  I think the new  `VulnerabilityReference`  from https://github.com/nexB/vulnerablecode/issues/206  takes care of    this. Apart from different versions of `cvss` and sometimes indicators like `MEDIUM` , there seems no other scoring system, so IMHO having the severity indicators inside `VulnerabilityReference` seems more  appropriate, since it will contain all things you mentioned, so we won't lose on context.

pombredanne:
[...]

PR #216, opened 2020-06-25
[EXPERIMENT] New models

sbs2001:

vuln

The problems and how the new models solve them:

Problem 1 : A row in VulnerabilityReference does not properly encapsulate the reference ids and urls. Consider an importer encounters 2 distinct reference ids for a vulnerability, say USN-A and DSA-B and 2 corresponding urls 'www.USN-A.org' and 'www.DSA-B.org'. With the current implementation, A VulnerabilityReference can contain only single url and a single reference_id. What ends up happening is we either end up making 4 VulnerabilityReference row, like 

|....|USN A          | NULL             |
|....|DSA B          | NULL             |
|....|NULL            | 'www.USN-A.org' |
|....|NULL            | 'www.DSA-B.org'|

 or  have rows of combinations of reference_id and url , which might be unrelated. 

Problem 2: The `summary` of a vulnerability does not really belong in the Vulnerability table. Remember that vendors publish advisories in context of vulnerable packages, so the summary is usually related to the affected package + the vulnerability's nature. What happens in current implementation is, that the `summary` is overwritted  and replaced by the `summary` found by  last importer. 

Problem 3: Vulnerability scores, check https://github.com/nexB/vulnerablecode/issues/157

What this model lacks : 

1. Can't store version intervals. Suppose a vulnerability affects versions of a package lying in the interval [1.0, 3.0], there is no way to store that . On the other hand storing data which says all versions of package foo < 3.0 are vulnerable is possible. Ranges must  have only one bound, for this model to work.

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -39,9 +39,16 @@ class Vulnerability(models.Model):
     A software vulnerability with minimal information. Identifiers other than CVE ID are stored as
     VulnerabilityReference.
     """
-    cve_id = models.CharField(max_length=50, help_text='CVE ID', unique=True, null=True)
-    summary = models.TextField(help_text='Summary of the vulnerability', blank=True)
-    cvss = models.FloatField(max_length=100, help_text='CVSS Score', null=True)
+    vuln_id = models.CharField(max_length=50, help_text='eg CVE ID, RUST SEC ID', unique=True, null=True)
+    reference_ids = pgfields.JSONField() 
+
+    # Whatever goes into vuln_id is a vulnerability identifier 
+    # which is undivisible i.e atomic vulnerability id. All CVEs fit into this.
+    
+    # reference_ids are usually but not limited to `advisory` ids like USN-4399-1
+    # https://usn.ubuntu.com/4399-1/.     
+    # Contents of reference_ids are a name/id given to collection of 
+    # other small vulnerbilties. For example USN-4399-1 refers to CVE-2020-8618, CVE-2020-8619
 
     def __str__(self):
         return self.cve_id or self.summary
@@ -56,26 +63,25 @@ class VulnerabilityReference(models.Model):
     package manager.
     """
     vulnerability = models.ForeignKey(
-        Vulnerability, on_delete=models.CASCADE)
-    source = models.CharField(
-        max_length=50, help_text='Source(s) name eg:NVD', blank=True)
-    reference_id = models.CharField(
-        max_length=50, help_text='Reference ID, eg:DSA-4465-1', blank=True)
-    url = models.URLField(
-        max_length=1024, help_text='URL of Vulnerability data', blank=True)
+        Vulnerability, on_delete=models.CASCADE, required=True)    
+    source = models.ForeignKey(
+        Importer, on_delete=models.CASCADE, required=True)
+    urls = pgfields.JSONField()
+    summary = models.TextField()
 
     class Meta:
-        unique_together = ('vulnerability', 'source', 'reference_id', 'url')
+        unique_together = ('vulnerability', 'source')
 
-    def __str__(self):
-        return f'{self.source} {self.reference_id} {self.url}'
-
+class VulnerabilityScore(models.Model):
+  vulnerability_reference = models.ForeignKey(VulnerabilityReference, on_delete=models.CASCADE, required=True)
+  type = models.CharField(max_length=50, help_text='Vulnerability score type', blank=True)
+  score = models.CharField()
 
 class Package(PackageURLMixin):
     """
     A software package with links to relevant vulnerabilities.
     """
-    vulnerabilities = models.ManyToManyField(to='Vulnerability', through='ImpactedPackage')
+    vulnerabilities = models.ManyToManyField(to='Vulnerability', through='Vulnerability_Package_Relation')
 
     class Meta:
         unique_together = ('name', 'namespace', 'type', 'version', 'qualifiers', 'subpath')
@@ -111,24 +117,20 @@ class Package(PackageURLMixin):
         return self.package_url
 
 
-class ImpactedPackage(models.Model):
+class Vulnerability_Package_Relation(models.Model):
     """
     Relates a vulnerability to package(s) impacted by it.
     """
+# {
     vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
     package = models.ForeignKey(Package, on_delete=models.CASCADE)
+    is_vulnerable = models.BooleanField()
+# } till this point we have a consensus in this model
+
+    version_range = models.CharField()
 
     class Meta:
         unique_together = ('vulnerability', 'package')
-
-
-class ResolvedPackage(models.Model):
-    """
-    Relates a vulnerability to package(s) that contain a fix or resolution of this vulnerability.
-    """
-    vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
-    package = models.ForeignKey(Package, on_delete=models.CASCADE)
-
 
 class Importer(models.Model):
     """
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -33,6 +33,53 @@ from packageurl import PackageURL
 
 from vulnerabilities.data_source import DataSource
 
+class Importer(models.Model):
+    """
+    Metadata and pointer to the implementation for a source of vulnerability data (aka security
+    advisories)
+    """
+    name = models.CharField(max_length=100, unique=True, help_text='Name of the importer')
+
+    license = models.CharField(
+        max_length=100,
+        blank=True,
+        help_text='License of the vulnerability data',
+    )
+
+    last_run = models.DateTimeField(null=True, help_text='UTC Timestamp of the last run')
+
+    data_source = models.CharField(
+        max_length=100,
+        help_text='Name of the data source implementation importable from vulnerabilities.importers'
+    )
+    data_source_cfg = pgfields.JSONField(
+        null=False,
+        default=dict,
+        help_text='Implementation-specific configuration for the data source',
+    )
+
+    def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> DataSource:
+        """
+        Return a configured and ready to use instance of this importers data source implementation.
+
+        batch_size - max. number of records to return on each iteration
+        cutoff_date - optional timestamp of the oldest data to include in the import
+        """
+        importers_module = importlib.import_module('vulnerabilities.importers')
+        klass = getattr(importers_module, self.data_source)
+
+        ds = klass(
+            batch_size,
+            last_run_date=self.last_run,
+            cutoff_date=cutoff_date,
+            config=self.data_source_cfg,
+        )
+
+        return ds
+
+    def __str__(self):
+        return self.name
+
 
 class Vulnerability(models.Model):
     """
@@ -63,9 +110,9 @@ class VulnerabilityReference(models.Model):
     package manager.
     """
     vulnerability = models.ForeignKey(
-        Vulnerability, on_delete=models.CASCADE, required=True)    
+        Vulnerability, on_delete=models.CASCADE)    
     source = models.ForeignKey(
-        Importer, on_delete=models.CASCADE, required=True)
+        Importer, on_delete=models.CASCADE)
     urls = pgfields.JSONField()
     summary = models.TextField()
 
@@ -73,9 +120,9 @@ class VulnerabilityReference(models.Model):
         unique_together = ('vulnerability', 'source')
 
 class VulnerabilityScore(models.Model):
-  vulnerability_reference = models.ForeignKey(VulnerabilityReference, on_delete=models.CASCADE, required=True)
+  vulnerability_reference = models.ForeignKey(VulnerabilityReference, on_delete=models.CASCADE)
   type = models.CharField(max_length=50, help_text='Vulnerability score type', blank=True)
-  score = models.CharField()
+  score = models.CharField(max_length=50)
 
 class Package(PackageURLMixin):
     """
@@ -127,54 +174,7 @@ class Vulnerability_Package_Relation(models.Model):
     is_vulnerable = models.BooleanField()
 # } till this point we have a consensus in this model
 
-    version_range = models.CharField()
+    version_range = models.CharField(max_length=50)
 
     class Meta:
         unique_together = ('vulnerability', 'package')
-
-class Importer(models.Model):
-    """
-    Metadata and pointer to the implementation for a source of vulnerability data (aka security
-    advisories)
-    """
-    name = models.CharField(max_length=100, unique=True, help_text='Name of the importer')
-
-    license = models.CharField(
-        max_length=100,
-        blank=True,
-        help_text='License of the vulnerability data',
-    )
-
-    last_run = models.DateTimeField(null=True, help_text='UTC Timestamp of the last run')
-
-    data_source = models.CharField(
-        max_length=100,
-        help_text='Name of the data source implementation importable from vulnerabilities.importers'
-    )
-    data_source_cfg = pgfields.JSONField(
-        null=False,
-        default=dict,
-        help_text='Implementation-specific configuration for the data source',
-    )
-
-    def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> DataSource:
-        """
-        Return a configured and ready to use instance of this importers data source implementation.
-
-        batch_size - max. number of records to return on each iteration
-        cutoff_date - optional timestamp of the oldest data to include in the import
-        """
-        importers_module = importlib.import_module('vulnerabilities.importers')
-        klass = getattr(importers_module, self.data_source)
-
-        ds = klass(
-            batch_size,
-            last_run_date=self.last_run,
-            cutoff_date=cutoff_date,
-            config=self.data_source_cfg,
-        )
-
-        return ds
-
-    def __str__(self):
-        return self.name
--- a/vulnerablecode/settings.py
+++ b/vulnerablecode/settings.py
@@ -47,6 +47,7 @@ INSTALLED_APPS = [
     'vulnerabilities',
     'rest_framework',
     'django_filters',
+    'django_extensions'
 ]
 
 MIDDLEWARE = [
--- a/vulnerabilities/admin.py
+++ b/vulnerabilities/admin.py
@@ -24,10 +24,9 @@
 from django.contrib import admin
 
 from vulnerabilities.models import (
-    ImpactedPackage,
+    Vulnerability_Package_Relation,
     Importer,
     Package,
-    ResolvedPackage,
     Vulnerability,
     VulnerabilityReference,
 )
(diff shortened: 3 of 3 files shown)


vulnerablecode/settings.py
```diff
@@ -47,6 +47,7 @@
     'vulnerabilities',
     'rest_framework',
     'django_filters',
+    'django_extensions'
```
haikoschol, line 50: Is this related to the proposal?

vulnerabilities/models.py
```diff
@@ -111,68 +164,17 @@ def __str__(self):
         return self.package_url
 
 
-class ImpactedPackage(models.Model):
+class Vulnerability_Package_Relation(models.Model):
```
haikoschol, line 167: I assume this name is a placeholder, right? How about `VulnerabilityImpact`? Not great, not terrible, IMHO.

vulnerabilities/models.py
```diff
@@ -33,15 +33,69 @@
...
 
 class Vulnerability(models.Model):
     """
     A software vulnerability with minimal information. Identifiers other than CVE ID are stored as
     VulnerabilityReference.
     """
-    cve_id = models.CharField(max_length=50, help_text='CVE ID', unique=True, null=True)
-    summary = models.TextField(help_text='Summary of the vulnerability', blank=True)
-    cvss = models.FloatField(max_length=100, help_text='CVSS Score', null=True)
+    vuln_id = models.CharField(max_length=50, help_text='eg CVE ID, RUST SEC ID', unique=True, null=True)
```
haikoschol, line 89: Should this field really be nullable? What do we do with a vulnerability that has no `vuln_id` and no `reference_ids`? If your response is that there is always at least one reference ID, then why not store that in `vuln_id`? Or in other words; I haven't quite understood the difference between `vuln_id` and `reference_ids` here.

vulnerablecode/settings.py
sbs2001: No, I had to add that to generate the diagram

vulnerabilities/models.py
sbs2001: `VulnerabilityImpact` doesn't mention package anywhere, we need a name which should make sense that the table is about vulnerability and package

vulnerabilities/models.py
sbs2001: > I haven't quite understood the difference between vuln_id and reference_ids here.

I expected that, hence I had added comments in the code to define what is vuln_id and what is reference_id,  here is a repaste:

>   Whatever goes into vuln_id is a vulnerability identifier 
    which is undivisible i.e atomic vulnerability id. All CVEs fit into this.
    reference_ids are usually but not limited to `advisory` ids like USN-4399-1     
    Contents of reference_ids are a name/id given to collection of 
    other small vulnerbilties. For example USN-4399-1 refers to CVE-2020-8618, CVE-2020-8619

>If your response is that there is always at least one reference ID

As far as the advisories I've looked at, yes there is some sort of id present, but I'm not  100% confident whether this will stay true. 

>then why not store that in vuln_id?

Yes. If it is a atomic vulnerability id, then it probably didn't  belonged in the reference_id in the first place. RUST-SEC ids are stored in  reference_id no matter whether they have CVE or not. If they don't have a CVE, they become atomic, because no other id will denote that specific vulnerability.


I also had this idea, which I didn't mentioned here, but a `vuln_id`'s value should also be present along with(if present) other reference_ids in the `reference_ids` column. The idea being `vuln_id` is also it's own reference_id. 

>Should this field really be nullable? What do we do with a vulnerability that has no vuln_id and no reference_ids?

As @pombredanne mentioned, we have to give them  our ids, but that's gonna introduce a whole lot of other complexities(how to make id's consistent ?).

My other point is , should we really worry about vulnerabilities without any id's . As far as I have inspected these advisories, only FriendsOfPHP were missing these , which was solved, because GH provide their ids for  FriendsOfPHP advisories.

sbs2001:
@haikoschol re 
>Where did this information come from?

https://docs.djangoproject.com/en/3.1/releases/3.1/#jsonfield-for-all-supported-database-backends , https://github.com/django/django/pull/12392 , it was a GSoC 2019 project :)

vulnerabilities/models.py
haikoschol: Yeah that would be better.

vulnerabilities/models.py
haikoschol: Sorry, I should have mentioned that I read the comment and still didn't get it (undivisible/atomic and "small vulnerabilities" confused me). But now I think I understand. Some advisories cover multiple vulnerabilities and if those have CVE IDs, they will all be mentioned.

The problem with storing IDs other than CVE in `vuln_id` is that we require them to be unique across all publishers of advisories. That might be the case coincidentally, but I don't think there are any efforts to ensure that. But, in practice it will probably work and if not that problem can be solved when it occurs.

> As @pombredanne mentioned, we have to give them our ids, but that's gonna introduce a whole lot of other complexities(how to make id's consistent ?).

I think that was just referring to the automatically added primary key column.

I don't think we should worry about vulnerabilities without IDs. But the only reason I can think of for making this column nullable is to be able to store vulnerabilities without IDs. Hence my question. :)

vulnerabilities/models.py
pombredanne: @sbs2001 I am not convinced by the name too (and we should not use snake case for Class or model names).
The attributes are about a Package (is_vulnerable and version_range are all about Package) so a better name could might PackageAssignedVulnerability or PackageRelatedVulnerability ... but I need to think more about that change as what are the benefits to combine the Impacted and Resolved models in one?

vulnerabilities/models.py
pombredanne: @haikoschol re:
> As far as I have inspected these advisories, only FriendsOfPHP were missing these 
AFAIK, they use  date as ID then https://github.com/FriendsOfPHP/security-advisories/commit/c6fc7221e152c3a73fed562ad9c815ccf258a476#diff-a1ec953bbcb767e15ba1a9edbe828550

pombredanne: @sbs2001 I thin we could do better with a simpler model. 
- On the `vuln_id`, if we want and need this, it would then becomes our own id that we assign automatically IMHO. I am not sure we need an id though I can some benefit for users.
- On the reference side, IMHO one URL + reference ID is a reference, I cannot see when we need more than one URL. Can you elaborate that?

pombredanne:
@sbs2001 re:
> Problem 1 : A row in VulnerabilityReference does not properly encapsulate the reference ids and urls. Consider an importer encounters 2 distinct reference ids for a vulnerability, say USN-A and DSA-B and 2 corresponding urls 'www.USN-A.org' and 'www.DSA-B.org'. With the current implementation, A VulnerabilityReference can contain only single url and a single reference_id. What ends up happening is we either end up making 4 VulnerabilityReference row, like
> |....|USN A | NULL |
|....|DSA B | NULL |
|....|NULL | 'www.USN-A.org' |
|....|NULL | 'www.DSA-B.org'|
> or have rows of combinations of reference_id and url , which might be unrelated.


Rather I think we should get in this example only two rows as:

id         |   URL
----------|------------------------
USN A |  'www.USN-A.org'
DSA B | 'www.DSA-B.org'

vulnerabilities/models.py
sbs2001: @pombredanne  re
>On the reference side, IMHO one URL + reference ID is a reference, I cannot see when we need more than one URL. Can you elaborate that?

I have explained it in this ticket itself, can you take a look at Problem 1 ?

sbs2001:
@pombredanne  re 
>Rather I think we should get in this example only two rows as:

Exactly, that's the expected thing. FWIW there are cases where there are reference ids without corresponding urls and vice versa but that's not the issue. 

The problem is, as the current `Importer` implementation is there is no way for us to 'tell' the importer mechanism which reference id is related to which url. I really don't have any good ideas to fix that than manually specifying which id has which url, but then it gets really messy in the corner cases which are mentioned  above(not really corner cases) . 

Check https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L111 . 

So the next best thing I could think of is throw all the urls  and ref_ids in a JSONField.

vulnerabilities/models.py
sbs2001: @pombredanne 
>AFAIK, they use date as ID then 

Probably yes. 

>On the reference side, IMHO one URL + reference ID is a reference, I cannot see when we need more than one URL. Can you elaborate that?

Sure, I have done that in a  comment below

vulnerabilities/models.py
sbs2001: @pombredanne 

About the name thing, it's just a placeholder, Btw `PackageRelatedVulnerability` makes tad more sense here. 

>what are the benefits to combine the Impacted and Resolved models in one?

Good question. There are some issues with  having 2 tables, Impacted and Resolved . 

Issue 1 : 
You can do this, which doesn't make any sense. 
```

In [2]: v1 = models.Vulnerability.objects.create(cve_id="CVE-foo")                                                           

In [3]: p1 = models.Package.objects.create(name="cream",type="ice",version='mango') 
   
In [4]: vp1 = models.ImpactedPackage.objects.create(vulnerability=v1, package=p1)                                            

In [5]: vp2 = models.ResolvedPackage.objects.create(vulnerability=v1, package=p1)    
``` 
This is pure garbage, nothing can be interpreted from these entries. 

With a single table + flag, I can use a `unique_together=('vulnerability','package')`

Issue 2 :  Check  https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L121 , I am not sure I understand the exact issue but it's something along the lines  that  updating vulnerability status of a already existing package is not possible. @haikoschol  can you explain this, with a snippet? 

Having a single table, changes delete to an update(of the flag), which bypasses this issue.

sbs2001:
@pombredanne  @haikoschol  btw please mention things which you think are correctly implemented, once I have your aproval I can get started on that thing ASAP.

vulnerabilities/models.py
sbs2001: @pombredanne 
>On the vuln_id, if we want and need this, it would then becomes our own id that we assign automatically IMHO. I am not sure we need an id though I can some benefit for users.

I don't understand this, can you elaborate this further?

pombredanne:
re: https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L111
I think this code is the faulty one, you should not invoke twice `get_or_create()`  and the model
```Python
            for id_ in advisory.reference_ids:
                models.VulnerabilityReference.objects.get_or_create(
                    vulnerability=vuln, reference_id=id_)

            for url in advisory.reference_urls:
                models.VulnerabilityReference.objects.get_or_create(vulnerability=vuln, url=url)
```

The issue here is NOT the DB model but the Advisory data structure in https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/data_source.py#L60

This 
```Python
    reference_urls: Sequence[str] = dataclasses.field(default_factory=list)
    reference_ids: Sequence[str] = dataclasses.field(default_factory=list)
```
should be instead its own data structure that strictly mirrors the DB model, e.g. more or less something like that:
https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/models.py#L53
And may be the Django models can be used as-is too, without "saving" them? research needed?

```Python

@dataclasses.dataclass
class VulnerabilityReference: # use same name as model for now
    source = str
    reference_id = str
    url = str

[.....]

@dataclasses.dataclass
class Advisory
    references: Sequence[VulnerabilityReference] = dataclasses.field(default_factory=list)
```



Then the code in Importers at https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L111 would be more or less something like this:
```Python
# build indexes of existing references to either update or create new ones
# and avoid creating duplicates
refs_by_id_url = {(ref.id, ref.url): ref for ref in vuln.references if ref.id and ref.url}
refs_by_id = {ref.id: ref for ref in vuln.references if ref.id}
refs_by_url = {ref.url: ref for ref in vuln.references if ref.url}

for adv_ref in advisory.references:
    if adv_ref.reference_id and adv_ref.url:
        existing_ref = refs_by_id_url.get((adv_ref.id, adv_ref.url))
    elif adv_ref.reference_id:
        existing_ref = refs_by_id.get(adv_ref.id)
    elif adv_ref.reference_url:
        existing_ref = refs_by_id.get(adv_ref.url)
            
    if existing_ref:
        # update ref as neeed
        if adv_ref.summary and not existing_ref.summary:
            existing_ref.summary = adv_ref.summary
        if adv_ref.reference_id and not existing_ref.reference_id:
            existing_ref.reference_id = adv_ref.reference_id
        if adv_ref.url and not existing_ref.url:
            existing_ref.url = adv_ref.url
        if adv_ref.source and not existing_ref.source:
            existing_ref.source = adv_ref.source
        existing_ref.save()
    else:
        new_ref = VulnerabilityReference(
            vulnerability=vuln, 
            reference_id=ref.reference_id, 
            url=adv_ref.url,
            summary=adv_ref.summary,
            source=adv_ref.source,
        )
        new_ref.save()
```

sbs2001:
@pombredanne 

>And may be the Django models can be used as-is too, without "saving" them? research needed?

Yes they can be used, with one caveat , as they don't a pk(yet) they are unhashable, hence can't be put in sets. I have experienced problems due to this in https://github.com/nexB/vulnerablecode/pull/219 , where bulk inserting/updating was done. But simple workaround for this exists.

>I think this code is the faulty one, you should not invoke twice get_or_create() and the model

It is, @haikoschol  and I have chat about this on our gitter channel. The code was written assuming there is no relation between url and the id , which is true in some cases, while false in others, hence the assumption. 

This can be actually solved by tweaking the advisory model albeit.

```python
............
class Advisory : 
............
   vulnerability_references = Sequence[Mapping[str,str]]
```
eg value of vulnerability_references = [{'url':'usn123.com','reference_id':'usn-123'}]

In the importer 

```python
for vulnerability_reference in advisory.vulnerability_references : 
    VulnerabilityReference.objects.get_or_create(**vulnerability_reference)
```

pombredanne:
IMHO the get_or_create is too simplistic and in most case an actual object with named attributes is cleaner than a mapping, hence why I do not like too much `Sequence[Mapping[str,str]]`

pombredanne:
Also with:

> ```
> for vulnerability_reference in advisory.vulnerability_references : 
>     VulnerabilityReference.objects.get_or_create(**vulnerability_reference)
> ```

... you are missing out that thing should be merged and updated rather than just created or got IMHO

pombredanne:
> The code was written assuming there is no relation between url and the id , which is true in some cases, while false in others, hence the assumption.

IMHO there is always a relation between the two.

sbs2001:
@pombredanne  re
>... you are missing out that thing should be merged and updated rather than just created or got IMHO 

ack :) I messed up , but that's minor we can handle that, by doing something along the lines of the snippet you posted above. 

>in most case an actual object with named attributes is cleaner than a mapping, hence why I do not like too much Sequence[Mapping[str,str]]

That's a valid point in this case, how about 

```python
@dataclass(frozen=True)
class VulnerabilityReferenceArgs : 
 url : str = ''
 reference_id : str =''
 def __post_init__(self) : 
  if not any([self.url, self.reference_id]):
      raise SomeError
```
And `Sequence[VulnerabilityReferenceArgs]`

pombredanne:
@sbs2001 

> That's a valid point in this case, how about
> 
> ```python
> class VulnerabilityReferenceArgs : 
>  url : str = ''
>  reference_id : str =''
>  def __post_init__(self) : 
>   if not any([self.url, self.reference_id]):
>       raise SomeError
> ```
> 
> And have `Sequence[VulnerabilityReferenceArgs]` instead

:+1: 
that works (and we can find a better name than "VulnerabilityReferenceArgs" :))

sbs2001:
zzz

vulnerabilities/models.py
sbs2001: Never mind @haikoschol  already left a brief explanation of this issue in the comments at https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/tests/test_import_runner.py#L201

pombredanne:
VulnerabilityScore should be related to a VulnerabilityReference and not to a Vulnerability

pombredanne:
@sbs2001  in reply to https://github.com/nexB/vulnerablecode/pull/216#discussion_r449740167

> Issue 1 :
> You can do this, which doesn't make any sense.
> In [1]: from vulnerabilities import models                                                                                   
> In [4]: vp1 = models.ImpactedPackage.objects.create(vulnerability=v1, package=p1)                                            
> In [5]: vp2 = models.ResolvedPackage.objects.create(vulnerability=v1, package=p1)    
> This is pure garbage, nothing can be interpreted from these entries.

This is only a theoretical problem. Just do not do it :)
And if you want to guard from it, ensure in code that this fails if you attempt this with proper overrides in the model/maneger.


> Issue 2 :  Check  https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L121 , I am not sure I understand the exact issue but it's something along the lines  that  updating vulnerability status of a already existing package is not possible. @haikoschol  can you explain this, with a snippet? 

> Having a single table, changes delete to an update(of the flag), which bypasses this issue.

Note sure I get the details, but I think we have abused using `get_or_create` when things are often more complex than that and we want to do many more checks that just a simple get to ensure we have good data.

sbs2001:
>VulnerabilityScore should be related to a VulnerabilityReference and not to a Vulnerability

I thought it would be ok since the old models had score in Vulnerability NOT in VulnerabilityReference. 

What we are trying to acheive here is to store multiple scores for a single Vulnerability. 

Using score related VulnerabilityReference  makes it very tricky to figure out of which VulnerabilityReference , is the VulnerabilityScore  based of.

pombredanne:
> What we are trying to acheive here is to store multiple scores for a single Vulnerability.

IMHO rather we are trying to attach the score to a given reference which is really the only thing we can be sure of. And a ref (with its score) may be linked in multiple Vulnerabilities

pombredanne:
> Using score related VulnerabilityReference makes it very tricky to figure out of which VulnerabilityReference , is the VulnerabilityScore based of.

Why so?

I see it this way, for a given vulnerability
- ref1(NVD CVE) says the score type cvss2 is 5.0
- ref1(NVD CVE) says the score type cvss3 is 6.0
- ref2(RedHat CVE) says the score type cvss2 is 7.0
which becomes really straight forward to navigate and display and query IMHO.

sbs2001:
@pombredanne re 
>This is only a theoretical problem. Just do not do it :)

I disagree, there are instances where the sources contradict themselves, in such cases IMHO we should bring it to their notice, community curation :) . I have posted in chat about how archlinux advisory has 2 contradicting entries

sbs2001:
>Why so?
I see it this way, for a given vulnerability
    ref1(NVD CVE) says the score type cvss2 is 5.0
    ref1(NVD CVE) says the score type cvss3 is 6.0
    ref2(RedHat CVE) says the score type cvss2 is 7.0
    which becomes really straight forward to navigate and display and query IMHO.

I have a  FK of 'source' already, so it would mean the same.

PR #219, opened 2020-06-28
Refactor ImpactedPackage and ResolvedPackage

sbs2001:
Changes are made in import_runner.py to refactor the models.
Other major changes in import_runner.py is to change the _process_updated_advisories
method to use less db queries and increase performance by several times.

Having a single table for these, fixes the following: 

**Issue 1 :**

You can do this, which doesn't make any sense.
```python

In [2]: v1 = models.Vulnerability.objects.create(cve_id="CVE-foo")                                                           

In [3]: p1 = models.Package.objects.create(name="cream",type="ice",version='mango') 
   
In [4]: vp1 = models.ImpactedPackage.objects.create(vulnerability=v1, package=p1)                                            

In [5]: vp2 = models.ResolvedPackage.objects.create(vulnerability=v1, package=p1)   
``` 

This is pure garbage, nothing can be interpreted from these entries.

With a single table + flag, we  can use a unique_together=('vulnerability','package')

**Issue 2 :** 

Updating the vulnerability status of a package is not possible due to  cascade deletes. 

Check https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/import_runner.py#L121 . 

And also check for more details https://github.com/nexB/vulnerablecode/blob/58d0376e7319d06387662cb393f3c39d9893088d/vulnerabilities/tests/test_import_runner.py#L201


vulnerabilities/import_runner.py
```diff
@@ -58,6 +61,46 @@ class ImportRunner:
...
+    def __del__(self):
+        models.Vulnerability_Package_Relation.objects.bulk_create(self._bulk_create_vuln_pkg_refs, ignore_conflicts=True)
+        models.Vulnerability_Package_Relation.objects.bulk_update(self._bulk_update_vuln_pkg_refs,['is_vulnerable'])
+        models.VulnerabilityReference.objects.bulk_create(self._bulk_create_vuln_refs, ignore_conflicts=True)
+
+    # TODO: Currently the implementation may contain duplicate model instances in `_bulk_create_vuln_refs` and `_bulk_create_vuln_pkg_refs`
+    #       to make this work , `ignore_conflicts=True` is used. Find some way to remove duplicates of model instances(before writing them db)
+    #       and remove the   `ignore_conflicts=True` flag.
+    @property  
+    def bulk_create_vuln_pkg_refs(self):
```
haikoschol, line 77: Property getters should not have side-effects. IMHO this approach has too much magic, including the need for the `__del__()` method.

vulnerabilities/import_runner.py
```diff
@@ -58,6 +61,46 @@ class ImportRunner:
...
+
+        if len(self._bulk_create_vuln_pkg_refs) >= self.batch_size : 
+            models.Vulnerability_Package_Relation.objects.bulk_create(self._bulk_create_vuln_pkg_refs, ignore_conflicts=True)
```
haikoschol, line 80: Are you sure the performance gains are not mostly due to `ignore_conflicts=True`?

sbs2001: No. The performance gain is due to removing the excess `get_or_create` and utilising the fact that, if 

A Vulnerability row didn't previously existed(created=True), then we for sure know that all objects which it is a foreign key in, don't exist too, hence use bulk create. Same goes with `Package`.

I had gotten it to work by simply having lists at a `Advisory` level for bulk inserts without ignore_conflicts, but that way sometimes I would bulk_create on a single item, or sometimes thousand items, everything depended on the nature Advisory being passed. 

The conflicts are caused due to duplicate model instances in. 
  ```
        self._bulk_create_vuln_pkg_refs = []
        self._bulk_update_vuln_pkg_refs = []
        self._bulk_create_vuln_refs = []

```
 I learned the hard way that I can't put these instances in a `set` because their hashes are created on basis of pk(which they get after being written to db) :) .

File: vulnerabilities/import_runner.py (hunk 1 above)
Comment (sbs2001) [reply vulnerabilities/import_runner.py:77]: Agreed this  and `ignore_conflicts=True` are totally bad practices. Good news is after tinkering about this, I can totally get rid of these magic methods and properties, even maintain `ImporterRunner`'s API like the old one.

**Comment by sbs2001** [comment]
@haikoschol  I have pushed a commit to stop using `ignore_conflicts=True` and properties and magic methods. It's still faster than the current implementation, but not as fast as the one with properties. 

I think I can acheive the same results as the one with properties, if I could remove duplicates of model instances(these dont yet have pk ) from a list. Do you have any ideas to do that ?

**Comment by haikoschol** [comment]
> I think I can acheive the same results as the one with properties, if I could remove duplicates of model instances(these dont yet have pk ) from a list. Do you have any ideas to do that ?

Have you tried the naive approach of just iterating the list and comparing all fields?

**Comment by sbs2001** [comment]
@haikoschol 
>Have you tried the naive approach of just iterating the list and comparing all fields?

That is a valid approach, I instead went with creating a dataclass and put it into a set and then instantiate model instances. I've tested it for `alpine`, `debian`, `rust`, `openssl` importers, all works fine. 

Could you suggest an OK name for the dataclass ?

File: vulnerabilities/import_runner.py [hunk 3]
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+                vulnerability=vuln, url=url))
 
-            for id_ in advisory.reference_ids:
-                models.VulnerabilityReference.objects.get_or_create(
-                    vulnerability=vuln, reference_id=id_)
+    else:
+        vuln_refs_qs = models.VulnerabilityReference.objects.filter(
+            vulnerability=vuln)
+        vuln_ids = {ref.reference_id for ref in vuln_refs_qs}
+        vuln_urls = {ref.url for ref in vuln_refs_qs}
```
pombredanne, line 187: Why do you treat the reference_id and url separately? a URL for a reference_id should not lead to multiple references.

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
-                #         qs[0].delete()
+    return vuln, vuln_created, vuln_references
 
-                models.ImpactedPackage.objects.get_or_create(
-                    vulnerability_id=vuln.id, package_id=pkg.id)
 
-            for rpkg_url in advisory.resolved_package_urls:
-                pkg, created = _get_or_create_package(rpkg_url)
+def _create_pkg_vuln_refs(vuln: models.Vulnerability, vuln_created: bool, purls: Sequence[PackageURL], is_vulnerable: bool):  # nopep8
+    new_refs, updated_refs = set(), set()
```
pombredanne, line 207: using one declaration per line is always more readable

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+    for purl in purls:
+        pkg, pkg_created = _get_or_create_package(purl)
+        vuln_pkg_ref = PackageRelatedVulnerability_inserter(
+            package=pkg, vulnerability=vuln, is_vulnerable=is_vulnerable)
+        if pkg_created or vuln_created:
+            new_refs.add(vuln_pkg_ref)
+
+        else:
+            qs = models.PackageRelatedVulnerability.objects.filter(
```
pombredanne, line 216: Is there a better name than qs?

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+                package=pkg, vulnerability=vuln)
+            if not qs:
+                new_refs.add(vuln_pkg_ref)
+            else:
+                # Note: PackageRelatedVulnerability has constraints
+                # unique_together = ('package', 'vulnerability', 'is_vulnerable')
+                # This fact is used below.
+                vuln_impact = {i.is_vulnerable for i in qs}
+                if len(vuln_impact) == 2 or is_vulnerable not in vuln_impact:
```
pombredanne, line 225: What does ==2  mean? that's pretty cryptic

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+                    conflicts = [i for i in qs]
```
pombredanne, line 226: If you need a copy of the items list, use may be: `foo = qs[:]`

vulnerabilities/import_runner.py
```diff
@@ -24,20 +24,59 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class PackageRelatedVulnerability_inserter:
+    vulnerability: models.Vulnerability
+    is_vulnerable: bool
+    package: models.Package
+
+    def to_model_object(self):
+        return models.PackageRelatedVulnerability(**dataclasses.asdict(self))
```
pombredanne, line 78: Is there a need to couple these two objects?

vulnerabilities/import_runner.py
```diff
@@ -169,48 +263,46 @@ def _get_or_create_package(p: PackageURL) -> Tuple[models.Package, bool]:
...
 def _bulk_insert_packages(
-        impacted: Set[PackageURL],
-        resolved: Set[PackageURL]
-) -> Tuple[Dict[PackageURL, models.Package], Dict[PackageURL, models.Package]]:
-
-    packages = [_package_url_to_package(p) for p in impacted.union(resolved)]
-    packages = models.Package.objects.bulk_create(packages)
+        impacted: List[PackageURL],
+        resolved: List[PackageURL]
+) -> Tuple[Dict[PackageURL, int], Dict[PackageURL, int]]:
```
pombredanne, line 283: Have you enabled type checking at least in the CI?
Otherwise, type hints are not always helpful and if the types are not checked somehow, they can be misleading and i most case they are not conducive to improve readability... So let's not get carried away with type hints for the sake of them. Instead enter a ticket so we can discuss and plan for their possible usage

vulnerabilities/import_runner.py
```diff
@@ -169,48 +263,46 @@ def _get_or_create_package(p: PackageURL) -> Tuple[models.Package, bool]:
...
 
-    for pkg in packages:
-        # Unfortunately, PackageURLMixin.package_url returns a string, not a PackageURL
-        purl = PackageURL.from_string(pkg.package_url)
-
-        if purl in impacted:
-            impacted_packages[purl] = pkg
-        elif purl in resolved:
-            resolved_packages[purl] = pkg
+    impacted_packages = dict(
```
pombredanne, line 290: I understand the zip here, but is there a simpler way to get there?
I am always nervous when we start using "parallel lists" to later zip them.

vulnerabilities/import_runner.py
```diff
@@ -169,48 +263,46 @@ def _get_or_create_package(p: PackageURL) -> Tuple[models.Package, bool]:
...
 
     return impacted_packages, resolved_packages
 
 
 def _bulk_insert_impacted_and_resolved_packages(
     batch: Set[Advisory],
     vulnerabilities: Set[models.Vulnerability],
-    impacted_packages: Dict[PackageURL, models.Package],
-    resolved_packages: Dict[PackageURL, models.Package],
+    impacted_packages: Dict[PackageURL, int],
```
pombredanne, line 301: What is this int about?

vulnerabilities/import_runner.py
```diff
@@ -221,30 +313,31 @@ def _bulk_insert_impacted_and_resolved_packages(
...
                 p = _package_url_to_package(impacted_purl)
                 p.save()
-                impacted_packages[impacted_purl] = p
+                impacted_packages[impacted_purl] = p.id
 
-            ip = models.ImpactedPackage(
+            ip = models.PackageRelatedVulnerability(
                 vulnerability=vuln,
-                package=p,
+                package_id=p,
```
pombredanne, line 320: is p a package or a package_id?

vulnerabilities/import_runner.py
```diff
@@ -221,30 +313,31 @@ def _bulk_insert_impacted_and_resolved_packages(
...
+            refs.append(ip)
 
         for resolved_purl in advisory.resolved_package_urls:
             # TODO Figure out when/how it happens that a package is missing from the dict and fix it
             p = resolved_packages.get(resolved_purl)
             if not p:
                 p = _package_url_to_package(resolved_purl)
                 p.save()
-                resolved_packages[resolved_purl] = p
+                resolved_packages[resolved_purl] = p.id
```
pombredanne, line 331: What are the benefits of using ids vs objects there?

pombredanne:
@sbs2001  after our discussion I am not fine with the merging of ImpactedPackage and ResolvedPackage in a single model. :)
You call for a simpler DB-level referential integrity makes sense!

vulnerabilities/import_runner.py
sbs2001: This is whole another issue, with the Advisory dataclass which you pointed out before. In this PR I haven't fixed that issue yet, so this mimics the already buggy code. I'll file another PR to fix the reference issue.

vulnerabilities/import_runner.py
sbs2001: This definitely needs an explanation. I have changed  `len(vuln_impact) == 2` to `vuln_impact == {True, False}`, having done that, it still deserves an explanation, I've added the following comments to the code. I hope this clarifies things. 
```
               is_vulnerable is a boolean which indicates the relationship of
                a vulnerability's impact on a package. vuln_impact is a set of
                all such booleans for a pair of (vulnerability, package). In cases
                where  vuln_impact == {True, False}, we know that conflicting relationships
                of (vulnerability, package) ALREADY EXIST in the DB.
                The other check `is_vulnerable not in vuln_impact` is used to know whether the
                data we just found is not conflicting with the data already existing in DB.
                In any of the above two cases we move the entries involved in ImportProblem
```

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
 
 
-def _process_updated_advisories(data_source: DataSource) -> None:
-    """
-    TODO: Make efficient; Current implementation needs way too many DB queries.
-    """
-    for batch in data_source.updated_advisories():
-        for advisory in batch:
-            vuln, _ = _get_or_create_vulnerability(advisory)
+def _create_vulnerability_and_references(advisory: Advisory):
```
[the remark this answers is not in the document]
sbs2001: @haikoschol  went with following https://pep8.org/#descriptive-naming-styles I continued with it. These methods won't be used in other modules anyways, java eq of `private` so it makes sense to me.  

Is a prefix of _  used to indicate the method is not tested ? That's new. 

I'm confused with these conventions now ;)

vulnerabilities/import_runner.py
sbs2001: Yes, did you read the comments in code? There only purpose  is to `remove duplicates of model instances(these dont yet have pk ) from a list` . It's terrible approach, but I can't think of a better way to do this. There's naive way  doing it as @haikoschol  suggested above, but that will probably more or less neutralize the performance gained by bulk inserts as that approach has time complexity of O(N^2) while this dataclass approach has O(N)

vulnerabilities/import_runner.py
```diff
@@ -24,20 +24,59 @@
...
 logger = logging.getLogger(__name__)
 
+# These _inserter classes are used to instantiate model objects.
+# Frozen dataclass  store args required to store instantiate
+# model objects, this way model objects can be hashed indirectly which
+# is required in this implementation.
+
+
+@dataclasses.dataclass(frozen=True)
+class VulnerabilityReference_inserter:
```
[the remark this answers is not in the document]
sbs2001: Done

vulnerabilities/import_runner.py
```diff
@@ -24,20 +24,59 @@
...
+        return models.VulnerabilityReference(**dataclasses.asdict(self))
+
+# These _inserter classes are used to instantiate model objects.
+# Frozen dataclass  store args required to store instantiate
+# model objects, this way model objects can be hashed indirectly which
+# is required in this implementation.
+
+
+@dataclasses.dataclass(frozen=True)
+class PackageRelatedVulnerability_inserter:
```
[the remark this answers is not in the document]
sbs2001: Done

vulnerabilities/import_runner.py
```diff
@@ -71,19 +111,43 @@ def run(self, cutoff_date: datetime.datetime = None) -> None:
...
+            tz=datetime.timezone.utc)
         self.importer.data_source_cfg = dataclasses.asdict(data_source.config)
         self.importer.save()
 
         logger.debug(f'Successfully finished import for {self.importer.name}.')
 
 
+def _process_updated_advisories(data_source: DataSource) -> None:
+    """
+    TODO: Break this method into smaller functions
```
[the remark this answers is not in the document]
sbs2001: No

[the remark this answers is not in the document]
sbs2001: Removed

vulnerabilities/import_runner.py
```diff
@@ -71,19 +111,43 @@ def run(self, cutoff_date: datetime.datetime = None) -> None:
...
+    """
+    bulk_create_vuln_refs = set()
+    bulk_create_vuln_pkg_refs = set()
+    for batch in data_source.updated_advisories():
+        for advisory in batch:
+
```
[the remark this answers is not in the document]
sbs2001: :) Removed

vulnerabilities/import_runner.py
```diff
@@ -71,19 +111,43 @@ def run(self, cutoff_date: datetime.datetime = None) -> None:
...
+            vuln, vuln_created, references = _create_vulnerability_and_references(
+                advisory)
+            bulk_create_vuln_refs.update(references)
+            inew_refs = _create_pkg_vuln_refs(
+                vuln, vuln_created, advisory.impacted_package_urls, is_vulnerable=True)
+            rnew_refs = _create_pkg_vuln_refs(
+                vuln, vuln_created, advisory.resolved_package_urls, is_vulnerable=False)
+            bulk_create_vuln_pkg_refs.update(inew_refs.union(rnew_refs))
+
+    models.VulnerabilityReference.objects.bulk_create(
```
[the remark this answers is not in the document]
sbs2001: We had this discussion in the chat. I think this is justified

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
-    """
-    for batch in data_source.updated_advisories():
-        for advisory in batch:
-            vuln, _ = _get_or_create_vulnerability(advisory)
+def _create_vulnerability_and_references(advisory: Advisory):
+    vuln, vuln_created = _get_or_create_vulnerability(advisory)
+    vuln_references = set()
+
+    if vuln_created:
+        # This means vulnerability didn't previously exist in the db, so bulk create
```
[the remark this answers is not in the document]
sbs2001: Yes, rephrased it albeit to make it hopefully more clear :)

vulnerabilities/import_runner.py
```diff
@@ -92,54 +156,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+        # is used without any hesitation
+        for id_ in set(advisory.reference_ids):
```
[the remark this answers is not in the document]
sbs2001: Good catch, removed the cast to set

vulnerabilities/import_runner.py
sbs2001: Turns out I don't need updated_refs anymore. This PR is mess ;)

vulnerabilities/import_runner.py
sbs2001: Using `existing_pkg_vuln_refs` instead, hope that makes sense

vulnerabilities/import_runner.py
sbs2001: Thanks, that's nice I'm using your suggested way now :)

vulnerabilities/import_runner.py
sbs2001: Reverted to previous approach, which was more readable.

vulnerabilities/import_runner.py
sbs2001: Reverting this. p is a package.

vulnerabilities/import_runner.py
sbs2001: Overenginnering/pre-mature optimization . The idea was to use Dict[PackageURL, int] where the int is pk of Package , to save some memory because these dictionaries would grow big very fast. Carrying the Objects is way more expensive than pks so  I had used it . 

Reverting this.

vulnerabilities/import_runner.py
sbs2001: As mentioned above this is : 

Overenginnering/pre-mature optimization . The idea was to use Dict[PackageURL, int] where the int is pk of Package , to save some memory because these dictionaries would grow big very fast. Carrying the Objects is way more expensive than pks so I had used it .

Reverting this.

vulnerabilities/import_runner.py
sbs2001: Agreed. In the CI it would be something along the lines of running  a simple Mypy check right ? That probably won't be much of pain. 

IMHO using `typing` help in most cases to atleast know what's going in and out, way better than comments, but that's just my opinion. Looking at this again I feel rather ashamed of this case though, it really seems I've used  `typing` in  a rather cryptic way.  

Reverting this method, to the old one which was readable.

sbs2001: the ticket https://github.com/nexB/vulnerablecode/issues/226 :)

vulnerabilities/import_runner.py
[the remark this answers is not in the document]
pombredanne: Just asking :)

vulnerabilities/import_runner.py
pombredanne: OK, that's quite important to get right!

pombredanne:
At this stage since this is blocking further updates, I am merging it with the caveats noted otherwise... :+1:

Issue #208, opened 2020-06-20
Revisit using migrations for initial data seeds


pombredanne:
There are possible issues with using migrations to add new data, such as if we want to squash migrations in the future and tying the addition of data to a historical model version of when they were added. It could be simpler to reconsider using migration and have a different seeding model like a class yielding some value, etc. That would be more flexible too IMHO.

haikoschol:
The Django docs describe how to provide initial data here: https://docs.djangoproject.com/en/3.0/howto/initial-data/

The alternative to data migrations described on that page is using fixtures. I don't remember why I went with migrations instead of fixtures. Probably because writing Python is less annoying than JSON. Or because data migrations are included when the user runs `./manage.py migrate` and fixtures require an additional command (which is easy enough to mention in the docs).

sbs2001:
@pombredanne , can you please suggest other concrete alternatives. As @haikoschol suggested  we can use fixtures , but is that going solve the problem? In fact I still don't understand the problem.

PR #221, opened 2020-06-29
Remove migrations and add importer yielder module

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/208. 

 A registry/fixture of importers are added. The ~class `ImporterYielder`~ module `importer_yielder` yields Importer objects as required. No need to  add any commands.


vulnerabilities/importer_registry.json
```diff
@@ -0,0 +1,182 @@
+[
```
haikoschol, line 1: Looks like the line endings here need fixing.

sbs2001: I don't understand what you mean.
Can you elaborate ? Do you mean ^M , that is not me ;) it's a stupid way of rendering json github uses, try raw mode.

haikoschol: > Can you elaborate ? Do you mean ^M , that is not me ;) it's a stupid way of rendering json github uses, try raw mode.

Interesting. Never seen that before. Looks like DOS line endings. nvm then :)

pombredanne: The line endings are DOS indeed. We should use LF *nix instead:
```
$ wget https://raw.githubusercontent.com/nexB/vulnerablecode/7bfa1c57d3cb4846067ff59ca2d6cde667fee29d/vulnerabilities/importer_registry.json
$ file importer_registry.json 
importer_registry.json: ASCII text, with CRLF, CR line terminators
```

vulnerabilities/importer_yielder.py
```diff
@@ -20,33 +20,40 @@
...
-def remove_openssl_importer(apps, _):
-    Importer = apps.get_model('vulnerabilities', 'Importer')
-    qs = Importer.objects.filter(name='openssl')
-    if qs:
-        qs[0].delete()
+    def __init__(self):
+        self.load_registry()
 
+    def load_registry(self):
+        registry_path = os.path.join(BASE_DIR, 'importer_registry.json')
```
pombredanne, line 37: Why using a JSON file as opposed to a plain Python script? What is the benefit?

sbs2001: Plain python  works too. I can't think of any benefit of using one over another, except some difference in resulting syntax. 

I think I went with JSON, since python script is not usually used for storing data.

sbs2001: This made sense after the discussion.

 Will use python script, because the data isn't too much to warrant a separate JSON, because doing so will introduce more unnecessary complexity. `importer_yielder.py` would encapsulate data+logic to seed the project.

PR #225, opened 2020-07-10
Refactor import runner(no change to ImpactedPackage, ResolvedPackage)

sbs2001:
It throws all the conflicting entries into a JSONField , so we can later ping the sources to resolve these, which is IMHO the correct way to deal with the contradictions rather than overwriting the latest entry . It also increases the performance by utilising the fact that if an entity(vulnerability or package) is created, then it's references don't exist in DB.


vulnerabilities/import_runner.py
```diff
@@ -100,46 +175,84 @@ def _process_added_advisories(data_source: DataSource) -> None:
...
+            package=pkg, vulnerability=vuln, is_vulnerable=is_vulnerable)
 
-                models.ResolvedPackage.objects.get_or_create(
-                    vulnerability_id=vuln.id, package_id=pkg.id)
+        if pkg_created or vuln_created:
+            new_refs.add(vuln_pkg_ref)
+
+        else:
+
+            # Check for conflicts
```
pombredanne, line 228: I am not sure I get the code block below. Can you elaborate?

sbs2001:
Closed in  Favour #219

Issue #223, opened 2020-07-04
Modify Advisory to encapsulate reference_id and urls properly


sbs2001:
Check https://github.com/nexB/vulnerablecode/pull/216#issuecomment-653743223

PR #227, opened 2020-07-16
Stick ref ids with urls

sbs2001:
This is based on https://github.com/nexB/vulnerablecode/pull/219 so please review that before this. 

This PR is implementation of https://github.com/nexB/vulnerablecode/pull/216#issuecomment-653778007

Fixes https://github.com/nexB/vulnerablecode/issues/223

### DIFF
--- a/vulnerabilities/data_source.py
+++ b/vulnerabilities/data_source.py
@@ -44,7 +44,7 @@ from vulnerabilities.oval_parser import OvalParser
 
 
 @dataclasses.dataclass
-class VulnerabilityReferenceUnit:
+class Reference:
 
     url: str = ''
     reference_id: str = ''
@@ -68,7 +68,7 @@ class Advisory:
     summary: str
     impacted_package_urls: Iterable[PackageURL]
     resolved_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
-    vuln_references: List[VulnerabilityReferenceUnit] = dataclasses.field(default_factory=list)
+    vuln_references: List[Reference] = dataclasses.field(default_factory=list)
     cve_id: Optional[str] = None
 
     def __hash__(self):
@@ -481,7 +481,7 @@ class OvalDataSource(DataSource):
             description = definition_data['description']
             affected_purls = set()
             safe_purls = set()
-            references = [VulnerabilityReferenceUnit(url=url)
+            references = [Reference(url=url)
                           for url in definition_data['reference_urls']]
 
             for test_data in definition_data['test_data']:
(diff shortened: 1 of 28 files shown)


vulnerabilities/data_source.py
```diff
@@ -57,16 +68,14 @@ class Advisory:
     summary: str
     impacted_package_urls: Iterable[PackageURL]
     resolved_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
-    reference_urls: Sequence[str] = dataclasses.field(default_factory=list)
-    reference_ids: Sequence[str] = dataclasses.field(default_factory=list)
+    vuln_references: List[VulnerabilityReferenceUnit] = dataclasses.field(default_factory=list)
     cve_id: Optional[str] = None
 
     def __hash__(self):
```
sbs2001, line 74: @pombredanne  and @haikoschol  what's the best way to handle this ? The this is being done atm is obviously not the idel way.

pombredanne: VulnerabilityReferenceUnit is not my preferred name, but your design above looks OK.

sbs2001: > Why do you think you need a hash here? 

Not 100% sure as @haikoschol  implemented this, probably because we want to put multiple Advisory's in  a set to avoid duplicates. 

I think Advisory can be made immutable with no pain, that way we won't need to expicitly implement __hash__ .

sbs2001: The hash is there for a reason :) . 

I'm getting this  even after making Advisory immutable(by passing frozen=True in the dataclass decorator.) 

```
  File "/home/shivam/coding/opensource/vulnerablecode/vulnerabilities/data_source.py", line 191, in batch_advisories
    yield set(b)
  File "<string>", line 3, in __hash__
TypeError: unhashable type: 'set'

```

pombredanne: @sbs2001 avoid using set. They are unordered and are seldom needed and you will no longer need hashing override

vulnerabilities/data_source.py
```diff
@@ -43,6 +43,17 @@
 from vulnerabilities.oval_parser import OvalParser
 
 
+@dataclasses.dataclass
+class VulnerabilityReferenceUnit:
```
pombredanne, line 47: Why not call it what it is? Simply ither a `VulnerabilityReference` or a `Reference`

sbs2001: Calling it Reference

vulnerabilities/data_source.py
sbs2001: @pombredanne  order doesn't matter in many places, since all Advisories will be dumped to database, it is just an intermediate data structure. Having said that, we really don't to have duplicate Advisories as those will cause IntegrityError

PR #228, opened 2020-07-17
Use node js git security advisories instead of using npm registry

sbs2001:
This PR switches the NPM importer to use https://github.com/nodejs/security-wg instead of  https://registry.npmjs.org/-/npm/v1/security/advisories

Reasons : 
 https://registry.npmjs.org/-/npm/v1/security/advisories has given trouble to us in past, recently it was down for some time. The git data source is way more robust.

We also get to use GitDataSource which has alot of goodies.

https://github.com/nodejs/security-wg is the upstream
 


pombredanne:
Please rebase and I will mereg

sbs2001:
> I may wonder if using a direct download instead of a git checkout may have been simpler... but that's a small thing

Good question. Thing is `GitDataSource` uses checkout to detect changes inside a git repository. So if you run an  Importer again after some duration it will only fetch the data which is new/changed.

PR #231, opened 2020-07-23
Add a package_managers module

sbs2001:
Before this commit every Importer had it's own VersionAPI class, which was
used to fetch all versions of a package from the corresponding package manager.

This was problematic due to 2 reasons (1)  Reusing these VersionAPI was hacky
(2) The importer files would contain most of the code related to fetching this
version info.

This is fixed by putting all this VersionAPI classes into one
vulnerabilities/package_managers.py module and the importers will now import
the classes from here.

### DIFF
--- a/vulnerabilities/package_managers.py
+++ b/vulnerabilities/package_managers.py
@@ -1,0 +1,301 @@
+# Copyright (c)  nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an 'AS IS' BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import asyncio
+from typing import Mapping
+from typing import Set
+from typing import List
+from urllib.error import HTTPError
+from urllib.request import urlopen
+import xml.etree.ElementTree as ET
+
+import requests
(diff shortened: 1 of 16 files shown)


vulnerabilities/package_managers.py
```diff
@@ -0,0 +1,301 @@
...
+            releases = set()
+
+            try:
+                with urlopen(f"https://crates.io/api/v1/crates/{package_name}") as response:
+                    response = json.load(response)
+                    for version_info in response["versions"]:
+                        releases.add(version_info["num"])
+            except HTTPError as e:
+                if e.code == 404:
+                    pass
```
pombredanne, line 118: Do you really want to pass on a 404?

vulnerabilities/package_managers.py
```diff
@@ -0,0 +1,301 @@
...
+
+        try:
+            xml_resp = ET.ElementTree(ET.fromstring(resp.decode("utf-8")))
+            self.cache[pkg_name] = self.extract_versions(xml_resp)
+        except ET.ParseError:
+            self.cache[pkg_name] = set()
+
+    @staticmethod
+    def artifact_url(artifact_comps: List[str]) -> str:
+        base_url = "https://repo.maven.apache.org/maven2/{}"
```
pombredanne, line 227: Maven central canonical URL is usually `https://repo1.maven.org/maven2/` and `https://repo.maven.apache.org/maven2/` is a secondary.

vulnerabilities/package_managers.py
```diff
@@ -0,0 +1,301 @@
...
+        return f"https://repo.packagist.org/p/{vendor}/{name}.json"
+
+    @staticmethod
+    def extract_versions(resp: dict, pkg_name: str) -> Set[str]:
+        all_versions = resp["packages"][pkg_name].keys()
+        # This filter ensures, that all_versions contains only released versions
+        all_versions = set(filter(lambda x: "dev" not in x, all_versions))
+        # See https://github.com/composer/composer/blob/44a4429978d1b3c6223277b875762b2930e83e8c/doc/articles/versions.md#tags  # nopep8
+        # for explanation of removing 'v'
+        all_versions = set(map(lambda x: x.replace("v", ""), all_versions))
```
pombredanne, line 299: I would avoid using set() here and there for lists of versions. They are unique upstream anyway and you lose the ordering which is IMHO relevant. Also it is often better to use a list comprehension or a generator expression rather than map and filter with lambdas: they are strictly equivalent and require less effort to read and understand

vulnerabilities/package_managers.py
sbs2001: I just moved this from https://github.com/nexB/vulnerablecode/blob/a15c726a83b515cc4566b79110901056279bd31e/vulnerabilities/importers/safety_db.py#L154 to separate file in this PR. Since I had implemented this importer's first edition, the idea was that given a package which does not exist in the pypi db, the endpoint would emit a 404. The pass just returns an empty set of versions for such packages. I don't see how this could be better handled

vulnerabilities/package_managers.py
sbs2001: Order doesn't matter here , set vs list came up before too.

I have changed filter,map,lambda to set comprehension, it is better and easy to understand too now. Thanks.

vulnerabilities/package_managers.py
sbs2001: Changed. Thanks for the info

sbs2001:
Closed in  favour of https://github.com/nexB/vulnerablecode/pull/236

PR #234, opened 2020-07-28
Populate the licenses

sbs2001:
Redhat license -> https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/legal-notice

Arch license -> https://github.com/archlinux/arch-security-tracker/blob/master/LICENSE

Ubuntu license -> https://launchpad.net/ubuntu-cve-tracker

Rest are found from their respective Git Repos


vulnerabilities/importer_yielder.py
```diff
@@ -25,7 +25,7 @@
 IMPORTER_REGISTRY = [
     {
         'name': 'rust',
-        'license': 'https://creativecommons.org/publicdomain/zero/1.0/',
+        'license': 'CC0-1.0',
         'last_run': None,
         'data_source': 'RustDataSource',
```
pombredanne, line 30: I think we discussed it before ... IMHO I would prefer to have a real class rather than a class name.

vulnerabilities/importer_yielder.py
```diff
@@ -129,7 +129,7 @@
     },
     {
         'name': 'redhat',
-        'license': '',
+        'license': 'CC BY 4.0',
```
pombredanne, line 132: Please use SPDX license indentifiers OR scancode license identifiers.... But do not make something new.

PR #236, opened 2020-07-29
Use ayncio for making  API  calls to obtain package versions

sbs2001:
This works by  creating coroutines for all API calls at once in bulk. This cuts the time waiting for server to respond before making another request.


pombredanne:
Can you rebase this?

Issue #71, opened 2019-09-26
Collect CVE from NVD


pombredanne:
Of course :)
See also
- https://github.com/CVEProject/cvelist
- https://nvd.nist.gov/

singh1114:
@pombredanne @haikoschol I can take on this. Can we discuss how we need to proceed with this?

This seems to be a good source of data: https://nvd.nist.gov/vuln/data-feeds

pombredanne:
@singh1114 that would be awesome! let's bring it on the chat to discuss in details.

PR #243, opened 2020-08-10
Add NVD importer

sbs2001:
This branch is based on https://github.com/nexB/vulnerablecode/pull/227. 

Finally an importer for NVD. 

Fixes https://github.com/nexB/vulnerablecode/issues/71

### DIFF
--- a/vulnerabilities/importers/nvd.py
+++ b/vulnerabilities/importers/nvd.py
@@ -20,18 +20,18 @@
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
+import dataclasses
 import gzip
 import json
-import dataclasses
 from dateutil import parser as dateparser
 from datetime import date
 
 import requests
 
-from vulnerabilities.data_source import DataSource
-from vulnerabilities.data_source import Reference
 from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import DataSource
 from vulnerabilities.data_source import DataSourceConfiguration
+from vulnerabilities.data_source import Reference
 
 
 @dataclasses.dataclass
@@ -47,11 +47,14 @@ class NVDDataSource(DataSource):
     CONFIG_CLASS = NVDDataSourceConfiguration
 
     def updated_advisories(self):
-        years = [1]
         current_year = date.today().year
         # NVD json feeds start from 2002.
         for year in range(2002, current_year + 1):
             download_url = BASE_URL.format(year)
+            # Etags are like hashes of web responses. We maintain
+            # (url, etag) mappings in the DB. `create_etag`  creates
+            # (url, etag) pair. If a (url, etag) already exists then the code
+            # skips processing the response further to avoid duplicate work
             if self.create_etag(download_url):
                 data = self.fetch(download_url)
                 yield self.to_advisories(data)
--- a/vulnerabilities/importers/nvd.py
+++ b/vulnerabilities/importers/nvd.py
@@ -49,7 +49,7 @@ class NVDDataSource(DataSource):
     def updated_advisories(self):
         current_year = date.today().year
         # NVD json feeds start from 2002.
-        for year in range(2002, current_year + 1):
+        for year in range(2002, current_year+1):
             download_url = BASE_URL.format(year)
             # Etags are like hashes of web responses. We maintain
             # (url, etag) mappings in the DB. `create_etag`  creates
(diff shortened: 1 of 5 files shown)


vulnerabilities/importers/nvd.py
```diff
@@ -0,0 +1,153 @@
...
+import dataclasses
+from dateutil import parser as dateparser
+from datetime import date
+
+import requests
+
+from vulnerabilities.data_source import DataSource
+from vulnerabilities.data_source import Reference
+from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import DataSourceConfiguration
```
pombredanne, line 34: nit: can you sort your imports?

vulnerabilities/importers/nvd.py
```diff
@@ -0,0 +1,153 @@
...
+
+BASE_URL = "https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-{}.json.gz"
+
+
+class NVDDataSource(DataSource):
+
+    CONFIG_CLASS = NVDDataSourceConfiguration
+
+    def updated_advisories(self):
+        years = [1]
```
pombredanne, line 50: What is this magic number about?

vulnerabilities/importers/nvd.py
```diff
@@ -0,0 +1,153 @@
...
+        current_year = date.today().year
+        # NVD json feeds start from 2002.
+        for year in range(2002, current_year + 1):
+            download_url = BASE_URL.format(year)
+            if self.create_etag(download_url):
```
pombredanne, line 55: Can you elaborate what are etags about in comments?

vulnerabilities/importers/nvd.py
```diff
@@ -0,0 +1,153 @@
...
+                cve_id=cve_id, summary=summary, vuln_references=references, impacted_package_urls=[]
+            )
+
+    @staticmethod
+    def extract_summary(cve_item):
+        # In 99% of cases len(cve_item['cve']['description']['description_data']) == 1 , so
+        # this usually returns  cve_item['cve']['description']['description_data'][0]['value']
+        # In the remaining 1% cases this returns the longest summary.
+        summaries = [desc["value"] for desc in cve_item["cve"]["description"]["description_data"]]
+        return max(summaries, key=len)
```
pombredanne, line 86: What about joining them with '\n' to handle all the cases nicely?
```Python
summaries = (desc["value"] for desc in cve_item["cve"]["description"]["description_data"])
summaries = (s.strip() for s in summaries if s)
return "\n".join(s for s in summaries if s)
```

vulnerabilities/importers/nvd.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def extract_references(self, cve_item):
+        refs = []
+        for reference in cve_item["cve"]["references"]["reference_data"]:
+            ref_id = self.find_ref_id(reference)
+            ref_url = reference["url"]
+
+            # Skip references which exceed db constraints
+            if ref_id and len(ref_id) > 50:
```
pombredanne, line 95: Ouch ... if you have a reference that crashes your schema, then your schema needs to be updated. Do not skip the data!

vulnerabilities/importers/nvd.py
sbs2001: done

vulnerabilities/importers/nvd.py
sbs2001: Thanks your solution is better with a caveat, in such cases , the `Summary` would be the joined summaries
image

It probably won't make sense when reading such cases. 

Btw I only found only 1 CVE from NVD 2019 which had multiple(2) summaries.

vulnerabilities/importers/nvd.py
sbs2001: Oopsie, that's a remenant of past implementation. Removing.

vulnerabilities/importers/nvd.py
sbs2001: How does it look now? Does this make sense?

pombredanne: perfect!

vulnerabilities/importers/nvd.py
pombredanne: Your call... Since this is only  one case in many... this is best to have one weird summary and all other being correct (so IMHO the join is still better ;) )

vulnerabilities/importers/nvd.py
pombredanne: That's would mean a model change IMHO

sbs2001: Let me first find the long ref id first. 

I found some bugs after adding the length filter. Things may have changed

vulnerabilities/importers/nvd.py
sbs2001: There are bunch of them in older CVE records. Choosing the longest one is the way to go  IMHo

vulnerabilities/importers/nvd.py
sbs2001: The extraction of reference ids from  NVD is  complicated, it contains lots of noise, I've skipped collecting the reference ids for now .

pombredanne (approved):
LGTM! Thanks!

PR #247, opened 2020-08-21
 Redesign  the rest API

sbs2001:
TODOs 

1. Update documentation. ~Add tests specific to checking new API additions~
~2. Deal with inputting  packageurls as query param.~

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -46,11 +46,15 @@ class Vulnerability(models.Model):
 
     @property
     def vulnerable_to(self):
-        return self.packagerelatedvulnerability_set.filter(is_vulnerable=True)
+        return self.package_set.filter(
+            packagerelatedvulnerability__is_vulnerable=True,
+        )
 
     @property
     def resolved_to(self):
-        return self.packagerelatedvulnerability_set.filter(is_vulnerable=False)
+        return self.package_set.filter(
+            packagerelatedvulnerability__is_vulnerable=False,
+        )
 
     def __str__(self):
         return self.cve_id or self.summary
@@ -90,11 +94,15 @@ class Package(PackageURLMixin):
 
     @property
     def vulnerable_to(self):
-        return self.packagerelatedvulnerability_set.filter(is_vulnerable=True)
+        return self.vulnerabilities.filter(
+            packagerelatedvulnerability__is_vulnerable=True,
+        )
 
     @property
     def resolved_to(self):
-        return self.packagerelatedvulnerability_set.filter(is_vulnerable=False)
+        return self.vulnerabilities.filter(
+            packagerelatedvulnerability__is_vulnerable=False,
+        )
 
     class Meta:
         unique_together = ("name", "namespace", "type", "version", "qualifiers", "subpath")
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
+# Copyright (c) nexB Inc. and others. All rights reserved.
 # http://nexb.com and https://github.com/nexB/vulnerablecode/
 # The VulnerableCode software is licensed under the Apache License version 2.0.
 # Data generated with VulnerableCode require an acknowledgment.
@@ -18,7 +18,7 @@
 #  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
 #  VulnerableCode should be considered or used as legal advice. Consult an Attorney
 #  for any legal advice.
-#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  VulnerableCode is a free software tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 import importlib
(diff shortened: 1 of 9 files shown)


vulnerabilities/models.py
```diff
@@ -39,59 +39,87 @@ class Vulnerability(models.Model):
...
-        help_text='Summary of the vulnerability', blank=True)
-    cvss = models.FloatField(max_length=100, help_text='CVSS Score', null=True)
+
+    cve_id = models.CharField(max_length=50, help_text="CVE ID", unique=True, null=True)
+    summary = models.TextField(help_text="Summary of the vulnerability", blank=True)
+    cvss = models.FloatField(max_length=100, help_text="CVSS Score", null=True)
+
+    @property
+    def vulnerable_to(self):
+        qs = PackageRelatedVulnerability.objects.filter(
```
tdruez, line 49: Use Many2Many QuerySet syntax instead.

vulnerabilities/models.py
```diff
@@ -39,59 +39,87 @@ class Vulnerability(models.Model):
...
     """
+
     vulnerabilities = models.ManyToManyField(
-        to='Vulnerability', through='PackageRelatedVulnerability')
+        to="Vulnerability", through="PackageRelatedVulnerability"
+    )
+
+    @property
+    def vulnerable_to(self):
+        qs = PackageRelatedVulnerability.objects.filter(
```
tdruez, line 99: Same as above, use Many2Many syntax.

vulnerabilities/models.py
```diff
@@ -39,59 +39,87 @@ class Vulnerability(models.Model):
...
-            'Extra qualifying data for a package such as the name of an OS, '
-            'architecture, distro, etc.'
+            "Extra qualifying data for a package such as the name of an OS, "
+            "architecture, distro, etc."
         ),
-        null=True
+        null=True,
     )
 
     def set_package_url(self, package_url):
```
tdruez, line 125: Why are we overriding the default method from PackageURLMixin?

vulnerablecode/urls.py
```diff
@@ -27,13 +27,17 @@
...
+# `DefaultRouter` requires `basename` when registering viewsets which don't
+# define a queryset.
+api_router.register(r"vulnerabilities", VulnerabilityViewSet, basename="vulnerability")
 
 
 urlpatterns = [
-    path('admin/', admin.site.urls),
-    re_path(r'^api/', include(api_router.urls))
+    path("admin/", admin.site.urls),
+    re_path(r"^api/", include(api_router.urls)),
```
tdruez, line 42: What is the need for `re_path` instead of `path` here?

vulnerablecode/urls.py
```diff
@@ -27,13 +27,17 @@
 from rest_framework.routers import DefaultRouter
 
 from vulnerabilities.api import PackageViewSet
+from vulnerabilities.api import VulnerabilityViewSet
 
 
-api_router = DefaultRouter()
-api_router.register(r'packages', PackageViewSet)
+api_router = DefaultRouter(trailing_slash=False)
```
tdruez, line 33: What's the reasoning behind overriding the default trailing_slash convention?

tdruez (changes requested):
The API code looks good.
See my comments for models and urls.

vulnerabilities/models.py
sbs2001: This one uses JSONField the old`set_package_url` converts qualifiers into normalized string. More on this https://github.com/package-url/packageurl-python/pull/35#discussion_r434320271

vulnerablecode/urls.py
sbs2001: I haven't touched that, that's black doing it's thing.  Having said that  we can use `path` their alright.

vulnerablecode/urls.py
sbs2001: I had to discuss this with you guys. 

For the `ListView` using trailing_slash the endpoints look  odd. The endpoints with queries look like :
`/api/vulnerabilities/?vulnerability_id=foo`

After the override you get the :
`/api/vulnerabilities?vulnerability_id=foo`

Not sure this matter tho, so I might be missing something here.

tdruez: I don't see the benefit of removing the trailing slash.

From https://www.django-rest-framework.org/api-guide/routers/

> Trailing slashes are conventional in Django, but are not used by default in some other frameworks such as Rails.

I think we should follow Django's conventions.

sbs2001: :+1:

vulnerabilities/models.py
```diff
@@ -39,75 +39,87 @@ class Vulnerability(models.Model):
...
+    cve_id = models.CharField(max_length=50, help_text="CVE ID", unique=True, null=True)
+    summary = models.TextField(help_text="Summary of the vulnerability", blank=True)
+    cvss = models.FloatField(max_length=100, help_text="CVSS Score", null=True)
 
     @property
     def vulnerable_to(self):
-        return self.packagerelatedvulnerability_set.filter(is_vulnerable=True)
+        return self.package_set.filter(
+            packagerelatedvulnerability__is_vulnerable=True,
+            packagerelatedvulnerability__vulnerability_id=self.id,
```
tdruez, line 51: I don't think this lookup is necessary since the related manager is starting on `self`.

tdruez (changes requested):
Looks good. Although I don't think those lookups are necessary:
`packagerelatedvulnerability__vulnerability_id=self.id,`
and 
`packagerelatedvulnerability__package_id=self.id,`

Please confirm and update accordingly.

tdruez (approved):
LGTM!

pombredanne (approved):
LGTM :+1:

Issue #242, opened 2020-08-08
Refactor import_runner.py


sbs2001:
https://github.com/nexB/vulnerablecode/blob/develop/vulnerabilities/import_runner.py 

The module has lot of duplicate code. We will also get rid of import mechanism for  `added_advisories`  because it was based on the assumption that intersecting data can't be obtained  via different data sources. A bug in packageurl-python caused confirmation of the assumption. This bug in purl-python is fixed via https://github.com/package-url/packageurl-python/pull/40, it is expected to break our code for good.

PR #239, opened 2020-08-04
[DO NOT MERGE]Handle conflicts within in-memory advisories

sbs2001:
Due to  this we had to add 'is_vulnerable' in PackageRelatedVulnerability. This happened because we were unable to  handle conflicts within in-memory advisories. This is fixed and the db constraints are changed too

PR #255, opened 2020-09-14
 Refactor importer_runner.py

sbs2001:
There are alot of problems with importer_runner.py . 

It has lots of redundant code.

It is expected to break once we use the new models from purl mixin.

Fixes https://github.com/nexB/vulnerablecode/issues/242 and superset of https://github.com/nexB/vulnerablecode/pull/239

TODOs : 

- [x]  Use a smaller fixture than the current `arch.json` to reduce time to run tests

- [x]  Handle in memory conflicts


vulnerabilities/import_runner.py
```diff
@@ -109,169 +111,125 @@ def run(self, cutoff_date: datetime.datetime = None) -> None:
...
-                if adv_ref.url and not existing_ref.url:
-                    existing_ref.url = adv_ref.url
+def find_conflicting_relations(relations):
 
-                existing_ref.save()
+    relation_tuples = [(rel.package, rel.vulnerability) for rel in relations]
+    relation_counter = Counter(relation_tuples).most_common()
+    conflicts = set()
+    for rel, count in relation_counter:
+        if count < 2:
```
pombredanne, line 199: What is this `2`? The way I read this, you are trying to get a rel (which should should likely write as `for (vuln, pack), count` rather than `for rel, count`)... which as of the most common has only a single occurrence (meaning there could be more common ones we skip).... then you are create two objects `PackageRelatedVulnerabilityInserter` one is_vulnerable and one not is_vulnerable... Can you elaborate what you are trying to do there?

sbs2001: @pombredanne  Sorry this is  cryptic and needs some doc strings.

The point of the function `find_conflicting_relations` is to detect entries like `(pkg=foo, vuln=bar, is_vulnerable=True)` and `(pkg=foo, vuln=bar, is_vulnerable=False)` from EXISTING TOGETHER in the `db insertion que`.

To do this : 

1. Take in  the `db insertion que` which is a SET of `PackageRelatedVulnerabilityInserter` this is basically `(pkg, vuln, is_vulnerable)`

2. Create a list of tuples from `PackageRelatedVulnerabilityInserter` by chopping off the `is_vulnerable` flag. These tuples are of format `(pkg, vuln)`

3. Now make a Counter object from this list of tuples.  Now here's the key idea , for a count of a particular `(pkg, vuln)` to be `>=2` it needs to have `>=2`  `PackageRelatedVulnerabilityInserter` objects to be instantiated from in the first place. 
 Since `db insertion que` is a SET, for this to be true it implies that those   `PackageRelatedVulnerabilityInserter` objects would be `(pkg, vuln, True)` and  `(pkg, vuln, False)`  which is conflicting data.

4. We then recreate this conflicting data and bubble it up.

Also note that :

`relation_counter = Counter(relation_tuples).most_common()`
which means while iterating we encounter most common elements first. This implies `count(nth item) >= count((n+1)th item) `. If count is less 2 than all the entries ahead don't have conflicting entries and we halt the search.

pombredanne: Thanks! this make sense. I think there may be ways to improve this in the future, but that's a perfect explanation. Adding the comments in https://github.com/nexB/vulnerablecode/pull/255/commits/7f1f476cf003ae975134181e0bc117368c6f1e22 is A-OK.

pombredanne (approved):
LGTM!
Thanks

Issue #229, opened 2020-07-19
Better deployment, more convenient usage


inksong:
Have you thought about packaging the vulnerablecode into a docker or using other packaging methods?
I know we are in WIP state, but I still hope to deploy it more conveniently.
It now appears that for those who only care about data, the entire system is not packaged and easy to use.

pombredanne:
@inksong have you been able to test the new Dockerfile and Docker compose?

PR #256, opened 2020-09-20
Dockerize vulnerablecode

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/229

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python@sha256:e9b7e3b4e9569808066c5901b8a9ad315a9f14ae8d3949ece22ae339fff2cad0
 
 # PYTHONUNBUFFERED=1 ensures that the python output is set straight
 # to the terminal without buffering it first
@@ -7,3 +7,8 @@ RUN mkdir /vulnerablecode
 WORKDIR /vulnerablecode
 ADD . /vulnerablecode/
 RUN pip install -r requirements.txt
+
+LABEL "base_image": "pkg:docker/python@sha256%3Ae9b7e3b4e9569808066c5901b8a9ad315a9f14ae8d3949ece22ae339fff2cad0"
+LABEL "dockerfile_url":  "https://github.com/nexB/vulnerablecode/blob/develop/Dockerfile"
+LABEL "homepage_url":  "https://github.com/nexB/vulnerablecode"
+LABEL "license": "Apache-2.0"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
 FROM python:3.8
+
+# PYTHONUNBUFFERED=1 ensures that the python output is set straight
+# to the terminal without buffering it first
 ENV PYTHONUNBUFFERED 1
 RUN mkdir /vulnerablecode
 WORKDIR /vulnerablecode
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3.8
+ENV PYTHONUNBUFFERED 1
+RUN mkdir /vulnerablecode
+WORKDIR /vulnerablecode
+ADD . /vulnerablecode/
+RUN pip install -r requirements.txt
--- a/README.md
+++ b/README.md
@@ -1,145 +1,267 @@
+  
 
 <div align="center">
+
 <h1>VulnerableCode</h1>
 
+  
+
 [![Build Status](https://travis-ci.org/nexB/vulnerablecode.svg?branch=develop)](https://travis-ci.org/nexB/vulnerablecode)
+
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+
 [![Python 3.8](https://img.shields.io/badge/python-3.8-blue.svg)](https://www.python.org/downloads/release/python-360/)
+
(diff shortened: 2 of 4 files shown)


sbs2001:
This is a repaste of the issue I am facing from gitter :

```
All works fine if the project is ran using docker-compose but I want the project to work without docker too, for that to happen I have specify "HOST":"localhost" in the settings.py manually which is not ideal. How should I tell django where to look for db depending upon whether it's ran in a container or normally ? I'm thinking of using an env variable in the docker-compose file and add some logic in settings.py to interpret this env variable. Is there a better way ?


```

README.md
```diff
@@ -12,134 +15,404 @@
...
+  
+
+  
+
 VulnerableCode is a FOSS database of vulnerabilities and the FOSS packages they impact. It is made by the FOSS community to improve and secure the open source software ecosystem.
 
+  
+
+  
+
```
pombredanne, line 33: Why all the new blank lines?

README.md
```diff
@@ -12,134 +15,404 @@
...
+- Docker Engine. Find instructions to install it [here](https://docs.docker.com/get-docker/)
+
+- Docker Compose. Find instructions to install it [here](https://docs.docker.com/compose/install/#install-compose)
+
+  
+
+Use `sudo docker-compose up` to start VulnerableCode.
```
pombredanne, line 129: Do we want to promote using sudo there?

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: '3'
```
pombredanne, line 1: what the verion there? the one for the compose file?

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.8
+ENV PYTHONUNBUFFERED 1
```
pombredanne, line 2: Can you explain why this with a comment?

Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+RUN mkdir /vulnerablecode
+WORKDIR /vulnerablecode
+ADD . /vulnerablecode/
+RUN pip install -r requirements.txt
```
pombredanne, line 6: Can we add some labels to the image such that we know its base image, origin and license, author, maintainers, etc? (See how complicated a Docker image analysis can be with https://github.com/nexB/scancode.io/blob/main/scanpipe/pipelines/docker.py )

pombredanne (changes requested):
Thanks!
LGTM except for some weird spacing and a suggestion on Docker image.

README.md
sbs2001: Without `sudo` I get 
```
shivam@shivam-UX430UAR:~/coding/development/vulnerablecode$ docker-compose up
ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
```

docker-compose.yml
sbs2001: yes. They have that all over in their docs.

README.md
sbs2001: Sorry. Can you check now ?

Dockerfile
sbs2001: https://stackoverflow.com/questions/59812009/what-is-the-use-of-pythonunbuffered-in-docker-file . 

I'll add comments there

README.md
```diff
@@ -38,40 +38,45 @@ The web interface enables community curation of data by enabling addition of new
...
+git clone https://github.com/nexB/vulnerablecode.git
+cd vulnerablecode
 ```
 
-### System requirements
+### Using Docker Compose
+The easiest way to set up VulnerableCode is by using the docker containers.
+For this you need to have the following installed.
+- Docker Engine. Find instructions to install it here
+- Docker Compose. Find instructions to install it here
```
pombredanne, line 54: What about this:
```suggestion
An easy way to set up VulnerableCode is with docker containers and docker compose.
For this you need to have the following installed.
- Docker Engine at https://docs.docker.com/engine/install/
- Docker Compose at https://docs.docker.com/compose/install/
```

pombredanne (changes requested):
See my suggestions inline. Note that at this stage we may want to start using ReST for the README and docs like for the other projects ... but this should another PR.

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+
+services:
+  web:
+    environment:
+        - DJANGO_DEV=1
+        - VC_DB_HOST=db
+    build: .
+    command: bash -c "python manage.py makemigrations && python manage.py migrate && python manage.py runserver 0.0.0.0:8000"
```
tdruez, line 9: Why the "web" service have to "makemigrations"?

sbs2001: I'm not sure what alternative you are suggesting.

 `web` service has only the django code(not the db) hence I don't see any other way we can do migrations.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+LABEL "base_image": "pkg:docker/python@sha256%3Ae9b7e3b4e9569808066c5901b8a9ad315a9f14ae8d3949ece22ae339fff2cad0"
+LABEL "dockerfile_url":  "https://github.com/nexB/vulnerablecode/blob/develop/Dockerfile"
+LABEL "homepage_url":  "https://github.com/nexB/vulnerablecode"
+LABEL "license": "Apache-2.0"
```
pombredanne, line 14: This is awesome :)

pombredanne (approved):
LGTM++ :bow: 
Merging

Issue #232, opened 2020-07-23
On the identification of vulnerabilities


pombredanne:
We need to have some externally stable identifier for a vulnerability. 
By stable I mean that should be there forever and immutable so it can be used reliably to reference a vulnerability.

A CVE may not exist so we have a problem with using CVE everywhere.
When there is no CVE:
1. we could request a CVE id https://cveform.mitre.org/ and that's a manual process that not straightforward and we may not have yet all the data required to get a CVE id
2. we could become our CNA which is yet another level of "ceremony"
3. we could just assign our own ids and bypass 1. and 2. for the sake of speed, practicality and automation

I suggest that we use 3. and have our own prefix (TBD, such as VULCODE-xxxx) and we assign ids to a vulnerability this way:
- a. It has a CVE, then we reuse the `CVE-xxxx` as its identifier
- b. It does not have a CVE, then we create the `VULCODE-xxxx` as its identifier
- c. later if that `VULCODE-xxx` vulnerability gets a CVE, we will replace the id with the `CVE-xxx` id and move the `VULCODE-xxx` id as a reference for that vulnerability.

From a usage standpoint, this means that we should be able to search a vulnerability not only based on its identifier (that may change over time) but also based on its references.

Alternatively to c. above we could have a dedicated field to store the previous `VULCODE-xxx` when this is replaced by an assigned  `CVE-xxx`id.

The model could look like this
- Vulnerability
   - identifier: `VULCODE-xxx`  or  `CVE-xxx`
   - vc_identifier: empty or `VULCODE-xxx`  if identifier is a `CVE-xxx`

Later we can have processes for 1. and/or 2. but that's completely independent

sbs2001:
I don't know whether this is same as using hashes but here it goes : 

We will use your suggested Vulnerability model. To make this work we will need the identifier to have some field to denote the importer which encountered the CVE-less vulnerability, eg VULCODE-NPM-0001 (TBD) , which tells me this vulnerability was encountered by https://github.com/nexB/vulnerablecode/blob/develop/vulnerabilities/importers/npm.py source . 

So every other instance will have different IDs for the same Vulnerability, this poses the problem in referencing the Vulnerability which is what your stability point is about. 

The easiest way we can fix this is : 

Start a git repo where we put VulnerableCode Central's (TBD)  all VULCODE-XXXX entries. The directory structure could look like (again TBD)

VULCODES/NPM/0001.json
VULCODES/PYPI/0001.json

[...]

PR #259, opened 2020-09-26
Handle vulnerabilities which don't have any vulnerability ids 

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/232

- [x] Get the `Vulnerability` model and it's methods right. Use a timestamp as a custom id for now.
- [x] Refactor all the codebase to comply to the new `Vulnerability` model.
- [x] Create a repo which would contain  all vulnerabilities with custom ids.
- [x]  Add option to `VulnerableCode` to either assign these ids or sync the ids from  the repo.
- [x] Automate dumping the vulnerabilities with custom ids to the repo
- [x] Decide the custom id and implement a data structure to make it feasible.


sbs2001:
For the models the requirements as we decided upon at https://github.com/nexB/vulnerablecode/issues/232 are (this is a partial repaste of @pombredanne 's comment)

- a. It has a CVE, then we reuse the `CVE-xxxx` as its identifier
- b. It does not have a CVE, then we create the `VULCODE-xxxx` as its identifier
- c. later if that `VULCODE-xxx` vulnerability gets a CVE, we will replace the id with the `CVE-xxx` id and move the `VULCODE-xxx` id as a reference for that vulnerability.

From a usage standpoint, this means that we should be able to search a vulnerability not only based on its identifier (that may change over time) but also based on its references.

Alternatively to c. above we could have a dedicated field to store the previous `VULCODE-xxx` when this is replaced by an assigned  `CVE-xxx`id.


The model could look like this
- Vulnerability
   - identifier: `VULCODE-xxx`  or  `CVE-xxx`
   - vc_identifier: empty or `VULCODE-xxx`  if identifier is a `CVE-xxx`

sbs2001:
The current models give me this 

```

In [2]: z = models.Vulnerability.objects.create(summary="tar ball")                                                                                                                                                  

In [3]: z.identifier                                                                                                                                                                                                 
Out[3]: '2020-09-26 06:32:18.832135'

```

sbs2001:
I'm using https://github.com/sbs2001/vulcodes as the repo for dev purpose.

pombredanne:
I am wondering ..."VULCODE-XXX" feels a little bit too generic, and I am kinda warming up to a the explicit albeit longer "VULNERABLECODE-XXX" prefix even if a tad long... it is explicit and meant to be replaced eventually by a CVE (and it is also very clear where it is coming from)?

vulnerabilities/management/commands/push.py
```diff
@@ -0,0 +1,93 @@
...
+        repo_url = options["remote_url"]
+        # TODO; Do some validation of `repo_url` here
+        push_data(repo_url)
+
+
+def push_data(url):
+    repo_location = tempfile.mkdtemp()
+    with cd(repo_location):
+        os.system(f"git clone {url}")
+        # TODO: Don't hardcode `vulcodes`
```
pombredanne, line 84: You should should use a location argument instead

pombredanne (commented):
Thanks!
Looking good except for the "vulcode/VULCODE" name that I would like to discuss more.

pombredanne:
> The model could look like this

    Vulnerability
        identifier: VULCODE-xxx or CVE-xxx
        vc_identifier: empty or VULCODE-xxx if identifier is a CVE-xxx


I wonder if we should not instead just treat the "past" vc_identifier as just an external reference?

pombredanne:
I revisited https://cve.mitre.org/data/refs/index.html  and I suggest this instead prefix`VULCOID` as a prefix which is rather unique and would be short for Vulnerable Code Id.
We could have ids in this form then where the left part is an ISO-like time stamp e.g. `VULCOID-YYYY-MM-DD-HH-MM-SS`
The id would case insensitive with an uppercase canonical form.
Would this work?

sbs2001:
@pombredanne   `VULCOID-YYYY-MM-DD-HH-MM-SS` works, have you considered  `VULCOID-YYYY-<incremental id>` ie something like CVE . FWIW It's easier to remember than the timestamp thus easy to reference.

pombredanne:
@sbs2001 re https://github.com/nexB/vulnerablecode/pull/259#issuecomment-734844259

> `VULCOID-YYYY-MM-DD-HH-MM-SS` works, have you considered `VULCOID-YYYY-<incremental id>` ie something like CVE . FWIW It's easier to remember than the timestamp thus easy to reference.

incremental id works too but requires more coordination than a context free timestamp. I am not sure a sequential number is easier to remember than a time stamp though it can be shorter.
We could go with:
 `VULCOID-YYYY-MM-DD-<incremental id>` as a compromise?

sbs2001:
@pombredanne 
>incremental id works too but requires more coordination than a context free timestamp

How so ? 

The timestamp is dependent on when the instance encountered the vulnerability. coordination is needed in that case too if we want 2 instances to "have a common language" .

IMHO coordination is unavoidable.

pombredanne:
IMHO `VULCOID-YYYY-MM-DD-HH-MM-SS` is the scheme that requires no coordination and creates the least risk of collision and is still somewhat readable by a human. If there are collision across DBs that's OK.

sbs2001:
@pombredanne It made sense to me about using timestamps instead of incremental ids after the chat. Main reason being, when using `incremental id's` it is guaranteed that collisions would occur ( between different instances) and hence they would the be needed to be resolved. This won't happen when using timestamps.

sbs2001:
@pombredanne  I am stashing the code to extract vulcoids to some other branch. This PR won't add that, that feature seems pre-mature to me.

vulnerabilities/importer_yielder.py
```diff
@@ -61,16 +61,16 @@
...
-#         'data_source_cfg': {
-#             'url': 'https://raw.githubusercontent.com/pyupio/safety-db/master/data/insecure_full.json',  # nopep8
-#             'etags': {}
-#         },
-#     },
+    {
+        'name': 'safetydb',
+        'license': 'cc-by-nc-4.0',
+        'last_run': None,
+        'data_source': 'SafetyDbDataSource',
```
pombredanne, line 68: Why re-enable this here?

vulnerabilities/importers/safety_db.py
```diff
@@ -25,16 +25,17 @@
 
 import asyncio
 import dataclasses
-import json
+import re
+import logging
```
pombredanne, line 29: nit: import sort order is not right... may be should use `isort`

pombredanne (changes requested):
Thanks! See my comments inline. IMHO the main issue we have is the name "identifier"
Either we use that everywhere, of may be we use `vulnerability_id` instead which would be more explicit and work in all the contexts.

vulnerabilities/import_runner.py
```diff
@@ -139,10 +139,20 @@ def process_advisories(data_source: DataSource) -> None:
...
     advisory_batches = chain(data_source.updated_advisories(), data_source.added_advisories())
     for batch in advisory_batches:
         for advisory in batch:
             try:
+
+                if not advisory.identifier:
+                    advisory.identifier = "VULCOID-" + vulcoid_timestamp.strftime(
+                        "%Y-%m-%d-%H:%M:%S"
+                    )
+                    vulcoid_timestamp = max(
```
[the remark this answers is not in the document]
sbs2001: I've added a comment for this, hopefully it makes sense now. Adding a function for this would blowup the complexity, because then we have to maintain the state of vulcoid throughout.

vulnerabilities/api.py
```diff
@@ -208,13 +208,13 @@ def bulk_search(self, request):
                 },
             )
 
-        for cve_id in request.data["vulnerabilities"]:
-            filter_list.append(cve_id)
+        for vulnerability_id in request.data["vulnerabilities"]:
```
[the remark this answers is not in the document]
sbs2001: The bulk apis need rewrite and follow the design we discussed earlier. I'll do that in another PR.

vulnerabilities/api.py
```diff
@@ -63,7 +63,7 @@ class Meta:
 
 
 class HyperLinkedVulnerabilitySerializer(serializers.HyperlinkedModelSerializer):
-    vulnerability_id = serializers.CharField(source="cve_id")
+    vulnerability_id = serializers.CharField(source="identifier")
```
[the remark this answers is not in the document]
sbs2001: The models has name of `Vulnerability` which serves as namespace. I remember you pointed out the problem with having something like `Vulnerablity.id` is that `id` is a python keyword. 

Having said that in the context of serializers having a field called `vulnerability_id`  which siphons data from `Vulnerability.identifier` doesn't feel problematic ?

vulnerabilities/data_source.py
```diff
@@ -531,6 +531,6 @@ def get_data_from_xml_doc(self, xml_doc: ET.ElementTree, pkg_metadata={}) -> Lis
                     summary=description,
                     impacted_package_urls=affected_purls,
                     resolved_package_urls=safe_purls,
-                    cve_id=vuln_id,
+                    identifier=vuln_id,
```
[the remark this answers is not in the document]
sbs2001: I think I got the naming wrong wrt `Advisory` dataclass. How about `Advisory.vulnerability_id`  ? that's verbose but explicit

vulnerabilities/migrations/0001_initial.py
```diff
@@ -47,7 +47,8 @@ class Migration(migrations.Migration):
             name='Vulnerability',
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('cve_id', models.CharField(help_text='CVE ID', max_length=50, null=True, unique=True)),
+                ('identifier', models.CharField(help_text='CVE_ID or VC_ID', max_length=50, null=True, unique=True)),
```
pombredanne, line 50: What about this help:

> Unique identifier for a vulnerability: this is either a published CVE id (as in CVE-2020-7965) if it exists. Otherwise this is a VulnerableCode-assigned VULCOID (as in VULCOID-2021-01-23-15-12). When a vulnerability CVE is assigned later we replace this with the CVE and keep the "old " VULCOID in the "old_vulnerability _id" field to support redirection to the CVE id.

vulnerabilities/models.py
```diff
@@ -39,8 +42,48 @@ class Vulnerability(models.Model):
...
+        help_text="empty if no  CVE else VC id",
+        unique=True,
+        null=True,
+    )
+    summary = models.TextField(
+        help_text="Summary of the vulnerability",
+        blank=True,
+    )
+
+    def save(self, *args, **kwargs):
```
pombredanne, line 65: @tdruez Shivam and I crafted this approach overwrite save and would like to get your review.

Here is the context:

A Vulnerability has an a `vulnerabilty_id` which is the CVE number when this exists. Otherwise we assign a unique id using `generate_vulcoid`. Since we use a string and not a number or UUID, but something based on a timestamp, we retry until we can get an id that is unique in the rare cases where we have collisions.

vulnerabilities/models.py
```diff
@@ -39,8 +42,48 @@ class Vulnerability(models.Model):
...
+        if self.vulnerability_id:
+            return super().save(*args, **kwargs)
+        # Generate unique VULCOID
+        ie = None
+        for attempt in range(1, 11):
+            try:
+                self.vulnerability_id = self.generate_vulcoid()
+                # Using the context manager due to https://stackoverflow.com/a/23326971
+                with transaction.atomic():
```
pombredanne, line 74: I am somewhat concerned about starting a transaction inside the save()

vulnerabilities/import_runner.py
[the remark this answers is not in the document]
sbs2001: We moved this to models.py

vulnerabilities/import_runner.py
```diff
@@ -154,8 +164,8 @@ def process_advisories(data_source: DataSource) -> None:
                     if vuln_created or not vuln_ref_exists(
                         vuln, vuln_ref.url, vuln_ref.reference_id
                     ):
-                        # A vulnerability reference can't exist if the
-                        # vulnerability is just created so insert it
+                        # A vulnerability reference can't exist if the vulnerability is
```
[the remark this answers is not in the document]
sbs2001: Fixed via new implementation at models.py

vulnerabilities/models.py
sbs2001: I share you concern too. 
For some reason without that transaction block I get a :
`django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.`

There could be a deeper issue maybe ?

vulnerabilities/api.py
```diff
@@ -177,7 +176,7 @@ def bulk_search(self, request):
 
 
 class VulnerabilityFilterSet(filters.FilterSet):
-    vulnerability_id = filters.CharFilter(field_name="cve_id")
+    vulnerability_id = filters.CharFilter(field_name="vulnerability_id")
```
tdruez, line 179: `field_name` seems unneeded since duplicated from the attribute.

vulnerabilities/models.py
tdruez: Going with a transaction management approach here, you are asking for bigger issues than the one you are trying to solve initially.
If you want to go with the timestamp approach, I would simply add the microseconds (".%f") to the string to ensure unicity.

tdruez (changes requested):

vulnerabilities/models.py
sbs2001: @tdruez  that would solve the issue, but we'd have longer ids, @pombredanne  ^ would that be too ugly ?

sbs2001: @tdruez  what would be cleaner way to do as @pombredanne  said 

> we retry until we can get an id that is unique in the rare cases where we have collisions

tdruez: > @sbs2001 but we'd have longer ids

Keep only 2 digits of the micro seconds then, you'll have something unique enough and only 1 char extra compared to the incremental id approach.

sbs2001: @tdruez  that could make the collisions even rarer. But there would be a still (smaller) chance of collisions ? I am fine with whatever id we use. The code to deal with exceptions, especially the requirement of `transaction` is weird.

tdruez: I don't think it's likely that you will ever generate 2 vulcoid on the same micro second.
Encountering collisions with an arbitrary range of 10 integers is more likely.

tdruez: Current `strftime("%Y-%m-%d%H%M%S")` -> `VULCOID-2021-02-16084802`

Suggested `strftime("%Y%m%d-%H%M-%S%f")[:-4]` -> `VULCOID-20210216-0848-3008`

sbs2001: Still collisions happen.

tdruez: What about keeping all the micro seconds? ie removing the [:-4]

tdruez: Alternatively, in place of the micro seconds, we could use a slice of a random UUID.

sbs2001: > What about keeping all the micro seconds?

I used that approach while prototyping this whole thing, https://github.com/nexB/vulnerablecode/blob/787ab6e4778b50ddeed6f64146f3a6f58a4f57b2/vulnerabilities/models.py#L48  and it works ! 
It looks ugly but it works :)

pombredanne: OK, let's be ugly but working then! Go for it.

vulnerabilities/importers/safety_db.py
```diff
@@ -103,29 +101,40 @@ def updated_advisories(self) -> Set[Advisory]:
...
                 )
 
-                cve_ids = advisory.get("cve") or [""]
-
-                # meaning if cve_ids is not [''] but either ['CVE-123'] or ['CVE-123, CVE-124']
-                if len(cve_ids[0]):
-                    cve_ids = [s.strip() for s in cve_ids.split(",")]
+                if advisory["cve"]:
+                    # Check on advisory["cve"] instead of using `get` because it can have null value
+                    cve_ids = re.findall(r"CVE-\d+-\d+", advisory["cve"])
```
pombredanne, line 128: nit: FWIW, I tend to prefer having this as external functions defined this way:
```
import re

get_cve_ids = re.compile(r"CVE-\d+-\d+", re.IGNORE_CASE).findall
```
Then you can use get_cve_ids and test it too

vulnerabilities/importers/safety_db.py
```diff
@@ -103,29 +101,40 @@ def updated_advisories(self) -> Set[Advisory]:
...
+            try:
+                validate_schema(self._api_response[package_name])
+
+            except Exception as e:
+                logger.error(e)
+                continue
+
             all_package_versions = self.versions.get(package_name)
-            if len(all_package_versions) == 0:
+            if not len(all_package_versions):
```
pombredanne, line 117:
```suggestion
            if not all_package_versions:
```

vulnerabilities/importers/safety_db.py
```diff
@@ -103,29 +101,40 @@ def updated_advisories(self) -> Set[Advisory]:
...
                 impacted_purls, resolved_purls = categorize_versions(
                     package_name, all_package_versions, advisory["specs"]
                 )
 
-                cve_ids = advisory.get("cve") or [""]
-
-                # meaning if cve_ids is not [''] but either ['CVE-123'] or ['CVE-123, CVE-124']
-                if len(cve_ids[0]):
-                    cve_ids = [s.strip() for s in cve_ids.split(",")]
+                if advisory["cve"]:
```
pombredanne, line 126: May be create a variable

vulnerabilities/models.py
```diff
@@ -39,8 +42,38 @@ class Vulnerability(models.Model):
     VulnerabilityReference.
     """
 
-    cve_id = models.CharField(max_length=50, help_text="CVE ID", unique=True, null=True)
-    summary = models.TextField(help_text="Summary of the vulnerability", blank=True)
+    vulnerability_id = models.CharField(
+        max_length=50,
+        help_text="Unique vulnerability_id for a vulnerability: this is either a published CVE id"
```
pombredanne, line 47:
```suggestion
        help_text="Unique identifier for a vulnerability: this is either a published CVE id"
```

vulnerabilities/models.py
```diff
@@ -39,8 +42,38 @@ class Vulnerability(models.Model):
...
+    summary = models.TextField(
+        help_text="Summary of the vulnerability",
+        blank=True,
+    )
+
+    def save(self, *args, **kwargs):
+        if self.vulnerability_id:
+            return super().save(*args, **kwargs)
+        self.vulnerability_id = self.generate_vulcoid()
+        return super().save(*args, **kwargs)
```
pombredanne, line 69: What about this instead:
```Python
        if not self.vulnerability_id:
            self.vulnerability_id = self.generate_vulcoid()
        return super().save(*args, **kwargs)
```

pombredanne (approved):
All good ... I have a few minor nit pickings but this is good to merge!
Thanks!

vulnerabilities/models.py
sbs2001: Done

vulnerabilities/models.py
sbs2001: Done

vulnerabilities/importers/safety_db.py
sbs2001: Doing the safetydb changes in other pr

vulnerabilities/importers/safety_db.py
sbs2001: Doing the safetydb changes in other pr

vulnerabilities/importers/safety_db.py
sbs2001: I think this belongs to a helpers module, from where we can reuse this

sbs2001: Doing this in other PR

vulnerabilities/api.py
sbs2001: Good catch :) fixing this

PR #260, opened 2020-09-27
Add importer for project KB MSR2019 dataset

sbs2001:

### DIFF
--- a/vulnerabilities/importers/project_kb_msr2019.py
+++ b/vulnerabilities/importers/project_kb_msr2019.py
@@ -1,0 +1,96 @@
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import csv
+import dataclasses
+import urllib.request
+
+# Reading CSV file from  a url using `requests` is bit too complicated.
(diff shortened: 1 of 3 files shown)


vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -0,0 +1,96 @@
...
+        return csv.reader(lines)
+
+    @staticmethod
+    def to_advisories(csv_reader):
+        # Project KB MSR csv file has no header row
+        all_advsiories = []
+        for row in csv_reader:
+            vuln_id, proj_home, fix_commit, _ = row
+            commit_link = proj_home + "/commit/" + fix_commit
+            all_advsiories.append(
```
pombredanne, line 87: nit: all_advsiories -> all_advisories, also why not just call this "advisories" ?

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -0,0 +1,96 @@
...
+        response = urllib.request.urlopen(self.url)
+        lines = [l.decode("utf-8") for l in response.readlines()]
+        return csv.reader(lines)
+
+    @staticmethod
+    def to_advisories(csv_reader):
+        # Project KB MSR csv file has no header row
+        all_advsiories = []
+        for row in csv_reader:
+            vuln_id, proj_home, fix_commit, _ = row
```
pombredanne, line 85: What's `_` here?

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -0,0 +1,96 @@
...
+            commit_link = proj_home + "/commit/" + fix_commit
+            all_advsiories.append(
+                Advisory(
+                    summary="",
+                    impacted_package_urls=[],
+                    vuln_references=[Reference(url=commit_link)],
+                    cve_id=vuln_id,
```
pombredanne, line 92: This is likely to become "identifier", correct?

pombredanne (changes requested):
Thanks. See my few nits inline.
I wonder if the format will be different in "kaybee" ?

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: yes

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: https://github.com/SAP/project-kb/blob/master/MSR2019/dataset/vulas_db_msr2019_release.csv

It's `pos` for all rows. No idea what it means.

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: Done

vulnerabilities/importers/project_kb_msr2019.py
pombredanne: Note that FWIW, you can always write:
`for vuln_id, proj_home, fix_commit, _ in csv_reader:`

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -0,0 +1,96 @@
...
+        lines = [l.decode("utf-8") for l in response.readlines()]
+        return csv.reader(lines)
+
+    @staticmethod
+    def to_advisories(csv_reader):
+        # Project KB MSR csv file has no header row
+        advsiories = []
+        for row in csv_reader:
+            vuln_id, proj_home, fix_commit, _ = row
+            commit_link = proj_home + "/commit/" + fix_commit
```
pombredanne, line 86: Nit: I would tend to prefer a .format or an f-string.

pombredanne (approved):
All good. See some FYI.

Issue #99, opened 2019-09-26
Collect HTTPD 


pombredanne:
https://httpd.apache.org/security_report.html

sbs2001:
Closed via https://github.com/nexB/vulnerablecode/pull/261

PR #261, opened 2020-09-27
 Add Apache HTTPD advisory importer

sbs2001:

### DIFF
--- a/vulnerabilities/importers/apache_httpd.py
+++ b/vulnerabilities/importers/apache_httpd.py
@@ -1,0 +1,106 @@
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from dataclasses import dataclass
+import xml.etree.ElementTree as ET
+
+import requests
+from packageurl import PackageURL
+
+from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import DataSource
+from vulnerabilities.data_source import DataSourceConfiguration
+
+
+@dataclass
+class ApacheHTTPDDataSourceConfiguration(DataSourceConfiguration):
+    etags: dict
+
+
+class ApacheHTTPDDataSource(DataSource):
+
+    CONFIG_CLASS = ApacheHTTPDDataSourceConfiguration
+    url = "https://httpd.apache.org/security/vulnerabilities-httpd.xml"
+
+    def updated_advisories(self):
+        # Etags are like hashes of web responses. We maintain
+        # (url, etag) mappings in the DB. `create_etag`  creates
+        # (url, etag) pair. If a (url, etag) already exists then the code
+        # skips processing the response further to avoid duplicate work
+
+        if self.create_etag(self.url):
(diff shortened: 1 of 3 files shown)


sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/99

vulnerabilities/importers/apache_httpd.py
```diff
@@ -0,0 +1,106 @@
...
+        for info in issue:
+            if info.tag == "cve":
+                cve = info.attrib["name"]
+
+            if info.tag == "title":
+                summary = info.text
+
+            if info.tag == "fixed":
+                resolved_packages.append(
+                    PackageURL(name="httpd", version=info.attrib["version"], type="generic")
```
pombredanne, line 84: that's a nit but I prefer putting args in their original order, e.g. `type` first.

vulnerabilities/importers/apache_httpd.py
```diff
@@ -0,0 +1,106 @@
...
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from dataclasses import dataclass
+import xml.etree.ElementTree as ET
```
pombredanne, line 24: What's the benefit of  a shorter ET import?
IMHO for readability and clarity I would use `from import xml.etree import ElementTree` instead

vulnerabilities/importers/apache_httpd.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if self.create_etag(self.url):
+            data = fetch_xml(self.url)
+            advisories = to_advisories(data)
+            return self.batch_advisories(advisories)
+
+        return []
+
+    def create_etag(self, url):
+        etag = requests.head(url).headers.get("ETag")
```
pombredanne, line 58: Just FYI, at some point of time, we should review together the etag story and why this is used in parallel of https://github.com/nexB/vulnerablecode/pull/261/files#diff-941e9f048d96b3af4998962020f96430R105 and not with it?
I would like to make sure that I grok the caching approach alright.

vulnerabilities/importers/apache_httpd.py
pombredanne: Also we may want to have something more distinctive:
- what about a namespace or `apache` or `apache.org`?
- or/and a qualifier for `download_url`?
- or if there are more than just httpd, we could create an `apache` type for all Apache Foundation projects? after all they (all Apache projects) share many characteristics that would make them Package URL-type worthy 
See https://svn.apache.org/repos/asf/comdev/projects.apache.org/trunk/data/projects.xml for instance

vulnerabilities/importers/apache_httpd.py
```diff
@@ -0,0 +1,106 @@
...
+                summary = info.text
+
+            if info.tag == "fixed":
+                resolved_packages.append(
+                    PackageURL(name="httpd", version=info.attrib["version"], type="generic")
+                )
+
+            if info.tag == "affects" or info.tag == "maybeaffects":
+                impacted_packages.append(
+                    PackageURL(name="httpd", version=info.attrib["version"], type="generic")
```
pombredanne, line 89: Same nit as above wrt. args order and Apache.

pombredanne (changes requested):
See a few nits inline.
Looking great otherwise!

vulnerabilities/importers/apache_httpd.py
sbs2001: Adding the namespaces, and qualifiers seems fine.

About creating `apache` type, that I don't think is a good idea because that would end up creating ambiguity in case an apache package is packaged, eg `apache-httpd.rpm`. What would then be package type of  `apache-httpd.rpm` ?

vulnerabilities/importers/apache_httpd.py
sbs2001: >What's the benefit of a shorter ET import? 

Nothing. 

>IMHO for readability and clarity I would use from import xml.etree import ElementTree instead


That works.

FWIW using the alias `ET` is a common practice. The docs used it https://docs.python.org/3/library/xml.etree.elementtree.html .

vulnerabilities/importers/apache_httpd.py
sbs2001: Sure thing. 

The gist of it is: `create_tag` uses a `HEADER` request to just fetch the etag. If this etag matches with old etag we stop the process because the data hasn't changed. Otherwise  use https://github.com/nexB/vulnerablecode/pull/261/files#diff-941e9f048d96b3af4998962020f96430R105 which is just normal download.

vulnerabilities/importers/apache_httpd.py
pombredanne: >What would then be package type of apache-httpd.rpm ?

That's always an `rpm` type then.
The rationale is the format of packaging and metadata determines the type

sbs2001: Using `type=apache name=httpd, version=x`

vulnerabilities/importers/apache_httpd.py
sbs2001: Using `ElementTree`

pombredanne (approved):
LGTM! Thank you for your patience!

Issue #96, opened 2019-09-26
Collect nginx 


pombredanne:
http://nginx.org/en/security_advisories.html

sbs2001:
Closed via https://github.com/nexB/vulnerablecode/pull/264

PR #264, opened 2020-10-07
Use packageurl version 0.9.3 and Add nginx importer

sbs2001:

### DIFF
--- a/vulnerabilities/importers/nginx.py
+++ b/vulnerabilities/importers/nginx.py
@@ -1,0 +1,190 @@
+# Copyright (c)  nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICE
+# Unless required by applicable law or agreed to in writing, software dist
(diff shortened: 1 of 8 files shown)


sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/96

vulnerabilities/importers/nginx.py
```diff
@@ -0,0 +1,190 @@
...
+#  VulnerableCode should be considered or used as legal advice. Consult an
+#  for any legal advice.
+#  VulnerableCode is a free software  from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import asyncio
+import dataclasses
+import json
+
+import requests
```
pombredanne, line 27: Nit: we should sort imports. But we should also switch to using black and isort, so we can do it all at once!

vulnerabilities/importers/nginx.py
```diff
@@ -0,0 +1,190 @@
...
+                    summary=summary,
+                    impacted_package_urls=vulnerable_packages,
+                    resolved_package_urls=fixed_packages,
+                )
+            )
+
+        return advisories
+
+    def extract_fixed_pkgs(self, vuln_info):
+        vuln_status, version_info = vuln_info.split(": ")
```
pombredanne, line 133: Using a `.partition()` is more robust as it always does a single split and always return three values making the tuple unpacking safer and more reliable. I tend to prefer it followed by a `.strip()` call.

pombredanne (approved):
I am merging this leave aside the few nitpickings... we should definitely add tests though!
Thanks!

PR #270, opened 2020-10-23
[Experiment] Store and evaluate package version ranges

sbs2001:
- Add a table VulnerablePackageVersionRange
- This table stores version less purl string and a FK of vulnerability
- Added manager method which on given  a purl string which does not exist in the db
  tries to find whether it satisfies a version  range, if yes then that vulnerability is returned


vulnerabilities/data_source.py
```diff
@@ -56,6 +56,21 @@ def __post_init__(self):
         if not any([self.url, self.reference_id]):
             raise TypeError
 
+@dataclasses.dataclass
+class VulnPkgVersionRange:
+
+# This dataclass is nested inside the `Advisory` class below. It is used
```
pombredanne, line 62: Should this be a docstring instead of a comment?

vulnerabilities/data_source.py
```diff
@@ -56,6 +56,21 @@ def __post_init__(self):
...
+# package to the db insertion mechanism. 
+
+# The `purl` in  this case is a package url without a version. Eg value `pkg:maven/axis/axis`
+
+# The `version_ranges` contains a comma separated version  range expressions. Eg value `<4.1.4,>=4.0.0`
+# for which  the `purl` is vulnerable to some vulnerability. This `vulnerability` is specified by the parent `Advisory`
+# dataclass.
+
+    purl: str
+    version_ranges: str
```
pombredanne, line 73: I called this a requirement in scancode https://github.com/nexB/scancode-toolkit/blob/ba4bbf216c6f44572662d16c76214a08b0a69e7e/src/packagedcode/models.py#L331
I wonder if we should rename it in scancode or here.... but in anycase I would like to keep in a single common name for the same concept.

vulnerabilities/data_source.py
```diff
@@ -56,6 +56,21 @@ def __post_init__(self):
         if not any([self.url, self.reference_id]):
             raise TypeError
 
+@dataclasses.dataclass
+class VulnPkgVersionRange:
```
pombredanne, line 60: Could we find a better name?
PackageSpecification may be? what else could be a good one?

vulnerabilities/data_source.py
```diff
@@ -70,9 +85,11 @@ class Advisory:
     """
     summary: str
     impacted_package_urls: Iterable[PackageURL]
-    resolved_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
+    resolved_package_urls: Iterable[PackageURL] 
+    vuln_pkg_version_ranges : Iterable[VulnPkgVersionRange]
```
pombredanne, line 89: I reckon you would want that `vuln_pkg_version_ranges` be optional? and replace the the `impacted_package_urls` ?
Or would we keep both?
Could you have both at the same time?
And if we have bot would these be rather names something like `impacted_package_urls_specification` , or ``impacted_package_urls_requirements` or `impacted_package_urls_version ranges` or ...?`

vulnerabilities/importer_yielder.py
```diff
@@ -179,7 +179,7 @@
         'data_source': 'GitHubAPIDataSource',
         'data_source_cfg': {
             'endpoint': 'https://api.github.com/graphql',
-            'ecosystems': ['MAVEN', 'NUGET', 'COMPOSER']
+            'ecosystems': ['MAVEN', 'NUGET']
```
pombredanne, line 182: Why dropping composer from here?

pombredanne (changes requested):
Thanks! I reckon this still an experiment and a WIP, but getting some more tests here would be essential. I think we will need this for sure.
I added some other comments for your consideration.

sbs2001:
@pombredanne  Thanks for the review. 

What kind of  cases should be covered by these tests ?

Re. naming  no strong opinions  here but IMHO `requirement` isn't as intuitive as `version_ranges` .

vulnerabilities/importer_yielder.py
pombredanne: And actually why even filtering based on "ecosystems" ? Why not get them all?

pombredanne:
> @pombredanne Thanks for the review.
> What kind of cases should be covered by these tests ?
> Re. naming no strong opinions here but IMHO `requirement` isn't as intuitive as `version_ranges` 

I am fine with `version_ranges`. The key is  consistency, so if you go with it, we need to adopt that in scancode-toolkit and scancode.io too. Can you also enter tickets there? (and link back here too?)

PR #271, opened 2020-10-30
Update docs

sbs2001:


README.rst
```diff
@@ -143,10 +143,32 @@ Tests
...
 -----------
 
+Many data importers use GitHub APIs. For this, first set up value of the ``GH_TOKEN`` environment variable by running :
+
+::
+
+    export GH_TOKEN=yourgithubtoken
+
+
+See `GitHub docs  <https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token>`_ for instructions on how to obtain your GitHub token.
```
pombredanne, line 153: Could we make this optional? Is there some level of API calls that can be dome without a token?

sbs2001: Without the GH token, the github importer (and any other importers leveraging GH API) will error out. Is this worth mentioning in README ?

pombredanne: that's ok for now.

pombredanne (approved):

Issue #244, opened 2020-08-19
Cannot import safetydb


rolfschr:
Hi,

I setup everything locally (tests & migrations run fine) and started the import process. I cannot seem to import safetydb. Please find below the error log. The likely culprit is [here](https://github.com/pyupio/safety-db/blob/master/data/insecure_full.json#L2640).
```
$ DJANGO_DEV=1 python manage.py import --all
Importing data from rust
Successfully imported data from rust
Importing data from alpine
Successfully imported data from alpine
Importing data from archlinux
Successfully imported data from archlinux
Importing data from debian
Successfully imported data from debian
Importing data from safetydb
Traceback (most recent call last):
  File "/nix/store/2ailgd4vdaslsybza4vv7g5phdm48v1x-python3.8-schema-0.7.1/lib/python3.8/site-packages/schema.py", line 393, in validate
    nvalue = Schema(svalue, error=e, ignore_extra_keys=i).validate(value)
  File "/nix/store/2ailgd4vdaslsybza4vv7g5phdm48v1x-python3.8-schema-0.7.1/lib/python3.8/site-packages/schema.py", line 354, in validate
    return type(data)(o.validate(d) for d in data)
  File "/nix/store/2ailgd4vdaslsybza4vv7g5phdm48v1x-python3.8-schema-0.7.1/lib/python3.8/site-packages/schema.py", line 354, in <genexpr>
    return type(data)(o.validate(d) for d in data)
  File "/nix/store/2ailgd4vdaslsybza4vv7g5phdm48v1x-python3.8-schema-0.7.1/lib/python3.8/site-packages/schema.py", line 164, in validate
    raise SchemaError(
schema.SchemaError: Or({'advisory': <class 'str'>, 'cve': Or(None, Regex('CVE-\\d+-\\d+')), 'id': Regex('^pyup.io-\\d'), 'specs': <class 'list'>, 'v': <class 'str'>}) did not validate {'advisory': 'RoundCube in debops 2.0.0 uses the user login and password credentials to authenticate to the SMTP (submission) service before sending e-mail messages. This allows the SMTP server to check the message details, block mail with forged sender address, etc. The default configuration uses encrypted connections to the IMAP and SMTP services to ensure confidentiality and security.', 'cve': '""', 'id': 'pyup.io-26403', 'specs': ['<2.0.0'], 'v': '<2.0.0'}
Key 'cve' error:
Or(None, Regex('CVE-\\d+-\\d+')) did not validate '""'
[...]

PR #273, opened 2020-10-30
Fix and renable safetydb importer

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/244

### DIFF
--- a/vulnerabilities/importer_yielder.py
+++ b/vulnerabilities/importer_yielder.py
@@ -61,16 +61,16 @@ IMPORTER_REGISTRY = [
             'debian_tracker_url': 'https://security-tracker.debian.org/tracker/data/json'
         },
     },
-#     {
-#         'name': 'safetydb',
-#         'license': 'cc-by-nc-4.0',
-#         'last_run': None,
-#         'data_source': 'SafetyDbDataSource',
-#         'data_source_cfg': {
-#             'url': 'https://raw.githubusercontent.com/pyupio/safety-db/master/data/insecure_full.json',  # nopep8
-#             'etags': {}
-#         },
-#     },
+    {
+        'name': 'safetydb',
+        'license': 'cc-by-nc-4.0',
+        'last_run': None,
+        'data_source': 'SafetyDbDataSource',
+        'data_source_cfg': {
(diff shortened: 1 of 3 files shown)


vulnerabilities/importer_yielder.py
```diff
@@ -61,16 +61,16 @@
...
-#         'name': 'safetydb',
-#         'license': 'cc-by-nc-4.0',
-#         'last_run': None,
-#         'data_source': 'SafetyDbDataSource',
-#         'data_source_cfg': {
-#             'url': 'https://raw.githubusercontent.com/pyupio/safety-db/master/data/insecure_full.json',  # nopep8
-#             'etags': {}
-#         },
-#     },
+    {
```
pombredanne, line 64: We need to have a way to selectively enable some non-free resources like this one. As long as this is under a cc-by-nc  license we cannot integrate these data by default IMHO 
@jayfk ping?

pombredanne (changes requested):
This looks fine but we cannot have this enabled by default license-wise for now. How could we have a way to enable this on demand?

sbs2001:
@pombredanne  from the discussion we had, you suggested we handle this  by using a config file which contains a list of licenses. 
If an importer's license is not present in this list either run/don't run the importer. 

I think this logic belongs in the `importer_yielder.py` file, no need to create another config file.

vulnerabilities/importer_yielder.py
```diff
@@ -61,16 +61,16 @@
...
+        'name': 'safetydb',
+        'license': 'cc-by-nc-4.0',
+        'last_run': None,
+        'data_source': 'SafetyDbDataSource',
+        'data_source_cfg': {
+            'url': 'https://raw.githubusercontent.com/pyupio/safety-db/master/data/insecure_full.json',  # nopep8
+            'etags': {}
+        },
+    },
```
pombredanne, line 73: Let's add instead a flag to disable this rather than commenting it out and leave this disabled.

sbs2001:
Commits included in https://github.com/nexB/vulnerablecode/pull/259. Hence closing

PR #278, opened 2020-11-20
Cleanup codebase and fix minor bugs and other improvements

sbs2001:
* Move the implementation of the methods for handling etags and different
   file  formats into a separate `helpers.py` module to prevent code duplication

 * Fix bugs in the msr2019 importer.

### DIFF
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -21,9 +21,10 @@
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 import json
-import requests
+
 import yaml
-import pytoml as toml
+import requests
+import toml
 
 
 def load_yaml(path):
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -1,0 +1,54 @@
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import json
+import requests
+import yaml
+import pytoml as toml
+
+
+def load_yaml(path):
+    with open(path) as f:
+        return yaml.safe_load(f)
+
+
+def load_json(path):
+    with open(path) as f:
+        return json.load(f)
+
+
+def load_toml(path):
+    with open(path) as f:
+        return toml.load(f)
+
+
+def create_etag(data_src, url, etag_key):
+    etag = requests.head(url).headers.get(etag_key)
+    if not etag:
+        return True
+
+    elif url in data_src.config.etags:
+        if data_src.config.etags[url] == etag:
+            return False
+
+    data_src.config.etags[url] = etag
+    return True
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -21,6 +21,7 @@
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 import json
+import re
 
 import yaml
 import requests
@@ -61,3 +62,6 @@ def create_etag(data_src, url, etag_key):
 
     data_src.config.etags[url] = etag
     return True
+
+
+is_cve = re.compile(r"CVE-\d+-\d+", re.IGNORECASE).match
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -43,6 +43,14 @@ def load_toml(path):
 
 
 def create_etag(data_src, url, etag_key):
+    """
+    Etags are like hashes of web responses. For a data source `data_src`,
+    we maintain (url, etag) mappings in the DB.  `create_etag`  creates
+    (`url`, etag) pair. If a (`url`, etag) already exists then the code
+    skips processing the response further to avoid duplicate work.
+
+    `etag_key` is the name of header which contains the etag for the url.
+    """
     etag = requests.head(url).headers.get(etag_key)
     if not etag:
         return True
--- a/vulnerabilities/importers/alpine_linux.py
+++ b/vulnerabilities/importers/alpine_linux.py
@@ -20,6 +20,7 @@
 #  for any legal advice.
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+import logging
 from typing import Any
 from typing import Iterable
 from typing import List
--- a/vulnerabilities/importers/alpine_linux.py
+++ b/vulnerabilities/importers/alpine_linux.py
@@ -20,14 +20,13 @@
 #  for any legal advice.
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
-import logging
+
 from typing import Any
 from typing import Iterable
 from typing import List
 from typing import Mapping
 from typing import Set
 
-import yaml
 from packageurl import PackageURL
 from schema import Or
 from schema import Regex
@@ -36,6 +35,7 @@ from schema import Schema
 from vulnerabilities.data_source import Advisory
 from vulnerabilities.data_source import GitDataSource
 from vulnerabilities.data_source import Reference
+from vulnerabilities.helpers import load_yaml
 
 
 def validate_schema(advisory_dict):
@@ -91,19 +91,17 @@ class AlpineDataSource(GitDataSource):
     def _process_file(self, path) -> List[Advisory]:
         advisories = []
 
-        with open(path) as f:
-            record = yaml.safe_load(f)
+        record = load_yaml(path)
+        if record["packages"] is None:
+            return advisories
+        validate_schema(record)
 
-            if record["packages"] is None:
-                return advisories
-            validate_schema(record)
-
-            for p in record["packages"]:
-                advisories.extend(
-                    self._load_advisories(
-                        p["pkg"], record["distroversion"], record["reponame"], record["archs"],
-                    )
+        for p in record["packages"]:
+            advisories.extend(
+                self._load_advisories(
+                    p["pkg"], record["distroversion"], record["reponame"], record["archs"],
                 )
+            )
 
         return advisories
 
--- a/vulnerabilities/importers/alpine_linux.py
+++ b/vulnerabilities/importers/alpine_linux.py
@@ -20,7 +20,6 @@
 #  for any legal advice.
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
-
 from typing import Any
 from typing import Iterable
 from typing import List
--- a/vulnerabilities/importers/debian_oval.py
+++ b/vulnerabilities/importers/debian_oval.py
@@ -57,11 +57,9 @@ class DebianOvalDataSource(OvalDataSource):
         self.pkg_manager_api = DebianVersionAPI()
 
     def _fetch(self):
-        base_url = "https://www.debian.org/security/oval/"
-        file_name = "oval-definitions-{}.xml"
         releases = self.config.releases
         for release in releases:
-            file_url = base_url + file_name.format(release)
+            file_url = f"https://www.debian.org/security/oval/oval-definitions-{release}.xml"
             if not create_etag(data_src=self, url=file_url, etag_key="ETag"):
                 continue
 
--- a/vulnerabilities/importers/debian_oval.py
+++ b/vulnerabilities/importers/debian_oval.py
@@ -35,6 +35,7 @@ import requests
 
 from vulnerabilities.data_source import OvalDataSource, DataSourceConfiguration
 from vulnerabilities.package_managers import DebianVersionAPI
+from vulnerabilities.helpers import create_etag
 
 
 @dataclasses.dataclass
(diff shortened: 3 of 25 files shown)


vulnerabilities/helpers.py
```diff
@@ -0,0 +1,54 @@
...
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import json
+import requests
```
pombredanne, line 24: Do you mind to separate stdlib imports and sort each block?

vulnerabilities/management/commands/import.py
```diff
@@ -102,6 +102,13 @@ def _import_data(self, importers, cutoff_date):
...
-            self.stdout.write(
-                self.style.SUCCESS(f'Successfully imported data from {importer.name}'))
+            try:
+                ImportRunner(importer, batch_size).run(cutoff_date=cutoff_date)
+                self.stdout.write(
+                    self.style.SUCCESS(f'Successfully imported data from {importer.name}'))
+
+            except Exception as e:
+                # TBD
+                self.stdout.write(
```
pombredanne, line 112: IMHO just writing to stdout is not good enough as the likeliness we will ever look at logs is low.
This should be written to the DB instead.

pombredanne (changes requested):
Thanks! Seem my comments inline

vulnerabilities/management/commands/import.py
sbs2001: I am reverting this. This seems out of scope for this PR.

vulnerabilities/helpers.py
sbs2001: Done

vulnerabilities/helpers.py
```diff
@@ -0,0 +1,55 @@
...
+
+import json
+
+import yaml
+import requests
+import toml
+
+
+def load_yaml(path):
+    with open(path) as f:
```
pombredanne, line 31: Just as a FYI, https://pypi.org/project/saneyaml/ is s small library that I pushed to handle safer and saner YAML

vulnerabilities/importers/alpine_linux.py
```diff
@@ -20,6 +20,7 @@
 #  for any legal advice.
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+import logging
```
pombredanne, line 23: This seems no longer used
```suggestion
```

pombredanne (changes requested):
There is a minor nit left for your consideration

vulnerabilities/helpers.py
sbs2001: I thought `yaml.safe_load` solved the problem which `saneyaml` was fixing, hence the switch to plain yaml.

vulnerabilities/importers/alpine_linux.py
sbs2001: The irony :)  
Good catch thanks !

vulnerabilities/helpers.py
```diff
@@ -0,0 +1,55 @@
...
+        return json.load(f)
+
+
+def load_toml(path):
+    with open(path) as f:
+        return toml.load(f)
+
+
+def create_etag(data_src, url, etag_key):
+    etag = requests.head(url).headers.get(etag_key)
```
pombredanne, line 46: This is not trivial and a docstring would be welcomed

vulnerabilities/importers/alpine_linux.py
```diff
@@ -90,19 +91,17 @@ def updated_advisories(self) -> Set[Advisory]:
...
 
-            for p in record["packages"]:
-                advisories.extend(
-                    self._load_advisories(
-                        p["pkg"], record["distroversion"], record["reponame"], record["archs"],
-                    )
+        for p in record["packages"]:
+            advisories.extend(
+                self._load_advisories(
+                    p["pkg"], record["distroversion"], record["reponame"], record["archs"],
```
pombredanne, line 102: Using kwargs would be more explicit than plain positional args and would show immediately about how you are mapping the external data to you model. Otherwise, this is harder to read and grok.

vulnerabilities/importers/debian_oval.py
```diff
@@ -52,35 +53,24 @@ def __init__(self, *args, **kwargs):
...
         self.pkg_manager_api = DebianVersionAPI()
 
     def _fetch(self):
-        base_url = 'https://www.debian.org/security/oval/'
-        file_name = 'oval-definitions-{}.xml'
+        base_url = "https://www.debian.org/security/oval/"
+        file_name = "oval-definitions-{}.xml"
         releases = self.config.releases
         for release in releases:
             file_url = base_url + file_name.format(release)
```
pombredanne, line 64: A single format or f-string would be better IMHO

vulnerabilities/importers/debian_oval.py
```diff
@@ -52,35 +53,24 @@ def __init__(self, *args, **kwargs):
         # we could avoid setting translations, and have it
         # set by default in the OvalParser, but we don't yet know
         # whether all OVAL providers use the same format
-        self.translations = {'less than': '<'}
+        self.translations = {"less than": "<"}
         self.pkg_manager_api = DebianVersionAPI()
 
     def _fetch(self):
```
pombredanne, line 59: Why marking it with _? This is a non trivial function that would need testing (and it does much more than just fetching as it is parsing  and returning parts of a PURL

vulnerabilities/importers/npm.py
```diff
@@ -121,7 +121,9 @@ def _versions_to_purls(package_name, versions):
 
 
 def categorize_versions(
-    all_versions: Set[str], aff_version_range: str, fixed_version_range: str,
+    all_versions: Set[str],
+    aff_version_range: str,
```
pombredanne, line 125: aff_version_range is not super obvious as a name

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -22,24 +22,24 @@
...
 import csv
 import dataclasses
+import re
 import urllib.request
 
 # Reading CSV file from  a url using `requests` is bit too complicated.
-# Use `urllib.request` for that purpose. Need `requests` because making
-# a HEADER request using `urllib.request` is too complicated.
-import requests
+# Use `urllib.request` for that purpose.
```
pombredanne, line 29: Make this comment block a doctring instead

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -80,15 +68,23 @@ def fetch(self):
...
     def to_advisories(csv_reader):
         # Project KB MSR csv file has no header row
-        advsiories = []
+        advisories = []
         for row in csv_reader:
             vuln_id, proj_home, fix_commit, _ = row
             commit_link = proj_home + "/commit/" + fix_commit
-            advsiories.append(
+
+            if re.compile(r"CVE-\d+-\d+").match(vuln_id):
```
pombredanne, line 76: Using this in a loop is likely a bad idea performance and clarity-wise
I would create this instead, likely outside and in a utilities module as this is likely useful in other places:
```
is_cve = re.compile(r"CVE-\d+-\d+", re.IGNORE_CASE).match
```
then use with:
```
if is_cve(vuln_id):
....
```

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -80,15 +68,23 @@ def fetch(self):
...
+                reference = Reference(url=commit_link)
+
+            else:
+                reference = Reference(url=commit_link, reference_id=vuln_id)
+                vuln_id = ""
```
pombredanne, line 81: Is this needed?

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -80,15 +68,23 @@ def fetch(self):
...
+
+            advisories.append(
                 Advisory(
                     summary="",
```
pombredanne, line 85: Is this needed?

vulnerabilities/importers/project_kb_msr2019.py
```diff
@@ -80,15 +68,23 @@ def fetch(self):
...
                     impacted_package_urls=[],
```
pombredanne, line 86: Is this needed?

vulnerabilities/importers/safety_db.py
```diff
@@ -137,21 +138,11 @@ def updated_advisories(self) -> Set[Advisory]:
...
-                return False
-        self.config.etags[url] = etag
-        return True
-
 
 def categorize_versions(
-    package_name: str, all_versions: Set[str], version_specs: Iterable[str],
+    package_name: str,
+    all_versions: Set[str],
+    version_specs: Iterable[str],
```
pombredanne, line 145: consistency nit: Do we call these specs or ranges?

vulnerabilities/importers/ubuntu.py
```diff
@@ -53,22 +54,22 @@ def __init__(self, *args, **kwargs):
         # we could avoid setting translations, and have it
         # set by default in the OvalParser, but we don't yet know
         # whether all OVAL providers use the same format
-        self.translations = {'less than': '<'}
+        self.translations = {"less than": "<"}
```
pombredanne, line 57: This is duplicated with https://github.com/nexB/vulnerablecode/pull/278/files#diff-616b86ac339cf22cb94532d0b2c631ff58d97935da99e6b9ac8fd95569ab2d50R56

vulnerabilities/importers/safety_db.py
```diff
@@ -92,7 +93,7 @@ def set_api(self, packages):
         asyncio.run(self._versions.load_api(packages))
 
     def _fetch(self) -> Mapping[str, Any]:
-        if self.create_etag(self.config.url):
+        if create_etag(data_src=self, url=self.config.url, etag_key="ETag"):
```
pombredanne, line 96: I am under the impression that fetching and dealing with etags is scattered in many places. Is it?

vulnerabilities/importers/ubuntu.py
```diff
@@ -53,22 +54,22 @@ def __init__(self, *args, **kwargs):
...
         self.pkg_manager_api = LaunchpadVersionAPI()
 
     def _fetch(self):
-        base_url = 'https://people.canonical.com/~ubuntu-security/oval/'
-        file_name = 'com.ubuntu.{}.cve.oval.xml.bz2'
+        base_url = "https://people.canonical.com/~ubuntu-security/oval/"
```
pombredanne, line 61: This seems to be duplicated with https://github.com/nexB/vulnerablecode/pull/278/files#diff-616b86ac339cf22cb94532d0b2c631ff58d97935da99e6b9ac8fd95569ab2d50R59

vulnerabilities/package_managers.py
```diff
@@ -29,6 +29,7 @@
 
 from aiohttp import ClientSession
 from aiohttp.client_exceptions import ClientResponseError
+from aiohttp.client_exceptions import ServerDisconnectedError
```
pombredanne, line 32: Using aiohttp or request or urllib in various places to just fetch does not make sense to me. We should have only one way (one or a few functions) to fetch across the codebase

pombredanne (commented):
Thanks!
See my comments inline FYI.

vulnerabilities/package_managers.py
sbs2001: `aiohttp` is used only for making tons of requests to package managers to fetch  versions. 

Other than that I have tried to use `requests` wherever possible.

vulnerabilities/helpers.py
sbs2001: That makes sense.

vulnerabilities/importers/alpine_linux.py
sbs2001: Didn't saw this, I only wanted to refactor the file loading logic. 

I agree with you, this is ugly code. The name `_load_advisories` doesn't make sense either.

vulnerabilities/importers/debian_oval.py
sbs2001: I've named methods with underscore prefix when they are intended to only be used internally(within only that file)  . 
Atleast that's what I interpret from https://pep8.org/#descriptive-naming-styles

> _single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose name starts with an underscore.

re. This is a non trivial function : It's basically downloading new files by leveraging the well tested `create_etag` function. That returning of parts of PURL is make the OvalDataSouce understand how to create PackageURL objects.   I'd call it borderline trivial :) . 

Anyways functions starting with `_`  doesn't mean they don't have tests or trivial, they are just too specialised to be used elsewhere.

Do you want to switch to plain function names ?

vulnerabilities/importers/debian_oval.py
sbs2001: Right

vulnerabilities/importers/npm.py
sbs2001: Not my code, but  I see no problem in the naming.  What could be a better name ? `affected_version_range` ?

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: This is awesome idea !

vulnerabilities/importers/safety_db.py
sbs2001: This PR fixes that

vulnerabilities/importers/ubuntu.py
sbs2001: What's duplicated ? The link is pointing to another unrelated comment.

vulnerabilities/importers/safety_db.py
sbs2001: I'm renaming that to version_ranges. The feeds call it specs, which is probably why we called it specs here.

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: YES . Either the data makes vuln_id some cve or some apache collections id. In the latter case we instead of assigning apache collections id as a vulnerability id we throw it as a reference.

vulnerabilities/package_managers.py
sbs2001: FYI `urllib` is used in msr importer because it is more convenient for downloading and reading csv from memory than `requests` .

vulnerabilities/importers/project_kb_msr2019.py
sbs2001: Done

pombredanne (changes requested):
Here is my feedback.

pombredanne (approved):
LGTM after deeper review. You can merge... there are a few follow up tickets needed for code cleanup and naming updates.

vulnerabilities/importers/alpine_linux.py
sbs2001: yes

vulnerabilities/importers/npm.py
sbs2001: Done

vulnerabilities/importers/openssl.py
```diff
@@ -49,27 +50,13 @@ def updated_advisories(self) -> Set[Advisory]:
         # (url, etag) mappings in the DB. `create_etag`  creates
         # (url, etag) pair. If a (url, etag) already exists then the code
         # skips processing the response further to avoid duplicate work
-        if self.create_etag(self.url):
+        if create_etag(data_src=self, url=self.url, etag_key="ETag"):
```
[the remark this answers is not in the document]
sbs2001: That makes sense. I want to follow the semantics of Django ORM here, does `get_or_create_etag` make more sense ?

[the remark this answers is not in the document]
sbs2001: I just realized  what I just said was stupid :) . `create_etag` makes sense since the function returns True only if the etag is created otherwise that's gonna return Falsie.  It only returns the info required to deduce whether the encountered etag is new or not.

Issue #280, opened 2020-11-20
Adapt rust importer to new data format adopted by rust advisory db


sbs2001:
See https://github.com/RustSec/advisory-db/pull/420 . Our implementation is for old format. We need to adapt to new MD + TOML hybrid data format.

sbs2001:
Closed via https://github.com/nexB/vulnerablecode/pull/281

PR #281, opened 2020-11-21
Adapt rust importer to new advisory format

sbs2001:
Rust advisory db has moved on from using plain toml to using markdown + toml 'front matter' . 
This PR enables the current rust importer to import data from the new format.


sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/280

vulnerabilities/importers/rust.py
```diff
@@ -175,6 +173,22 @@ def categorize_versions(
...
 
-def load_toml(path):
-    with open(path) as f:
-        return toml.load(f)
+def load_toml_from_md(md_path):
+    # This collects the text between all the pairs of ```toml\n and ```\n characters
+    # and converts it into a parsed toml document and returns it.
+    parsed_data = {}
+    with open(md_path) as f:
+        lines = f.readlines()
```
pombredanne, line 181: I wonder if this could be made better.

vulnerabilities/importers/rust.py
```diff
@@ -175,6 +173,22 @@ def categorize_versions(
...
+                j += 1
+
+                while lines[j] != "```\n":
+                    toml_lines.append(lines[j])
+                    j += 1
+
+                parsed_toml = toml.loads("".join(toml_lines))
+                parsed_data.update(parsed_toml)
+
+    return parsed_data
```
pombredanne, line 194: Here is a more ambitious replacement for this function broken in units that are easier to tests and that also contains passing doctests meaning that you can run `pytest <path to this file>` and this will execute these tests. It also provide a documented example and specifies the key behavior of the functions: 

```Python
import toml


def get_advisory_data(location):
    """
    Return a mapping of vulnerability data from a RustSec advisory file at
    ``location``.
    RustSec advisories documents are .md files starting with a block of TOML
    identified as the text inside a tripple-backtick TOML block. Per
    https://github.com/RustSec/advisory-db#advisory-format:
        Advisories are formatted in Markdown with TOML "front matter".
    """

    with open(location) as lines:
        toml_lines = get_toml_lines(lines)
        return data_from_toml_lines(toml_lines)


def get_toml_lines(lines):
    """
    Yield lines of TOML extracted from an iterable of text ``lines``.
    The lines are expected to be from a RustSec Markdown advisory file with
    embedded TOML metadata.

    For example::

    >>> text = '''
    ... ```toml
    ... [advisory]
    ... id = "RUST-001"
    ...
    ... [versions]
    ... patch = [">= 1.2.1"]
    ... ```
    ... # Use-after-free with objects returned by `Stream`'s `get_format_info`
    ...
    ... Affected versions contained a pair of use-after-free issues with the objects.
    ... '''
    >>> list(get_toml_lines(text.splitlines()))
    ['', '[advisory]', 'id = "RUST-001"', '', '[versions]', 'patch = [">= 1.2.1"]']
    """

    for line in lines:
        line = line.strip()
        if line.startswith("```toml"):
            continue
        elif line.endswith("```"):
            break
        else:
            yield line


def data_from_toml_lines(lines):
    """
    Return a mapping of data from an iterable of TOML text ``lines``.

    For example::

    >>> lines = ['[advisory]', 'id = "RUST1"', '', '[versions]', 'patch = [">= 1"]']
    >>> data_from_toml_lines(lines)
    {'advisory': {'id': 'RUST1'}, 'versions': {'patch': ['>= 1']}}
    """
    return toml.loads("\n".join(lines))
```

pombredanne (changes requested):
Thanks!
See some comments inline. Also try to avoid the big test files. Testting the opening of a zip file is rather trivial. One the other hand having a few test markdown as text  files would make sense.
 
>  +515 KB vulnerabilities/tests/test_data/advisory-db.zip
>  +365 KB (170%) vulnerabilities/tests/test_data/rust-advisory-db.zip

vulnerabilities/importers/rust.py
sbs2001: You are right, it is better to break this method into smaller chunks.

Having said that thesuggested `get_toml_lines` oversimplifies things. One eg where it won't work is if there are lines before the the front matter (not sure these exist) or if there are multiple toml front matters in the same file eg https://github.com/RustSec/advisory-db/blob/master/rust/cargo/CVE-2019-16760.md . This is trivial to fix and probably not the core of the suggestion :) 

Also I don't think `data_from_toml_lines`  is a good idea.  It is essentially wrapping a not so complicated one liner.  What do you think ?

sbs2001:
@pombredanne  about zip files : 

Those are compressed git repositories. I tried trimming these but I was getting mysterious bugs and just gave up on that. We have `rust-advisory-db.zip` and `advisory-db.zip` instead of a single  `rust-advisory-db.zip` because the tests in `test_data_source.py` require the old kind of rust-advisory-db.

vulnerabilities/importers/rust.py
pombredanne: > One eg where it won't work is if there are lines before the the front matter (not sure these exist) or if there are multiple toml front matters in the same file eg https://github.com/RustSec/advisory-db/blob/master/rust/cargo/CVE-2019-16760.md . 

https://github.com/RustSec/advisory-db/blob/master/rust/cargo/CVE-2019-16760.md  has only one front matter AFAICS. The spec is clear that there is one front matter and not more than one. 

@tarcieri can you confirm that the new markdown/toml Rust advisory format has only ONE front matter?

It would not be named a front matter if there were more than one IMHO. There could be other toml code blocks alright BUT these are not part of the metadata for the advisory.
If you bundle all the toml block in one, you will create weird and unrelated data from what I can see.


> Also I don't think data_from_toml_lines is a good idea. It is essentially wrapping a not so complicated one liner. What do you think ?


Actually this helps separate the lines extraction and the conversion to data. This in turns help to test each separately AND without having to depend on files and file-like objects, which helped me having a set of doctests. In fact this function was created initially and solely to help with testing (and helping with testing is always a good thing IMHO)

pombredanne:
re:

> Those are compressed git repositories. I tried trimming these but I was getting mysterious bugs and just gave up on that. 


IMHO we should be able to create smaller zip with fewer files. How did you do this? 


> We have rust-advisory-db.zip and advisory-db.zip instead of a single rust-advisory-db.zip because the tests in test_data_source.py require the old kind of rust-advisory-db.

Then may be renaming these to have a clearer name would help.

vulnerabilities/importers/rust.py
tarcieri: There is only one front matter section, correct, and it *must* be at the beginning of the file.

sbs2001: @pombredanne  I've made committed the suggested changes.

vulnerabilities/importers/rust.py
```diff
@@ -23,13 +23,14 @@
...
 from itertools import chain
+import re
 from typing import Optional, Mapping
 from typing import Set
 from typing import Tuple
 from urllib.error import HTTPError
 from urllib.request import urlopen
 
-import pytoml as toml
+import toml
```
pombredanne, line 33: nit: sort the imports

vulnerabilities/importers/rust.py
```diff
@@ -66,15 +67,13 @@ def updated_advisories(self) -> Set[Advisory]:
         return self._load_advisories(self._updated_files)
 
     def _load_advisories(self, files) -> Set[Advisory]:
-        files = [f for f in files if not f.endswith("-0000.toml")]  # skip temporary files
+        files = [f for f in files if not f.endswith("-0000.md")]  # skip temporary files
```
pombredanne, line 70: are you sure it will always end with "0000.md" ?

tarcieri: It will always be named `RUSTSEC-0000-0000.md`

pombredanne:
```suggestion
                # per @tarcieri It will always be named RUSTSEC-0000-0000.md
                # https://github.com/nexB/vulnerablecode/pull/281/files#r528899864
                files = [f for f in files if not f.endswith("-0000.md")]  # skip temporary files
```

pombredanne (approved):
I made a small suggestion for your consideration. LGTM otherwise!
Thank you

Issue #165, opened 2020-03-17
Debian importer is  treating lots of non-cves like cves


sbs2001:
```
>>> z = pickle.load(open('debi.pickle','rb'))
>>> for i in z :
...     if not i['cve_id'].startswith('CVE') :
...             print(i['cve_id'])
... 
TEMP-0105562-0FE13B
TEMP-0582798-329FE7
TEMP-0000000-DEED53
TEMP-0000000-75B37A
TEMP-0000000-106DD8
TEMP-0774898-681A65
TEMP-0532740-DB1B64
TEMP-0000000-481246
TEMP-0579087-7F12A8
TEMP-0841257-B7CD60
........................
```
These are further dumped directly https://github.com/nexB/vulnerablecode/blob/c5c69a46f2476056158cfb87bdc304baabba5eba/vulnerabilities/data_dump.py#L38 . So basically we are having non-cve ids  in cve fields.

The `debi.pickle` was obtained by pickling https://github.com/nexB/vulnerablecode/blob/c5c69a46f2476056158cfb87bdc304baabba5eba/vulnerabilities/scraper/debian.py#L76

haikoschol:
The bug is in `extract_vulnerabilities()` assuming that the object key in the JSON response always contains a CVE ID. Apparently Debian publishes advisories with temporary IDs when no CVE ID is available yet. The obvious next question then is how do those temporary IDs turn in to CVE IDs and how would we detect that and update our database accordingly?

sbs2001:
https://deb.freexian.com/extended-lts/tracker/data/fake-names apparently these are automatically generated. 

> These names are not stable and can change when the database is updated, so they should not be used in external references. 

Should we just ignore these vulnerabilities for now ?

haikoschol:
Either we ignore them or we don't store the temp ID and hope the summary won't change so we can find the entry again when it needs to be updated with a CVE ID.

sbs2001:
I think ignoring would be better instead of later finding them and adding CVE ID, because if in case it doesn't get CVE ID we will end up with garbage data.

sbs2001:
Closed in favour of https://github.com/nexB/vulnerablecode/issues/300

PR #285, opened 2020-11-25
Stop debian importer from collecting temp vulnerabilities 

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/165

### DIFF
--- a/vulnerabilities/importers/debian.py
+++ b/vulnerabilities/importers/debian.py
@@ -111,6 +111,10 @@ class DebianDataSource(DataSource):
             if not cve_id.startswith("CVE"):
                 continue
 
+            # vulnerabilities starting with something may not be public yet
+            # see for instance https://web.archive.org/web/20201215213725/https://security-tracker.debian.org/tracker/TEMP-0000000-A2EB44  # nopep8
+            # TODO: this would need to be revisited though to ensure we are not missing out on anything  # nopep8
+
             for release_name, release_record in record["releases"].items():
                 if not release_record.get("repositories", {}).get(release_name):
                     continue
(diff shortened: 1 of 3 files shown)


pombredanne (changes requested):
I think it makes sense, but I would prefer to have a function flag that's called `ignore_temporary_vulnerabilities=True` added so what we are doing is very explicit

vulnerabilities/importers/debian.py
```diff
@@ -108,6 +108,8 @@ def _parse(self, pkg_name: str, records: Mapping[str, Any]) -> List[Advisory]:
 
         for cve_id, record in records.items():
             impacted_purls, resolved_purls = set(), set()
+            if not cve_id.startswith("CVE"):
```
pombredanne, line 111:
```suggestion
            if not cve_id.startswith("CVE"):
                 # vulnerabilities starting with something may not be public yet
                 # see for instance https://web.archive.org/web/20201215213725/https://security-tracker.debian.org/tracker/TEMP-0000000-A2EB44
                 # TODO: this would need to be revisited though to ensure we are not missing out on anything
```

sbs2001: Shouldn't that be `with something else may not be public` ?

sbs2001:
@pombredanne  re: ignore_temporary_vulnerabilities=True

We add a config like that at https://github.com/nexB/vulnerablecode/blob/5e551675152b5dea32a50b6f2348a1fbff7c58b6/vulnerabilities/importer_yielder.py#L61 .

pombredanne:
> ignore_temporary_vulnerabilities=True

The thing with a config flag like that is that nobody will know what to do with it. You could  add this but a TODO code comment  is simpler and add a ticket so we can revisit the issue latest

sbs2001:
@pombredanne  re: a ticket so we can revisit the issue latest

Yeah this makes sense. In fact RustSecDb too has temp advisories so this is not limited to debian

sbs2001:
I have added a comment to revisit this. I'm merging this for now .

PR #290, opened 2020-11-29
Store severity scores 

### LINKED ISSUE #157: Change database schema to accommodate different types of severity indicators

sbs2001:
Implementation of https://github.com/nexB/vulnerablecode/issues/157#issuecomment-734830969

Fixes #157 

- [x]  Add data structures to store severity scores


vulnerabilities/tests/test_nvd.py
```diff
@@ -139,9 +140,18 @@ def test_to_advisories(self):
                     [
                         Reference(
                             url="http://code.google.com/p/gperftools/source/browse/tags/perftools-0.4/ChangeLog",  # nopep8
+                            scores=[],
```
pombredanne, line 143: Is there any need to initialize this at all?
Same for below https://github.com/nexB/vulnerablecode/pull/290/files#diff-5bd858dbcd0bcd0a70d34d3b2c47bcc08ab731569df61e265711ad8c8bedcb6cR147

vulnerabilities/tests/test_nvd.py
```diff
@@ -139,9 +140,18 @@ def test_to_advisories(self):
...
                         ),
                         Reference(
                             url="http://kqueue.org/blog/2012/03/05/memory-allocator-security-revisited/",  # nopep8
+                            scores=[],
+                        ),
+                        Reference(
+                            url="https://nvd.nist.gov/vuln/detail/CVE-2005-4895",  # nopep8
+                            scores=[
```
pombredanne, line 151: Just a note on the naming consistency:
why would a list of `VulnerabilitySeverity` be called `scores`?
Either `VulnerabilitySeverity` should be renamed to `Score` which may be best.
Or `scores` should be renamed to `severities`

vulnerabilities/tests/test_nvd.py
```diff
@@ -154,6 +164,10 @@ def test_to_advisories(self):
         found_advisories = list(self.data_src.to_advisories(self.nvd_data))
         # Only 1 advisory because other advisory is hardware related
         assert len(found_advisories) == 1
-        found_advisories[0].vuln_references = sorted(found_advisories[0].vuln_references, key=lambda x: x.url)  # nopep8
+        found_advisories[0].vuln_references = sorted(
+            found_advisories[0].vuln_references, key=lambda x: x.url
+        )  # nopep8
+        # for id,ref in found_advisories[0].vuln_references :
```
pombredanne, line 170: This commented code may be best deleted?

pombredanne (changes requested):
Thanks!
See my comments in line for your consideration.

vulnerabilities/tests/test_nvd.py
sbs2001: That makes sense. 

I am going with  `scores-> severities`  since .

vulnerabilities/tests/test_nvd.py
sbs2001: Nope

vulnerabilities/importers/redhat.py
```diff
@@ -40,9 +43,9 @@ def __enter__(self):
     def updated_advisories(self):
         processed_advisories = []
         for advisory_data in self.redhat_response:
-            processed_advisories.append(to_advisory(advisory_data))
+            yield [to_advisory(advisory_data)]
```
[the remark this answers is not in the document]
sbs2001: That was for convenience in testing db writes, reverting it.

vulnerabilities/data_source.py
```diff
@@ -46,11 +46,18 @@
...
+
+
 @dataclasses.dataclass
 class Reference:
 
-    url: str = ''
-    reference_id: str = ''
+    scores: List[VulnerabilitySeverity] = dataclasses.field(default_factory=list)
+    url: str = ""
+    reference_id: str = ""
```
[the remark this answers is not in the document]
sbs2001: Why ?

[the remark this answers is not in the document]
pombredanne: Usually one put identifiers first in order and then go from the most to the least significant attributes grouped logically. Putting an id last means that a standard representation of the object would have the score first which would be weird.
It also means that the objects sort order would be based on scores which does not make sense as this is not something major there

vulnerabilities/importers/redhat.py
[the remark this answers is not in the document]
pombredanne: My point was not about the yield... but the yield of a list.

vulnerabilities/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
pombredanne, line 1: This should be deleted and then added to the `.gitignore`

sbs2001: Oops. Done

vulnerabilities/tests/test_nvd.py
sbs2001: Deleted

vulnerabilities/tests/test_nvd.py
sbs2001: Done

vulnerabilities/tests/test_nvd.py
sbs2001: Fixed

vulnerabilities/models.py
```diff
@@ -162,7 +162,9 @@ class Importer(models.Model):
     name = models.CharField(max_length=100, unique=True, help_text="Name of the importer")
 
     license = models.CharField(
-        max_length=100, blank=True, help_text="License of the vulnerability data",
+        max_length=100,
```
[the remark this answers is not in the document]
sbs2001: you opened https://github.com/nexB/vulnerablecode/issues/293 .

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,50 @@ def to_advisory(advisory_data):
     references = []
     if advisory_data.get("bugzilla"):
         bugzilla = advisory_data.get("bugzilla")
+        url = "https://bugzilla.redhat.com/show_bug.cgi?id={}".format(bugzilla)
+        bugzilla_severity = VulnerabilitySeverity(
+            severity_type="REDHAT_BUGZILLA_SEVERITY",
```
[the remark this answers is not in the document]
sbs2001: My best effort is RH_BUGZILLA_SEVERITY which I guess is not enough, please do the naming

vulnerabilities/importers/redhat.py
```diff
@@ -40,9 +43,9 @@ def __enter__(self):
     def updated_advisories(self):
         processed_advisories = []
```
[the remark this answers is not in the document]
sbs2001: Fixed.

vulnerabilities/data_source.py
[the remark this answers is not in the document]
sbs2001: Ok, fixed .

vulnerabilities/importers/redhat.py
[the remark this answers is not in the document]
sbs2001: I see,  FWIW  the definition of `batch_advisories`  is . It yields sets of advisories. The importer_runner  expects the `updated_advisories` to yield a BATCH of advisories.

```
    def batch_advisories(self, advisories: List[Advisory]) -> Set[Advisory]:
        """
        Yield batches of the passed in list of advisories.
        """
        advisories = advisories[:]  # copy the list as we are mutating it in the loop below

        while advisories:
            b, advisories = advisories[: self.batch_size], advisories[self.batch_size:]
            yield set(b)
```

I don't see any significant memory-use reduction in this case since the data is not too much.

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,50 @@ def to_advisory(advisory_data):
...
+
+        url = "https://access.redhat.com/errata/{}".format(rh_adv)
+
+        # RH provides 3 types of advisories RHSA, RHBA, RHEA. Only RHSA's contain severity score.
+        # See https://access.redhat.com/articles/2130961 for more details.
+
+        if "RHSA" in rh_adv:
+            rhsa_aggregate_severity = VulnerabilitySeverity(
+                severity_type="RHSA_AGGREGATE_SEVERITY",
+                severity_value=requests.get(f"https://access.redhat.com/hydra/rest/securitydata/cvrf/{rh_adv}.json").json()["cvrfdoc"]["aggregate_severity"],  # nopep8
```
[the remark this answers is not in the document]
sbs2001: Fixed

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,50 @@ def to_advisory(advisory_data):
...
+                    reference_id=rh_adv,
+                )
+            )
+
+        else:
+            references.append(Reference(scores=[], url=url, reference_id=rh_adv))
+
+    redhat_cvss3 = VulnerabilitySeverity(
+        severity_type="REDHAT_CVSS3",
+        severity_value=requests.get(advisory_data["resource_url"]).json()["cvss3"]["cvss3_base_score"],  # nopep8
```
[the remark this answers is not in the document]
sbs2001: Fixed

vulnerabilities/importers/redhat.py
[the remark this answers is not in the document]
pombredanne: ok, let's keep it as is then, but batching is an optimization that would likely be best done later, as it may not help with clarity ;)

vulnerabilities/importers/nvd.py
```diff
@@ -89,6 +98,28 @@ def extract_summary(cve_item):
         summaries = [desc["value"] for desc in cve_item["cve"]["description"]["description_data"]]
         return max(summaries, key=len)
 
+    @staticmethod
+    def extract_severity_scores(cve_item):
+        severity_scores = []
+
+        if cve_item["impact"].get("baseMetricV3"):
```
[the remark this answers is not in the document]
sbs2001: You said to keep it simple, hence I didn't bother about vectors. 

re robustness . 

It always contains ["impact"] see https://csrc.nist.gov/schema/nvd/feed/1.1-Beta/nvd_cve_feed_json_1.1_beta.schema 


Also https://nvd.nist.gov/vuln/data-feeds says

>JSON Vulnerability Feeds  – Each vulnerability in the file includes a description and associated reference links from the CVE® dictionary feed, as well as CVSS base scores, vulnerable product configuration, and weakness categorization.

vulnerabilities/importers/redhat.py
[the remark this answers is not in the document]
pombredanne: What are you trying to convey there? I think that this is information about how the severity score value was "formed" or "computed". `CVSSv2` and `CVVSv3` are good examples, but then they are just codes.
So either:
1. we have codes that could eventually be related to more details likely in a table with some texts, urls, etc. 
2. OR this type is self-standing and should be as explicit as possible in a short string. REDHAT_BUGZILLA_SEVERITY and RH_BUGZILLA_SEVERITY does not tell me much. I would need to do quite a bit searching to find what you meant.
The best I can think of if we have a single string for the type would be to use a URL such as 
- https://bugzilla.redhat.com/page.cgi?id=fields.html#bug_severity
- https://www.first.org/cvss/v3-0/
- https://www.first.org/cvss/v2/

IMHO the 1st solution may be better but that's weak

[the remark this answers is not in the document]
sbs2001: What's the difference in (1) and (2) if you are using urls either directly or indirectly (via code-> mapping) ? 

How would (1) solve the problem which you had with (2) ? 
>REDHAT_BUGZILLA_SEVERITY and RH_BUGZILLA_SEVERITY does not tell me much. I would need to do quite a bit searching to find what you meant.

It would only make it worse, since now there's 100% chance you need to lookup what the code means. 

I'm more inclined towards either using something like `REDHAT_BUGZILLA_SEVERITY` or the urls you mentioned.

vulnerabilities/data_source.py
```diff
@@ -46,11 +46,18 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class VulnerabilitySeverity:
+    severity_type: str
```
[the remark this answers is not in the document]
pombredanne: Or another possibility to consider:
```

@dataclasses.dataclass
class SeverityScheme:
    # a short name for that represents that scheme such as CVSSv2
    name: str
    # a url to documentation about that scheme
    url: str 
    # notes about that scheme 
    notes=""

    def as_score(self, value):
        """
        Return a normalized numeric score for this scheme given a raw value.
        For instance htis can be used to convert a CVSS vector to a base score.
        """
        raise NotImplementedError


severity_schemes = [
    SeverityScheme(
        identifier = 'cvssv2',
        name ='CVSSv2',
        url='https://www.first.org/cvss/v2/',
        notes='We store the vector as value and compute scores from that.'
    ),
    SeverityScheme(
        identifier = 'cvssv3',
        name ='CVSSv3',
        url='https://www.first.org/cvss/v3-0/',
        notes='We store the vector as value and compute scores from that.'
    ),
    SeverityScheme(
        identifier = 'rhbs',
        name ='RedHat bug severity',
        url='https://bugzilla.redhat.com/page.cgi?id=fields.html#bug_severity',
    ),
]


@dataclasses.dataclass
class Severity:
    # This represents the scheme used to compute the severity.
    # Typically a combination of scale and some scoring procedure
    scheme: SeverityScheme
    value: str

# note that in the Django models we might store the SeverityScheme.identifier
# and use the list of severity_schemes as a base for Choice in that field

```

vulnerabilities/models.py
```diff
@@ -198,3 +200,17 @@ def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> Dat
 
     def __str__(self):
         return self.name
+
+
+class VulnerabilitySeverity(models.Model):
+
+    severity_type = models.CharField(
```
pombredanne, line 207: See the suggested changes in https://github.com/nexB/vulnerablecode/pull/290/files#r540875185
If these are considered as OK, then this field may become `scoring_system_identifier` and below `severity_value` could become `value`?

Also I would suggest this order for fields and for the unique together:
- "vulnerability"
- "value"
- "scoring_system_identifier"
 -"reference"

Telling this story:
For a "vulnerability", I have a severity "value" computed using a scoring system identified by "scoring_system_identifier" and provided by this "reference"

pombredanne (commented):
See my suggested changes for the models.

vulnerabilities/data_source.py
sbs2001: @pombredanne  along the same lines, how about instead of `ScoringSystem` dataclass, we have a `ScoringSystem` django model, and have something to provide intial seed rows of  `ScoringSystems` . 

On the `Severity` dataclass side, we could turn `scoring_system: ScoringSystem` into something like `scoring_system: str` where the `scoring_system` would have value of the identifier.

sbs2001: The advantage of this would be all the `ScoringSystem`s would live in db together with the values, giving us nice mapping of `vuln->severity->severity_typ_details` . 

On the other hand when using `ScoringSystem` dataclass, all the ScoringSytems would live in the memory so our mapping would be `vuln->severity` , if I have to know more about the provided `severity` I would need to lookup  the python file containing  `scoring_systems`  .

pombredanne:
@sbs2001 you wrote:

> I thought about https://github.com/nexB/vulnerablecode/pull/290#discussion_r540875185 , and I think storing the ScoringSystems in the db would be a better approach 

It can but then how scoring systems will we have? At most a couple per data sources, and in reality only a handful. And this is purely static data, and not really usable as data for some query except for the identifier so I am not sure this needs the added ceremony and complexity when stored in the DB. (e.g. you then need fixtures to bootstrap etc.). I see  this instead as an enhanced but still very small list of value choices for a single field.

vulnerabilities/data_source.py
```diff
@@ -68,17 +75,18 @@ class Advisory:
...
     cve_id: Optional[str] = None
 
     def __hash__(self):
-        s = '{}{}{}{}'.format(
+        s = "{}{}{}{}".format(
             self.summary,
-            ''.join(sorted([str(p) for p in self.impacted_package_urls])),
-            ''.join(sorted([str(p) for p in self.resolved_package_urls])),
+            "".join(sorted([str(p) for p in self.impacted_package_urls])),
+            "".join(sorted([str(p) for p in self.resolved_package_urls])),
```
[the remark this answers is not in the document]
sbs2001: I'll do this in another PR, possibly removing this function

vulnerabilities/data_source.py
sbs2001: Using objects instead, after thinking more about this. 

Main reason being, having a document instead of objects does not enforce documenting different types of severity types :)

vulnerabilities/importers/redhat.py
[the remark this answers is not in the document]
sbs2001: This resolved, we chose to have a in-memory registry of scoring systems

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,53 @@ def to_advisory(advisory_data):
     references = []
     if advisory_data.get("bugzilla"):
         bugzilla = advisory_data.get("bugzilla")
+        url = "https://bugzilla.redhat.com/show_bug.cgi?id={}".format(bugzilla)
```
pombredanne, line 81: In the lines above, this does not make sense and is a risky repeat:
```
    if advisory_data.get("bugzilla"):
        bugzilla = advisory_data.get("bugzilla")
```
... It should be instead:
```
bugzilla = advisory_data.get("bugzilla")
    if bugzilla:
...  
```
No need to fix this just now, yet we will need to get rid of these kind of repetitions.

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,53 @@ def to_advisory(advisory_data):
...
 
-    for rhsa in advisory_data["advisories"]:
-        references.append(
-            Reference(
-                url="https://access.redhat.com/errata/{}".format(rhsa), reference_id=rhsa,
+    for rh_adv in advisory_data["advisories"]:
+        # RH provides 3 types of advisories RHSA, RHBA, RHEA. Only RHSA's contain severity score.
+        # See https://access.redhat.com/articles/2130961 for more details.
+
+        if "RHSA" in rh_adv:
```
pombredanne, line 101: Are you positive 100% this is ALWAYS uppercase?
A case-insensitive check is usually safer.

vulnerabilities/importers/redhat.py
```diff
@@ -75,21 +78,53 @@ def to_advisory(advisory_data):
...
+                )
+            )
+
+        else:
+            references.append(Reference(severities=[], url=url, reference_id=rh_adv))
+
+    redhat_cve_entry = requests.get(advisory_data["resource_url"]).json()
+    redhat_cvss = redhat_cve_entry["cvss3"]["cvss3_base_score"]
+    redhat_cvss3 = VulnerabilitySeverity(
+        system=scoring_systems["rh_cvssv3"],
```
pombredanne, line 123: Is `rh_cvssv3` really a thing? e.g. are they using rules that are different from the standard `cvssv3`? or rather they just provide a different score using the `cvssv3` rules?
IMHO this is the later, and the `advisory_data["resource_url"])` is enough to qualify where this specific valueof CVSSv3 comes from.

vulnerabilities/migrations/0002_vulnerabilityseverityscore.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.0.7 on 2020-11-27 13:43
```
pombredanne, line 1: Could some of these migrations be squashed in a fewer number of migrations, for clarity?
We are creating then deleting then creating something else... all these details do not make sense to keep

vulnerabilities/models.py
```diff
@@ -199,3 +200,25 @@ def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> Dat
 
     def __str__(self):
         return self.name
+
+
+class VulnerabilitySeverity(models.Model):
+
+    scoring_system_choices = [(system.identifier, system.name) for system in scoring_systems.values()]  # nopep8
```
pombredanne, line 207: Why nopep8?
What about this?
```suggestion
    scoring_system_choices = tuple((ss.identifier, ss.name) for ss in scoring_systems.values())
```
BTW, I am also suggesting to use a tuple for something not supposed to change at runtime

vulnerabilities/severity_systems.py
```diff
@@ -0,0 +1,54 @@
...
+    name: str
+    # a url to documentation about that sscoring system
+    url: str
+    # notes about that scoring system
+    notes: str = ""
+
+    def as_score(self, value):
+        """
+        Return a normalized numeric score for this scoring system  given a raw
+        value. For instance htis can be used to convert a CVSS vector to a base
```
pombredanne, line 20:
```suggestion
        value. For instance this can be used to convert a CVSS vector to a base
```

vulnerabilities/severity_systems.py
```diff
@@ -0,0 +1,54 @@
...
+        identifier="rhbs",
+        name="RedHat Bugzilla severity",
+        url="https://bugzilla.redhat.com/page.cgi?id=fields.html#bug_severity",
+    ),
+    "rhas": ScoringSystem(
+        identifier="rhas",
+        name="RedHat Aggregate severity",
+        url="https://access.redhat.com/security/updates/classification/",
+    ),
+    "rh_cvssv3": ScoringSystem(
```
pombredanne, line 49: See my comments above. This is computed using all the CVSSv3 rules AFAIK, the only thing is that RH provide their own score values... not a new system for this.

vulnerabilities/tests/test_nvd.py
```diff
@@ -143,6 +145,13 @@ def test_to_advisories(self):
                         Reference(
                             url="http://kqueue.org/blog/2012/03/05/memory-allocator-security-revisited/",  # nopep8
                         ),
+                        Reference(
+                            url="https://nvd.nist.gov/vuln/detail/CVE-2005-4895",  # nopep8
+                            severities=[
+                                VulnerabilitySeverity(system=scoring_systems["cvssv2"], value="5.0")
```
pombredanne, line 151: We need to be clear about what we store: the vector or the value? I thought we discussed storing the vector for now.
See https://github.com/nexB/vulnerablecode/pull/290/files#diff-d0d10e4267bd7b78581a8264137e41fad843af2fe6c813234a661e5a72930c2cR37
> "We store the vector as value and compute scores from that.",

and https://github.com/nexB/vulnerablecode/pull/290/files#diff-d0d10e4267bd7b78581a8264137e41fad843af2fe6c813234a661e5a72930c2cR17 to compute a score

vulnerabilities/tests/test_redhat_importer.py
```diff
@@ -83,18 +85,38 @@ def test_to_advisory(self):
...
+                                    system=scoring_systems["rhas"], value=2.2
+                                )
+                            ],
                         ),
                         Reference(
                             url="https://access.redhat.com/hydra/rest/securitydata/cve/CVE-2016-9401.json",  # nopep8
                             reference_id="",
+                            severities=[
+                                VulnerabilitySeverity(
+                                    system=scoring_systems["rh_cvssv3"], value=6.0
```
pombredanne, line 117: this is not a scoring system

vulnerabilities/tests/test_redhat_importer.py
```diff
@@ -138,26 +160,56 @@ def test_to_advisory(self):
...
+                                    system=scoring_systems["rhas"], value=2.2
+                                )
+                            ],
                         ),
                         Reference(
                             url="https://access.redhat.com/hydra/rest/securitydata/cve/CVE-2016-10200.json",  # nopep8
                             reference_id="",
+                            severities=[
+                                VulnerabilitySeverity(
+                                    system=scoring_systems["rh_cvssv3"], value=6.0
```
pombredanne, line 210: this is not a scoring system

pombredanne (changes requested):
Thanks!
See my comments inline

vulnerabilities/importers/redhat.py
sbs2001: >Is rh_cvssv3 really a thing?

yes it is ! https://access.redhat.com/solutions/762393

vulnerabilities/migrations/0002_vulnerabilityseverityscore.py
sbs2001: Yes, I  will squash this once everything else is fine.

vulnerabilities/models.py
sbs2001: > Why nopep8?

Black ignores nopep8, I wanted this on a single line . 

re Tuple . 

Yes

vulnerabilities/importers/redhat.py
sbs2001: Thanks for this.

vulnerabilities/importers/redhat.py
sbs2001: Done

vulnerabilities/import_runner.py
```diff
@@ -267,8 +240,8 @@ def _get_or_create_vulnerability(
 
     except Exception:
         logger.error(
-            f"Failed to _get_or_create_vulnerability: {query_kwargs!r}:\n"
-            + traceback.format_exc())
+            f"Failed to _get_or_create_vulnerability: {query_kwargs!r}:\n" + traceback.format_exc()
```
pombredanne, line 243: One (new) idiom to consider is:
```Python
    except Exception as e:
        msg = f"Failed to _get_or_create_vulnerability: {query_kwargs!r}"
        logger.error("\n".join([msg, traceback.format_exc()])
        raise Exception(msg) from e

```

vulnerabilities/importers/nvd.py
```diff
@@ -90,6 +100,28 @@ def extract_summary(cve_item):
...
+            severity_scores.append(
```
pombredanne, line 108: For each CVSS scores we really want to store:
1. a score proper
2. the vector string as another "scoring system" , possibly like `cvssv2_vector` and `cvssv3_vector` that are each their own extra `VulnerabilitySeverity`

vulnerabilities/tests/test_redhat_importer.py
pombredanne: A CVSSV3 VulnerabilitySeverity created by RedHat is already in the context of a RHSA Reference above. RedHat just uses the same principles as the NVD when the use a CVSSv3, e.g. the same system. They just provide different values.

vulnerabilities/severity_systems.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        raise NotImplementedError
+
+
+scoring_systems = {
+    "cvssv2": ScoringSystem(
+        identifier="cvssv2",
+        name="CVSSv2",
+        url="https://www.first.org/cvss/v2/",
+        notes="We store the vector as value and compute scores from that.",
```
pombredanne, line 31: Based on the new design suggestion here, we will store the vector and values each as different `ScoringSystem`

vulnerabilities/importers/redhat.py
```diff
@@ -17,30 +17,33 @@
...
 
     def __enter__(self):
-
         self.redhat_response = fetch()
 
     def updated_advisories(self):
         processed_advisories = []
         for advisory_data in self.redhat_response:
-            processed_advisories.append(to_advisory(advisory_data))
+            processed_advisories.extend(to_advisory(advisory_data))
```
pombredanne, line 46: This is likely incorrect as https://github.com/nexB/vulnerablecode/blob/783292b44181f1ad0abaf650ff1d7cac95d26c7e/vulnerabilities/importers/redhat.py#L70 return a single `Advisory` instance here:
https://github.com/nexB/vulnerablecode/blob/783292b44181f1ad0abaf650ff1d7cac95d26c7e/vulnerabilities/importers/redhat.py#L129 and not a list or iterable that could be used in `extend()`

vulnerabilities/models.py
```diff
@@ -199,3 +200,25 @@ def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> Dat
...
+
+class VulnerabilitySeverity(models.Model):
+
+    scoring_system_choices = ((system.identifier, system.name) for system in scoring_systems.values())  # nopep8
+    vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
+    value = models.CharField(max_length=50, help_text="Example: 9.0, Important, High")
+    scoring_system = models.CharField(
+        max_length=50,
+        choices=scoring_system_choices,
+        help_text="Identifier for the scoring system used. Avaiable choices are: {} ".format(
```
pombredanne, line 213:
```suggestion
        help_text="Identifier for the scoring system used. Available choices are: {} ".format(
```
Also... it may be easier to read if you put the join below in a variable.

vulnerabilities/tests/test_nvd.py
pombredanne: We will store both based on the new suggested design here, each as their own `VulnerabilitySeverity` and that for each Reference that provides these.

vulnerabilities/tests/test_redhat_importer.py
```diff
@@ -189,8 +251,18 @@ def test_to_advisory(self):
...
+            "cvss3": {"cvss3_base_score": 6.0},
+        }
         for adv in data:
-            adv = redhat.to_advisory(adv)
-            adv.vuln_references = sorted(adv.vuln_references, key=lambda x: x.url)
-            found_data.add(adv)
+            with unittest.mock.patch(
+                "vulnerabilities.importers.redhat.requests.get", return_value=mock_resp
+            ):
+                adv = redhat.to_advisory(adv)
```
pombredanne, line 264: See my other comments on the returned value of to_advisory()

pombredanne (changes requested):
Almost there ! see my comments inline for your consideration.

vulnerabilities/importers/redhat.py
```diff
@@ -65,31 +67,74 @@ def fetch():
...
             if rpm_to_purl(rpm):
                 affected_purls.append(rpm_to_purl(rpm))
 
     references = []
-    if advisory_data.get("bugzilla"):
-        bugzilla = advisory_data.get("bugzilla")
+    bugzilla = advisory_data.get("bugzilla")
+    if bugzilla:
+        url = "https://bugzilla.redhat.com/show_bug.cgi?id={}".format(bugzilla)
+        bugzilla_data = requests.get(f"https://bugzilla.redhat.com/rest/bug/{bugzilla}").json()
```
pombredanne, line 80: Just as a side note, this is the kind of JSON we would likely need to store forever as back auditable evidence when we will do this later ... which likely calls for a central place where we fetch things from

vulnerabilities/models.py
```diff
@@ -199,3 +200,25 @@ def make_data_source(self, batch_size: int, cutoff_date: datetime = None) -> Dat
...
+class VulnerabilitySeverity(models.Model):
+
+    scoring_system_choices = ((system.identifier, system.name) for system in scoring_systems.values())  # nopep8
+    vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
+    value = models.CharField(max_length=50, help_text="Example: 9.0, Important, High")
+    scoring_system = models.CharField(
+        max_length=50,
+        choices=scoring_system_choices,
+        help_text="Identifier for the scoring system used. Available choices are: {} ".format(
+                  ", ".join(
```
pombredanne, line 214: For readability, an intermediate variable would be better for this

vulnerabilities/tests/test_redhat_importer.py
```diff
@@ -65,132 +67,81 @@ def test_to_advisory(self):
...
-                        qualifiers=OrderedDict(),
+                        qualifiers={},
                         subpath=None,
                     ),
                     PackageURL(
                         type="rpm",
                         namespace="redhat",
                         name="bash",
                         version="4.1.2-48.el6",
-                        qualifiers=OrderedDict(),
```
pombredanne, line 76: Feel free to remove OrderedDict from the whole codebase BTW since the dicts are always ordered now.

pombredanne (approved):
LGTM! let's merge.

PR #291, opened 2020-11-30
Improve GitHub importer

sbs2001:
* The github importer used to abandon a vulnerability if
  the affected package had invalid name. This PR changes that
  behaviour to instead  skip the package and import the vulnerability

* This commit also enables github importer to import vulnerabilities for python
  and rubygems packages. This is crucial since we can't  use safetydb

### DIFF
--- a/vulnerabilities/importer_yielder.py
+++ b/vulnerabilities/importer_yielder.py
@@ -179,7 +179,7 @@ IMPORTER_REGISTRY = [
         'data_source': 'GitHubAPIDataSource',
         'data_source_cfg': {
             'endpoint': 'https://api.github.com/graphql',
-            'ecosystems': ['MAVEN', 'NUGET', 'COMPOSER']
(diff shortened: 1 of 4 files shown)


pombredanne (approved):
This looks fine, yet having a single class at that package manager level that track the version API and the package URL type in one place

sbs2001:
@pombredanne correct me if I am wrong. The way I understood your suggestion is 

1. The github ecosystem -> package type conversion code is dupilicated. Once for the actual package type inference and secondly for setting the version api. 
2. To fix this move the package type info in the corresponding version api classes.

pombredanne:
> @pombredanne correct me if I am wrong. The way I understood your suggestion is
>     1. The github ecosystem -> package type conversion code is dupilicated. Once for the actual package type inference and secondly for setting the version api.
>     2. To fix this move the package type info in the corresponding version api classes.

Somehow my comment was truncated!
Yes you got it right

PR #295, opened 2020-12-11
Use GH action instead travis for CI.

sbs2001:


pombredanne (changes requested):
Please do not remove Travis right away: we need to ensure things work OK before.

sbs2001:
Done

pombredanne (approved):
Let's try this! :+1:

Issue #284, opened 2020-11-24
REST API: Bulk requests for packages and vulnerabilities


MarcelBochtler:
We're planning to [add support for vulnerablecode](https://github.com/oss-review-toolkit/ort/issues/3287) into the [ORT advisor](https://github.com/oss-review-toolkit/ort#advisor).
The existing advisor module for Nexus IQ Server is using a single POST request with all [PURLs](https://github.com/package-url/purl-spec) in the body to retrieve vulnerability informations about all the packages, see: https://github.com/oss-review-toolkit/ort/blob/master/clients/nexus-iq/src/main/kotlin/NexusIqService.kt#L109-L110

To be able to use VulnerableCode in a performant way, VulnerableCode's REST API should offer a bulk request API for packages and for vulnerabilities.

pombredanne:
@MarcelBochtler Excellent request! Let's design how you would like the API to behave.

MarcelBochtler:
As a GET request would not be feasible to handle all PURLs as a query parameter I'd suggest a POST endpoint containing required package information in its body.

To be consistent to the existing API, I'd suggest to offer this functionality for `packages` and for `vulnerabilites`.

## Packages

This should return the list of package details incl. the vulnerabilities' `reference_id`s

For my use case the PURL would be used in the body. E.g.
```json
{
    "packages": [
        {
            "purl": "pkg:pypi/jaraco.logging@1.5"
        },
        {
            "purl": "pkg:maven/org.hamcrest/hamcrest-core@1.3"
        }
    ]
}
```

As an alternative you could offer this API to use the other package identifiers like `type`, `namespace`, `name` and a `version`:
```json
{
    "packages": [
        {
            "identifier": {
                "format": "maven",
                "coordinates": {
                    "artifactId": "hamcrest-core",
                    "groupId": "org.hamcrest",
                    "version": "1.3"
                }
            }
        },
        {
            "identifier": {
                "coordinates": {
                    "name": "jaraco.logging",
                    "version": "1.5"
                }
            }
        }
    ]
}
```
This is based on [Sonatype's component details API](https://help.sonatype.com/iqserver/automating/rest-apis/component-details-rest-api---v2).
I didn't like the differently named fields for the coordinates, that's why I used the PURL implementation.

## Vulnerabilites
```json
{
    "vulnerabilities": [
        {
            "reference_id": "CVE-2020-15250"
        },
        {
            "reference_id": "CVE-2020-8352"
        }
    ]
}
```
[...]

PR #303, opened 2020-12-21
Add endpoints for bulk requesting vulnerabilities and packages

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/284

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -23,11 +23,14 @@
 
 from urllib.parse import unquote
 
+from django.db.models import Q
 from django.urls import reverse
 from django_filters import rest_framework as filters
 from packageurl import PackageURL
 from rest_framework import serializers
 from rest_framework import viewsets
+from rest_framework.decorators import action
+from rest_framework.response import Response
 
 from vulnerabilities.models import Package
 from vulnerabilities.models import Vulnerability
@@ -126,6 +129,26 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = PackageFilterSet
 
+    @action(detail=False, methods=["post"])
+    def fetch(self, request):
+
+        filter_list = Q()
+        # TODO: Do some validation here
+        for purl in request.data["packages"]:
+            filter_list |= Q(
+                **{k: v for k, v in PackageURL.from_string(purl).to_dict().items() if v}
+            )
+
+        res = Package.objects.filter(filter_list)
+        response = {}
+        for purl in request.data["packages"]:
+            response[purl] = {}
+            for p in res:
+                if p.package_url == purl:
+                    response[purl] = PackageSerializer(p, context={"request": request}).data
+
+        return Response(response)
+
 
 class VulnerabilityFilterSet(filters.FilterSet):
     vulnerability_id = filters.CharFilter(field_name="cve_id")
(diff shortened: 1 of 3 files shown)


vulnerabilities/api.py
```diff
@@ -141,3 +166,22 @@ class VulnerabilityViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+    # `fetch` is a placeholder
+    # TODO: Find a good name for this endpoint
+    @action(detail=False, methods=["post"])
+    def fetch(self, request):
+        filter_list = Q()
+        response = {}
+        # TODO: Do some validation here of request body
+        for cve_id in request.data["vulnerabilities"]:
+            filter_list |= Q(cve_id=cve_id)
```
tdruez, line 178: You could use the `__in` lookup instead.

tdruez (commented):
Why not using proper filtering support instead of custom endpoints?

sbs2001:
@tdruez  

>Why not using proper filtering support instead of custom endpoints?

The use case is different, since the request is going to be a `POST` and the request body is also slightly different hence needs some parsing.

vulnerabilities/api.py
sbs2001: Right :)

tdruez:
> @sbs2001 The use case is different, since the request is going to be a POST and the request body is also slightly different hence needs some parsing.

What's the reason for those actions to be POST in the first place?

sbs2001:
@tdruez  re
> What's the reason for those actions to be POST in the first place?

https://github.com/nexB/vulnerablecode/issues/284#issue-749738940 . The idea is essentially to retrieve vulnerability info of multiple detected packages via a single request.

tdruez:
@sbs2001 I get the idea, but again, why POST is your chosen solution over GET?

sbs2001:
@tdruez  typically something like ort could detect 100+ packages. If we use a GET request, then either:
1. we could make 100+ requests each retrieving a package. There is a concern about performance about this solution.  
2. Or we can a make a single GET request with a request body containing each purl. IMHO that is very unusual and rarely done.

So POST request makes sense, also it's tried and tested by sonatype at https://ossindex.sonatype.org/rest

sbs2001:
`POST /api/vulnerabilities/bulk-search ` 
For the request body :
```python
{
            "vulnerabilities": [
                "CVE-2009-1382",
                "CVE-2014-8242",
                "RANDOM-fictional-CVE"
                ]
            }
```
The response would look like 
```python
{
    "CVE-2009-1382": {
        "url": "http://127.0.0.1:8000/api/vulnerabilities/2/",
        "unresolved_packages": [],
        "resolved_packages": [
            {
                "url": "http://127.0.0.1:8000/api/packages/2/",
                "purl": "pkg:deb/debian/mimetex@1.74-1?distro=jessie"
            },
            {
                "url": "http://127.0.0.1:8000/api/packages/3/",
                "purl": "pkg:deb/debian/mimetex@1.50-1.1?distro=jessie"
            }
        ]
    },
    "CVE-2014-8242": {
        "url": "http://127.0.0.1:8000/api/vulnerabilities/1/",
        "unresolved_packages": [
            {
                "url": "http://127.0.0.1:8000/api/packages/1/",
                "purl": "pkg:deb/debian/librsync@0.9.7-10?distro=jessie"
            }
        ],
        "resolved_packages": []
    },
    "RANDOM-fictional-CVE": {}
}
```

sbs2001:
For ` POST /api/packages/bulk-search `
For the request body 
```python
{
            "packages": [
                "pkg:deb/debian/librsync@0.9.7-10?distro=jessie",
                "pkg:deb/debian/mimetex@1.50-1.1?distro=jessie",
                "pkg:deb/lalaland/doesnotexist@1.50-1.1?distro=jessie"
            ]
        }
```

the response would look like : 

```python 
{
    "pkg:deb/debian/librsync@0.9.7-10?distro=jessie": {
        "resolved_vulnerabilities": [],
        "unresolved_vulnerabilities": [
            {
                "url": "http://127.0.0.1:8000/api/vulnerabilities/1/",
                "vulnerability_id": "CVE-2014-8242"
            }
        ]
    },
    "pkg:deb/debian/mimetex@1.50-1.1?distro=jessie": {
        "resolved_vulnerabilities": [
            {
                "url": "http://127.0.0.1:8000/api/vulnerabilities/2/",
                "vulnerability_id": "CVE-2009-1382"
            },
            {
                "url": "http://127.0.0.1:8000/api/vulnerabilities/3/",
                "vulnerability_id": "CVE-2009-2459"
            }
        ],
        "unresolved_vulnerabilities": []
    },
    "pkg:deb/lalaland/doesnotexist@1.50-1.1?distro=jessie": {}
}
```

PR #305, opened 2021-01-05
added verbose plural name for PackageRelatedVulnerability model

Shivam-316:
This is my first PR for this project, I would really like to contribute more.


vulnerabilities/migrations/0002_auto_20210105_1725.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.0.7 on 2021-01-05 17:25
```
sbs2001, line 1: Could you squash this into one migration file ? 

To do this you would need to delete existing migrations and generate a migration script

vulnerabilities/models.py
```diff
@@ -146,6 +146,7 @@ class PackageRelatedVulnerability(models.Model):
 
     class Meta:
         unique_together = ("package", "vulnerability")
+        verbose_name_plural = "Package related vulnerabilities"
```
sbs2001, line 149: verbose_name_plural = "PackageRelatedVulnerabilities" would be better

sbs2001 (changes requested):

sbs2001:
Hey @Shivam-316  thanks for the PR see my review comments inline :)

PR #319, opened 2021-01-28
Make RedHat CVE import more robust

pombredanne:
This importer failed at times toward the end of the fetch step when
reaching (and trying repeatidly) to re-fetch the last page.
The fix consists in fetching larger batches at once (10K instead of 1k)
which measn fewer API calls and less risk to be throttled and properly
handling exceptions and HTTP response codes and breaking rather than
retrying. This also adds some minimal logging.

### DIFF
--- a/vulnerabilities/importers/redhat.py
+++ b/vulnerabilities/importers/redhat.py
@@ -54,22 +54,24 @@ def fetch():
     """
     cves = []
     page_no = 1
-    url_template = "https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={}"
+    url_template = (
+        "https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={}"
+    )
 
     cve_data = None
     while True:
         current_url = url_template.format(page_no)
         try:
-            print(f'Fetching: {current_url}')
+            print(f"Fetching: {current_url}")
             response = requests.get(current_url)
             if response.status_code != requests.codes.ok:
                 # TODO: log me
-                print(f'Failed to fetch results from {current_url}')
+                print(f"Failed to fetch results from {current_url}")
                 break
             cve_data = response.json()
         except Exception as e:
             # TODO: log me
-            msg = f'Failed to fetch results from {current_url}:\n{e}'
+            msg = f"Failed to fetch results from {current_url}:\n{e}"
             print(msg)
             break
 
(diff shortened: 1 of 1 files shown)


vulnerabilities/importers/redhat.py
```diff
@@ -35,71 +35,89 @@ class RedhatDataSource(DataSource):
...
-    response = []
+    """
+    Return a list of CVE data mappings fetched from the RedHat API.
+    See:
+        https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/index
+    """
+    cves = []
     page_no = 1
-    url = "https://access.redhat.com/hydra/rest/securitydata/cve.json?page={}"
+    url_template = "https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={}"
```
sbs2001, line 53: This is way better !

vulnerabilities/importers/redhat.py
```diff
@@ -35,71 +35,89 @@ class RedhatDataSource(DataSource):
...
-        summary=advisory_data["bugzilla_description"],
         cve_id=advisory_data["CVE"],
+        summary=advisory_data["bugzilla_description"],
         impacted_package_urls=affected_purls,
         vuln_references=references,
     )
 
 
 def rpm_to_purl(rpm_string):
+    # FIXME: there is code in scancode to handle RPM conversion AND this should
```
sbs2001, line 119: I tried nevra.py, but it's horrible here since, redhat has weird names with multiple `-`.  

The method used here is more conservative.

sbs2001:
@pombredanne  FWIW, https://github.com/nexB/vulnerablecode/pull/290 also makes some changes to redhat, but your changes probably won't conflict with them.

vulnerabilities/importers/redhat.py
pombredanne: > I tried nevra.py, but it's horrible here since, redhat has weird names with multiple -.

I need to see actual real examples of real RPM package names that nevra.py cannot parse. If so this is a serious bug. 

And https://github.com/nexB/vulnerablecode/blob/8efdbd190d1249e5f4bcda044f98ea50af8a86a4/vulnerabilities/tests/test_redhat_importer.py#L43 is not enough tests and foo is a not a real name:  s=you should avoid to make use fake names here IMHO

We really want to have that correct and correctly done in one place only, that we can package in a library if needed.

pombredanne:
Tests are missing and need to be added here.

sbs2001:
Could you apply black on this ?

sbs2001:
All is working fine now, merge this as per your convenience

Issue #321, opened 2021-01-29
Disable Etag


pombredanne:
The current Etag handling IMHO are more problematic than helpful at this stage, especially when there is a failure during import. See https://github.com/nexB/vulnerablecode/commit/255b7f9be723ec32273d772d073ce6f80cbdf016#diff-938a299f8406c1d3defaec48838bc4f6f1307635f5d2ba36e9777227cedbb383R46 

1. We are not using Etag correctly: Etags are designed to be passed back in HTTP headers. We should instead send the proper HTTP header as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match and integrate this finely in the processing as this typically needs to use streaming=True requests, and proper handling of the HTTP return code. In all cases this should end up being a single request, not a HEAD followed by another GET request
 
2. Etag are unlikely meant to stored long term in the DB like we are now
3. the `create_etag` function name is misleading as we are creating, checking and saving the Etag in this `create_etag` function
4. If some import fails, we have to manually delete the records in the DB to restart an import
5. the time it takes to download data seems to be very small when compared with the time to perform the import
6. Once the initial import is done, incremental import should be much smaller, making Etgas even less relevant

Therefore, I think we need to reconsider using Etag.

sbs2001:
@pombredanne  

Thanks for (1) ! The current implementation is so crude. 

All other points make 100% sense to me, except (6) . Since only way to know whether the data is new, is to check it against what we in the db one to one. We can't just skip a file. Hence metadata like etag is used.  Any other way to do this ?

sbs2001:
Maybe adding a flag, to skip interpretting etag would help ?

PR #322, opened 2021-01-29
Improve Ubuntu OVAL importer

pombredanne:
There are a few issues with the Ubuntu Oval importer that was stuck.
This resolves some of these issues:
- Do not fail when there is a version range that cannot be parsed as a RangeSpecifier(). 
- Improve error handling
- Add simple logging using print statements to help monitor import progress
- Also apply minor code formatting.
- I also added a code comment about the Etag handling which IMHO are more problematic than helpful at this stage, especially when there is a failure during import. See also #321

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,4 +54,5 @@ addopts =
     -rfEsxXw
     --strict
     --ignore setup.py
+    --ignore vulnerabilities/lib_oval.py
     --doctest-modules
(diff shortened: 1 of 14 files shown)


pombredanne:
Note that I will need to add tests before we can merge and also the Ubuntu importer still does NOT complete after running for a day.

sbs2001:
@pombredanne  Any luck with completing the whole import ?

pombredanne:
> @pombredanne Any luck with completing the whole import ?

@sbs2001 I stopped it after several hours and about ~8000 completed (e.g. after the first archive)

sbs2001:
I'm running the import at https://github.com/sbs2001/vulnerablecode/runs/1792912071?check_suite_focus=true

sbs2001:
@pombredanne  I think we should ditch pycodestyle and use black for both checking + formatting like https://github.com/nexB/skeleton/issues/13#issuecomment-722486247

Issue #315, opened 2021-01-27
Enable basic Django admin features


pombredanne:
we should be able to do minimal searching and filtering there

PR #330, opened 2021-01-30
Add django admin  functionality for searching and filtering objects

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/315 

### DIFF
--- a/vulnerabilities/admin.py
+++ b/vulnerabilities/admin.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
+# Copyright (c) nexB Inc. and others. All rights reserved.
 # http://nexb.com and https://github.com/nexB/vulnerablecode/
 # The VulnerableCode software is licensed under the Apache License version 2.0.
 # Data generated with VulnerableCode require an acknowledgment.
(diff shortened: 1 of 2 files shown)


vulnerabilities/admin.py
```diff
@@ -34,22 +34,24 @@
 
 @admin.register(Vulnerability)
 class VulnerabilityAdmin(admin.ModelAdmin):
-    pass
+    search_fields = ["cve_id"]
```
pombredanne, line 37: I really want to get #259 merged ASAP

vulnerabilities/models.py
```diff
@@ -144,6 +144,9 @@ class PackageRelatedVulnerability(models.Model):
     vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
     is_vulnerable = models.BooleanField()
 
+    def __str__(self):
+        return self.package.package_url + " " + self.vulnerability.cve_id
```
pombredanne, line 148:
```suggestion
        return f"{self.package.package_url} {self.vulnerability.cve_id}"
```

Also this may not be something we want to have everywhere just for Django admin display.

pombredanne (approved):
See my minor nit below.
We will need to revisit this

Issue #324, opened 2021-01-29
web UI: The package name should not be mandatory 


pombredanne:
It should be possible to "browse" for a type+namespace  for instance, and this is not possible for now

sbs2001:
I don't know how this and https://github.com/nexB/vulnerablecode/issues/320#issuecomment-769702534 would make sense together, without using partial purls .

Issue #320, opened 2021-01-28
Make UI work


pombredanne:
I cannot seem to be able to find anything except exact CVE ids using the web UI.
In particular, I cannot find packages. In contrast thing work find in the API.
I think we should either disable the UI entirely or make it work ASAP

sbs2001:
@pombredanne  re 

> I cannot find packages

The package name lookup is exact,  that's causing this issue. Do we need to be able to lookup packages via their partial names ? In any case that would be  a quick fix. 

>anything except exact CVE ids

partial CVE id lookup works too, 

Eg, 
![Screenshot 2021-01-29 10:04:08](https://user-images.githubusercontent.com/28975399/106232255-8df0b900-6219-11eb-86bc-1488fb5d56ef.png)

Or 
![image](https://user-images.githubusercontent.com/28975399/106232287-a6f96a00-6219-11eb-93b0-40977c98ba5b.png)


Could you elaborate the issue with CVE lookup ?

pombredanne:
ok, so I was doing some import and may be there were not finished.  
I found the UI where I cannot enter values for each Package URL field not intuitive
And the packages were found alright after all... BUT they are not visible until you scroll down, because there is a a big empty vertical space  and I did not see any results upfront.
We need to work to make this more intuitive and simple.
IMHO only one search field needed for now, and no wasted space.
Some like this may be:
![Screenshot from 2021-01-29 10-33-23-v2](https://user-images.githubusercontent.com/675997/106260029-41f54280-6220-11eb-8e6b-59132f0afb29.png)

pombredanne:
> Could you elaborate the issue with CVE lookup ?

That works correctly, that's a mistake of mine.

pombredanne:
I think we need to design this more. I do not like the 0 and 1 counts in the list ... rather than counts we should have links and real values displayed. Also there is a problem where in this case:
1. the versions are not sorted
2. pkg:pypi/lxml@4.2.1 is vulnerable to CVE-2020-27783 alright BUT it is patched by pkg:pypi/lxml@4.6.2  and this does still show up as a count of 0

![Screenshot from 2021-01-29 11-00-54](https://user-images.githubusercontent.com/675997/106261150-87663f80-6221-11eb-9416-44909ac2980a.png)

Then: 
![Screenshot from 2021-01-29 11-01-16](https://user-images.githubusercontent.com/675997/106261161-8a613000-6221-11eb-968a-c0e2b0cd9ce8.png)

[...]

PR #335, opened 2021-02-03
Improve UI

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/324, https://github.com/nexB/vulnerablecode/issues/320


vulnerabilities/templates/base.html
```diff
@@ -26,12 +27,19 @@
                         <a class ="navbar-item" href = "{% url 'vulnerability_search' %}" >
                         Explore
                         </a>
-                        <a class ="navbar-item" href = "{% url 'vulnerability_create' %}">
-                        Create
-                        </a>
+
```
pombredanne, line 30: That's a minor nit but you want to keep your HTML  properly formatted.
Here this would mean using consistent indentation and spaing (e.g. essentially as if it were Python)

sbs2001: Sure thing,  fixing it

PR #336, opened 2021-02-05
Add istio importer and tests

### LINKED ISSUE #302: Collect Istio
See https://istio.io/latest/news/security/

tushar912:
## Description
Added importer and tests for istio
Also added test files in <code>test_data/istio</code>

Fixes #302

### DIFF
--- a/vulnerabilities/importer_yielder.py
+++ b/vulnerabilities/importer_yielder.py
@@ -229,11 +229,27 @@ IMPORTER_REGISTRY = [
         "data_source_cfg": {"etags": {}},
     },
     {
+        "name": "apache_tomcat",
+        "license": "",
+        "last_run": None,
+        "data_source": "ApacheTomcatDataSource",
+        "data_source_cfg": {"etags": {}},
+    },
+    {
         "name": "apache_kafka",
         "license": "",
         "last_run": None,
         "data_source": "ApacheKafkaDataSource",
         "data_source_cfg": {},
+    },
+    {
+        'name': 'istio',
+        'license': '',
+        'last_run': None,
+        'data_source': 'IstioDataSource',
+        'data_source_cfg': {
+            'repository_url': 'https://github.com/istio/istio.io'
+        },
     },
 ]
 
--- a/vulnerabilities/importer_yielder.py
+++ b/vulnerabilities/importer_yielder.py
@@ -102,10 +102,7 @@ IMPORTER_REGISTRY = [
         "license": "",
         "last_run": None,
         "data_source": "SUSEBackportsDataSource",
-        "data_source_cfg": {
-            "url": "http://ftp.suse.com/pub/projects/security/yaml/",
-            "etags": {},
-        },
+        "data_source_cfg": {"url": "http://ftp.suse.com/pub/projects/security/yaml/", "etags": {}},
     },
     {
         "name": "suse_scores",
@@ -119,10 +116,7 @@ IMPORTER_REGISTRY = [
         "license": "",
         "last_run": None,
         "data_source": "DebianOvalDataSource",
-        "data_source_cfg": {
-            "etags": {},
-            "releases": ["wheezy", "stretch", "jessie", "buster"],
-        },
+        "data_source_cfg": {"etags": {}, "releases": ["wheezy", "stretch", "jessie", "buster"]},
     },
     {
         "name": "redhat",
@@ -136,9 +130,7 @@ IMPORTER_REGISTRY = [
         "license": "",
         "last_run": None,
         "data_source": "NVDDataSource",
-        "data_source_cfg": {
-            "etags": {},
-        },
+        "data_source_cfg": {"etags": {}},
     },
     {
         "name": "gentoo",
@@ -229,13 +221,6 @@ IMPORTER_REGISTRY = [
         "data_source_cfg": {"etags": {}},
     },
     {
-        "name": "apache_tomcat",
-        "license": "",
-        "last_run": None,
-        "data_source": "ApacheTomcatDataSource",
-        "data_source_cfg": {"etags": {}},
-    },
-    {
         "name": "apache_kafka",
         "license": "",
         "last_run": None,
@@ -243,13 +228,11 @@ IMPORTER_REGISTRY = [
         "data_source_cfg": {},
     },
     {
-        'name': 'istio',
-        'license': '',
-        'last_run': None,
-        'data_source': 'IstioDataSource',
-        'data_source_cfg': {
-            'repository_url': 'https://github.com/istio/istio.io'
-        },
+        "name": "istio",
+        "license": "apache-2.0",
+        "last_run": None,
+        "data_source": "IstioDataSource",
+        "data_source_cfg": {"repository_url": "https://github.com/istio/istio.io"},
     },
 ]
 
@@ -258,9 +241,7 @@ def load_importers():
 
     for importer in IMPORTER_REGISTRY:
         imp, created = Importer.objects.get_or_create(
-            name=importer["name"],
-            data_source=importer["data_source"],
-            license=importer["license"],
+            name=importer["name"], data_source=importer["data_source"], license=importer["license"]
         )
 
         if created:
--- a/vulnerabilities/importers/istio.py
+++ b/vulnerabilities/importers/istio.py
@@ -22,6 +22,7 @@
 
 import asyncio
 from typing import List, Set
+import yaml
 
 from dephell_specifier import RangeSpecifier
 from packageurl import PackageURL
@@ -80,31 +81,27 @@ class IstioDataSource(GitDataSource):
         safe_pkg_versions = set(all_version_list) - set(vuln_pkg_versions)
         return safe_pkg_versions, vuln_pkg_versions
 
-    def get_data_from_md(self, file):
-        data = {}
-        for line in file:
+    def get_data_from_yaml_lines(self, yaml_lines):
+
+        return yaml.safe_load("\n".join(yaml_lines))
+
+    def get_yaml_lines(self, lines):
+
+        for line in lines:
             line = line.strip()
-            line = line.split()
-            if len(line) > 0 and line is not None:
+            if line.startswith("---"):
+                continue
+            elif line.endswith("---"):
+                break
+            else:
+                yield line
 
-                start = line[0]
+    def process_file(self, path):
 
-                if start == "title:":
-                    data["title"] = " ".join(line[1:])
-                elif start == "description:":
-                    data["description"] = " ".join(line[1:])
-                elif start == "cves:":
-                    data["cves"] = " ".join(line[1:])
-                    data["cves"] = data["cves"].replace("[", "")
-                    data["cves"] = data["cves"].replace("]", "")
-                    data["cves"] = data["cves"].split(",")
+        advisories = []
 
-                elif start == "releases:":
-                    data["releases"] = " ".join(line[1:])
-                    data["releases"] = data["releases"].replace("[", "")
-                    data["releases"] = data["releases"].replace("]", "")
-                    data["releases"] = data["releases"].replace('"', "")
-                    data["releases"] = data["releases"].split(",")
+        data = self.get_data_from_md(path)
+
         releases = []
         if data.get("releases"):
             for release in data["releases"]:
@@ -114,60 +111,54 @@ class IstioDataSource(GitDataSource):
                     lbound = ">=" + release[0]
                     ubound = "<=" + release[2]
                     releases.append(lbound + "," + ubound)
+
         data["releases"] = releases
 
-        return data
+        if not data.get("cves"):
+            data["cves"] = [""]
 
-    def process_file(self, path):
+        for cve_id in data["cves"]:
 
-        advisories = []
+            if not cve_id.startswith("CVE"):
+                continue
 
-        with open(path) as f:
-            data = {}
+            safe_pkg_versions = []
+            vuln_pkg_versions = []
 
-            data = self.get_data_from_md(f)
+            if not data.get("releases"):
+                data["releases"] = []
 
-            if not data.get("cves"):
-                data["cves"] = [""]
+            safe_pkg_versions, vuln_pkg_versions = self.get_versions_for_pkg_from_range_list(
+                data["releases"]
+            )
 
-            for cve_id in data["cves"]:
+            safe_purls = []
+            vuln_purls = []
 
-                if not cve_id.startswith("CVE"):
-                    continue
+            cve_id = cve_id
 
-                safe_pkg_versions = []
-                vuln_pkg_versions = []
+            safe_purls = {
+                PackageURL(name="istio", type="golang", version=version)
+                for version in safe_pkg_versions
+            }
 
-                if not data.get("releases"):
-                    data["releases"] = []
+            vuln_purls = {
+                PackageURL(name="istio", type="golang", version=version)
+                for version in vuln_pkg_versions
+            }
 
-                (
-                    safe_pkg_versions,
-                    vuln_pkg_versions,
-                ) = self.get_versions_for_pkg_from_range_list(data["releases"])
-
-                safe_purls = []
-                vuln_purls = []
-
-                cve_id = cve_id
-
-                safe_purls = {
-                    PackageURL(name="istio", type="golang", version=version)
-                    for version in safe_pkg_versions
-                }
-
-                vuln_purls = {
-                    PackageURL(name="istio", type="golang", version=version)
-                    for version in vuln_pkg_versions
-                }
-
-                advisories.append(
-                    Advisory(
-                        summary=data["description"],
-                        impacted_package_urls=vuln_purls,
-                        resolved_package_urls=safe_purls,
-                        cve_id=cve_id,
-                    )
+            advisories.append(
+                Advisory(
+                    summary=data["description"],
+                    impacted_package_urls=vuln_purls,
+                    resolved_package_urls=safe_purls,
+                    cve_id=cve_id,
                 )
+            )
 
         return advisories
+
+    def get_data_from_md(self, path):
+        with open(path) as f:
+            yaml_lines = self.get_yaml_lines(f)
+            return self.get_data_from_yaml_lines(yaml_lines)
--- a/vulnerabilities/importers/istio.py
+++ b/vulnerabilities/importers/istio.py
@@ -1,0 +1,173 @@
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+import asyncio
+from typing import List, Set
+
+from dephell_specifier import RangeSpecifier
+from packageurl import PackageURL
+
+from vulnerabilities.data_source import Advisory, GitDataSource, Reference
+from vulnerabilities.package_managers import GitHubTagsAPI
+
+
+class IstioDataSource(GitDataSource):
+    def __enter__(self):
+        super(IstioDataSource, self).__enter__()
+
+        if not getattr(self, "_added_files", None):
+            self._added_files, self._updated_files = self.file_changes(
+                recursive=True, file_ext="md", subdir="./content/en/news/security"
+            )
+        self.version_api = GitHubTagsAPI()
+        self.set_api()
+
+    def set_api(self):
+        asyncio.run(self.version_api.load_api(["istio/istio"]))
+
+    def updated_advisories(self) -> Set[Advisory]:
+        files = self._updated_files
+        advisories = []
+        for f in files:
+            processed_data = self.process_file(f)
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def added_advisories(self) -> Set[Advisory]:
+        files = self._added_files
+        advisories = []
+        for f in files:
+            processed_data = self.process_file(f)
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def get_versions_for_pkg_from_range_list(self, version_range_list):
+        # Takes a list of version ranges(affected) of a package
+        # as parameter and returns a tuple of safe package versions and
+        # vulnerable package versions
+
+        safe_pkg_versions = []
+        vuln_pkg_versions = []
+        all_version_list = self.version_api.get("istio/istio")
+        if not version_range_list:
+            return all_version_list, []
+        version_ranges = {RangeSpecifier(r) for r in version_range_list}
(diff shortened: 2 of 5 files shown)


vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+            return all_version_list, []
+        version_ranges = {RangeSpecifier(r) for r in version_range_list}
+        for version in all_version_list:
+            if any([version in v for v in version_ranges]):
+                vuln_pkg_versions.append(version)
+
+        safe_pkg_versions = set(all_version_list) - set(vuln_pkg_versions)
+        return safe_pkg_versions, vuln_pkg_versions
+
+    def get_data_from_md(self, file):
```
sbs2001, line 83: This can be made moe simple and robust by following the pattern at https://github.com/nexB/vulnerablecode/blob/6609bcfdebb4c28bd30dfd9d42011323d238ee4c/vulnerabilities/importers/rust.py#L236 except in this case yaml lines would be isolated from the md file then parsed into  a dictionary

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        data["releases"] = releases
+
+        return data
+
+    def process_file(self, path):
+
+        advisories = []
+
+        with open(path) as f:
+            data = {}
```
sbs2001, line 126: Why declare this, when it is being are captured  below. This makes it redundant

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+                if not cve_id.startswith("CVE"):
+                    continue
+
+                safe_pkg_versions = []
+                vuln_pkg_versions = []
+
+                if not data.get("releases"):
+                    data["releases"] = []
+
+                (
```
sbs2001, line 144: No need to wrap them in a tuple

sbs2001 (changes requested):

vulnerabilities/importers/istio.py
tushar912: Oh , I should have realized this earlier .Thanks I would make the change

vulnerabilities/importers/istio.py
tushar912: Ok

vulnerabilities/importers/istio.py
tushar912: Ok

vulnerabilities/importers/istio.py
pombredanne: This looks very much like a markdown file with a structured front matter in YAML.
So not exactly like Rust but almost (replace toml with yaml. and use `saneyaml` to parse this)
See https://istio.io/latest/about/contribute/front-matter/ that is a clear hint IMHO. You should not do your own parsing.

vulnerabilities/tests/test_istio.py
```diff
@@ -0,0 +1,120 @@
...
+from vulnerabilities.data_source import Advisory, Reference
+from vulnerabilities.importers.istio import IstioDataSource
+from vulnerabilities.package_managers import GitHubTagsAPI
+
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class TestIstioDataSource(TestCase):
+    @classmethod
+    def setUpClass(cls):
```
pombredanne, line 38: Do you really need a setup when you have exactly one test function?
It creates IMHO un-necessary coupling between all the future unit tests and this is not meant to be shared for now in anycase.

Also you likely want to decouple the testing of the low level parsing of the .md with yaml frontmatter and the part that creates proper advisories from that

pombredanne (changes requested):
Thank you for this! I have added a few comments for your consideration.

vulnerabilities/importer_yielder.py
```diff
@@ -244,6 +244,15 @@
             "etags": {}
         },
     },
+    {
+        'name': 'istio',
+        'license': '',
```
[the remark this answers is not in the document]
tushar912: Sure

vulnerabilities/importers/istio.py
tushar912: I changed it to use yaml parsing.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+
+from dephell_specifier import RangeSpecifier
+from packageurl import PackageURL
+
+from vulnerabilities.data_source import Advisory, GitDataSource, Reference
+from vulnerabilities.package_managers import GitHubTagsAPI
+
+
+class IstioDataSource(GitDataSource):
+    def __enter__(self):
```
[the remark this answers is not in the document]
tushar912: This context manager is for initializing the <code>GitHubTagsAPI</code> and loading the data in its cache. The getattr is just to make sure that the file changes are updated only if they don't exist previously.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        files = self._added_files
+        advisories = []
+        for f in files:
+            processed_data = self.process_file(f)
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def get_versions_for_pkg_from_range_list(self, version_range_list):
+        # Takes a list of version ranges(affected) of a package
```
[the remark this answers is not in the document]
tushar912: Sure

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        # as parameter and returns a tuple of safe package versions and
+        # vulnerable package versions
+
+        safe_pkg_versions = []
+        vuln_pkg_versions = []
+        all_version_list = self.version_api.get("istio/istio")
```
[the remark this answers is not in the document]
tushar912: Sure, I would change them.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        if not version_range_list:
```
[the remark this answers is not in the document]
tushar912: Ok

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+            return all_version_list, []
```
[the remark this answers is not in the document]
tushar912: I guess if there is no version range in releases that is there is no version that is vulnerable all of them should be safe.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        version_ranges = {RangeSpecifier(r) for r in version_range_list}
+        for version in all_version_list:
+            if any([version in v for v in version_ranges]):
```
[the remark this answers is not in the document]
tushar912: <code>version_ranges</code> is a list of range objects I don't think we could use <code>if version in version_ranges</code> here as version is in a range which is part of the set not in the set itself.

vulnerabilities/tests/test_data/istio/test_file.md
```diff
@@ -0,0 +1,15 @@
+---
```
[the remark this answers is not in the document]
tushar912: Yes, I changed it to use yaml parsing

vulnerabilities/tests/test_istio.py
tushar912: Ok

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        # Takes a list of version ranges(affected) of a package
+        # as parameter and returns a tuple of safe package versions and
+        # vulnerable package versions
+
+        safe_pkg_versions = []
+        vuln_pkg_versions = []
+        all_version_list = self.version_api.get("istio/istio")
+        if not version_range_list:
+            return all_version_list, []
+        version_ranges = {RangeSpecifier(r) for r in version_range_list}
```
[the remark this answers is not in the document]
tushar912: I could have used either of set or list .I could change to list if you say.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+        for version in all_version_list:
+            if any([version in v for v in version_ranges]):
+                vuln_pkg_versions.append(version)
+
+        safe_pkg_versions = set(all_version_list) - set(vuln_pkg_versions)
```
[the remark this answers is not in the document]
tushar912: I could not have used sets earlier as they don't support append operation and I needed sets to use the difference operation.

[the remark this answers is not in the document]
tushar912: So I used the set typecast

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+                    vuln_pkg_versions,
+                ) = self.get_versions_for_pkg_from_range_list(data["releases"])
+
+                safe_purls = []
+                vuln_purls = []
+
+                cve_id = cve_id
+
+                safe_purls = {
+                    PackageURL(name="istio", type="golang", version=version)
```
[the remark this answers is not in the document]
tushar912: Ok

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+                    for version in safe_pkg_versions
+                }
+
+                vuln_purls = {
+                    PackageURL(name="istio", type="golang", version=version)
```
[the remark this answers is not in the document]
tushar912: Sure

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+
+        advisories = []
+
+        with open(path) as f:
+            data = {}
+
+            data = self.get_data_from_md(f)
+
+            if not data.get("cves"):
+                data["cves"] = [""]
```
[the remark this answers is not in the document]
tushar912: If an advisory doesn't have <code>cve_id</code> we may still want to consider it as having "" as <code>cve_id</code>

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,173 @@
...
+
+            for cve_id in data["cves"]:
+
+                if not cve_id.startswith("CVE"):
```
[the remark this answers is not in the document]
tushar912: Do you mean the case of the word "CVE" .

vulnerabilities/importers/istio.py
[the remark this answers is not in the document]
tushar912: @pombredanne I am still unable to understand how to create two package types for a single purl. Can you explain it a bit.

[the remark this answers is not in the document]
pombredanne: @tushar912 sorry if I was not clear: I meant to have multiple purls: one `pkg:/golang` and one `pkg:github` since the project has both personalities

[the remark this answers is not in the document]
tushar912: Oh Ok

tushar912:
@pombredanne @sbs2001 I have made some changes as per the suggestions made by you. Please review again.

vulnerabilities/importer_yielder.py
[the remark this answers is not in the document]
sbs2001: I guess it's apache-2 something from https://github.com/istio/istio.io/blob/master/LICENSE ?

vulnerabilities/importers/istio.py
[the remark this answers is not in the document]
sbs2001: the condition and `all_version` should be at top. We don't want to create objects just to have them destroyed.

vulnerabilities/importers/istio.py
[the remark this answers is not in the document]
sbs2001: @pombredanne  The codebase has (over) use of sets all over the place mostly because we were paranoid of duplicates and order didn't matter.

@tushar912  use a list, if that's ok ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+                    summary=data["description"],
+                    impacted_package_urls=vuln_purls,
+                    resolved_package_urls=safe_purls,
+                    cve_id=cve_id,
+                )
+            )
+
+        return advisories
+
+    def get_data_from_md(self, path):
```
sbs2001, line 198: Nice !

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+        vuln_pkg_versions = []
+        all_version = self.version_api.get("istio/istio")
+        if not version_range_list:
+            return all_version, []
+        version_ranges = {RangeSpecifier(r) for r in version_range_list}
+        for version in all_version:
+            if any([version in v for v in version_ranges]):
+                vuln_pkg_versions.append(version)
+
+        safe_pkg_versions = set(all_version) - set(vuln_pkg_versions)
```
sbs2001, line 83: @pombredanne  this really captures how we do the classification, and need to change this. 

@tushar912  Ignore this .

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+    def updated_advisories(self) -> Set[Advisory]:
+        files = self._updated_files
+        advisories = []
+        for f in files:
+            processed_data = self.process_file(f)
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def added_advisories(self) -> Set[Advisory]:
```
sbs2001, line 59: No need to implement this method

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+        if data.get("releases"):
+            for release in data["releases"]:
+                release = release.strip()
+                release = release.split(" ")
+                if len(release) > 2:
+                    lbound = ">=" + release[0]
+                    ubound = "<=" + release[2]
+                    releases.append(lbound + "," + ubound)
+
+        data["releases"] = releases
```
sbs2001, line 141: This a list of string version ranges now, change the key name here to something that reflects that

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+            if not data.get("releases"):
+                data["releases"] = []
+
+            safe_pkg_versions, vuln_pkg_versions = self.get_versions_for_pkg_from_range_list(
+                data["releases"])
+
+            safe_purls = []
+            vuln_purls = []
+
+            cve_id = cve_id
```
sbs2001, line 163: Why ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+                release = release.strip()
+                release = release.split(" ")
+                if len(release) > 2:
+                    lbound = ">=" + release[0]
+                    ubound = "<=" + release[2]
+                    releases.append(lbound + "," + ubound)
+
+        data["releases"] = releases
+
+        if not data.get("cves"):
```
sbs2001, line 143: This could fail like it did in the last PR where there was case, where something like `data["cves"]=None` happens ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,205 @@
...
+            data["cves"] = [""]
+
+        for cve_id in data["cves"]:
+
+            if not cve_id.startswith("CVE"):
```
sbs2001, line 148: I wonder what when this condition gets satisfied ?  If it doesn't have a `CVE` then that makes it even more special and we should collect it :)

sbs2001 (changes requested):

vulnerabilities/importers/istio.py
tushar912: Thanks

vulnerabilities/importers/istio.py
tushar912: Ok

vulnerabilities/importers/istio.py
tushar912: Sure

vulnerabilities/importers/istio.py
tushar912: Oh, I should remove this

vulnerabilities/importers/istio.py
tushar912: Here if <code>data["cves']= None</code> then this condition satisfies . So I think this should work ?

tushar912: Oh I understood you mean if its a string named <code>'None'</code> .Am I right ?

vulnerabilities/importers/istio.py
tushar912: Actually I added that condition because some files had cves with 'N/A' . I think I got it wrong we should have "" as cve for such cases am I right?

vulnerabilities/importers/istio.py
tushar912: If I remove this method the vulnerabilities are not getting written to database.

vulnerabilities/importers/istio.py
sbs2001: yes we should have empty string as cve. But in any case, we don't give up the data, create cve-less advisory if required.

vulnerabilities/importers/istio.py
sbs2001: That's because when you run the importer first time, it saves a timestamp in your db. For the subsequent runs, a diff is calulated wrt to the stored timestamp and files are yielded accordingly. 

You can try  something like :
```
z = models.Importer.objects.get(name=<your_importer_name>)
z.last_run = None
z.save()
```
and run it again

vulnerabilities/importers/istio.py
tushar912: I think this case will be taken care by 
```python
if not cve_id.startswith("CVE"):
                cve_id = ""
```

tushar912:
@sbs2001 I have made the changes.

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,193 @@
...
+            safe_pkg_versions = []
+            vuln_pkg_versions = []
+
+            if not data.get("release_ranges"):
+                data["release_ranges"] = []
+
+            safe_pkg_versions, vuln_pkg_versions = self.get_versions_for_pkg_from_range_list(
+                                                   data["release_ranges"])
+
+            safe_purls = []
```
sbs2001, line 150: Is there a need to declare this here, since you are assigning them the union below ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,193 @@
...
+    def updated_advisories(self) -> Set[Advisory]:
+        files = self._updated_files
+        advisories = []
+        for f in files:
+            processed_data = self.process_file(f)
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def get_versions_for_pkg_from_range_list(self, version_range_list):
```
sbs2001, line 59: The function name is too verbose. How about  `get_pkg_versions_from_ranges` ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,193 @@
...
+        """Takes a list of version ranges(affected) of a package
+        as parameter and returns a tuple of safe package versions and
+        vulnerable package versions"""
+        all_version = self.version_api.get("istio/istio")
+        if not version_range_list:
+            return all_version, []
```
sbs2001, line 65: Could you point me to an advisory where `if not version_range_list` is True ?

sbs2001 (changes requested):

vulnerabilities/importers/istio.py
tushar912: I should remove it.

vulnerabilities/importers/istio.py
tushar912: Sure

vulnerabilities/importers/istio.py
tushar912: I did not find such case .I added this just to be sure that version_range_list is not none.  Also while going through the advisories I have found that releases also contain a case like "All versions prior to ..." which I missed .I would make change for this ASAP.

tushar912: If you insist I can remove <code>if not version_range_list</code> case.

tushar912:
@sbs2001 I have made the changes, also can you answer https://github.com/nexB/vulnerablecode/pull/336#discussion_r579107582.

vulnerabilities/importers/istio.py
sbs2001: > If you insist I can remove if not version_range_list case.

no :) , I was just curious.

vulnerabilities/importers/istio.py
sbs2001: No I was an idiot , and mistook this for another case, sorry . This fine as it is

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,199 @@
...
+                    releases.append("<" + release[4])
+                # If it is of form "a to b"
+                elif "to" in release:
+                    release = release.strip()
+                    release = release.split(" ")
+                    lbound = ">=" + release[0]
+                    ubound = "<=" + release[2]
+                    releases.append(lbound + "," + ubound)
+                # If it is a single release
+                elif isinstance(release, int):
```
sbs2001, line 137: Could you point me to an example of this ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,199 @@
...
+            safe_purls_golang = {
+                PackageURL(type="golang", name="istio", version=version)
+                for version in safe_pkg_versions
+            }
+
+            safe_purls_github = {
+                PackageURL(type="github", name="istio", version=version)
+                for version in safe_pkg_versions
+            }
+            safe_purls = safe_purls_github | safe_purls_golang
```
sbs2001, line 168: Nit : Maybe use `setA.union(setB)` ?

sbs2001 (changes requested):
@tushar912  thanks this almost at a stage to be merged, please see my comments inline.

vulnerabilities/importers/istio.py
sbs2001: I realized this is black's effect. Interesting

sbs2001:
@tushar912  Due to recent model  changes, you would need to rename `Advisory.cve_id` ->`Advisory.vulnerability_id`

vulnerabilities/importers/istio.py
tushar912: Sure

vulnerabilities/importers/istio.py
tushar912: https://github.com/istio/istio.io/blob/master/content/en/news/security/istio-security-2020-004/index.md

sbs2001: Could this be done without using `isinstance` ? This is going to fail in case the release is of form `1.x.y` . Maybe some string check ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,199 @@
...
+            if processed_data:
+                advisories.extend(processed_data)
+        return self.batch_advisories(advisories)
+
+    def get_pkg_versions_from_ranges(self, version_range_list):
+        """Takes a list of version ranges(affected) of a package
+        as parameter and returns a tuple of safe package versions and
+        vulnerable package versions"""
+        all_version = self.version_api.get("istio/istio")
+        # if not version_range_list:
```
sbs2001, line 64: I guess this can be removed now ?

sbs2001 (changes requested):
See my comment of usage of `isinstance`  and remove the commented code and we're good to merge :)

vulnerabilities/importers/istio.py
tushar912: Should I use a regex check?

vulnerabilities/importers/istio.py
tushar912: Oh I missed this. I will remove it.

tushar912:
@sbs2001 Sorry for the delay. I have added a better check for single release using regex.

vulnerabilities/importers/istio.py
sbs2001: I would try to avoid regex unless absolutely needed, because those are very cryptic when reading.

If there's no other way could you do something like https://github.com/nexB/vulnerablecode/blob/f81816fcd95d00736dc17c9aaad20e91cb6ce201/vulnerabilities/helpers.py#L81  ?

vulnerabilities/importers/istio.py
```diff
@@ -0,0 +1,199 @@
...
+                    releases.append("<" + release[4])
+                # If it is of form "a to b"
+                elif "to" in release:
+                    release = release.strip()
+                    release = release.split(" ")
+                    lbound = ">=" + release[0]
+                    ubound = "<=" + release[2]
+                    releases.append(lbound + "," + ubound)
+                # If it is a single release
+                elif is_release(release):
```
sbs2001, line 134: Nice

sbs2001 (approved):
Thanks @tushar912  for your patience and I'm merging this.

PR #337, opened 2021-02-06
Add message when no vulnerabilities are found for a vuln_id

### LINKED ISSUE #328: Searching for vulnerability id which is not present does not show any message
Currently if we search for a vulnerability id which is not present the search page reloads without change. It would be better to have a message like "no result found" .

tushar912:
Fixes #328 

vul


sbs2001 (approved):
Thanks !

PR #342, opened 2021-02-09
Collect kafka cves

sbs2001:


sbs2001:
https://issues.apache.org/jira/browse/KAFKA-12320 is for providing structured data.

PR #347, opened 2021-02-12
Fix regex in schema validator in alpine importer

sbs2001:


sbs2001:
@rolfschr  could you fix the nix tests ?

rolfschr:
@sbs2001 Yes, I will have a look into it.

Issue #334, opened 2021-02-03
Use immutable dataclasses


sbs2001:

sbs2001:
Realized immutable dataclasses are of limited value since they reduce flexibility in data sources and duplication is not a big issue

PR #363, opened 2021-02-25
Refactor codebase and tests to treat Advisory class mutable

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/334


vulnerabilities/importers/alpine_linux.py
```diff
@@ -187,7 +187,7 @@ def _load_advisories(
                         impacted_package_urls=[],
                         resolved_package_urls=resolved_purls,
                         vuln_references=references,
-                        vulnerability_id=vuln_ids[0] if vuln_ids[0] != "CVE-????-?????" else None,
+                        vulnerability_id=vuln_ids[0] if vuln_ids[0] != "CVE-????-?????" else "",
```
pombredanne, line 190: Add TODO or ticket to revisit this cryptic code

vulnerabilities/data_source.py
```diff
@@ -209,7 +200,7 @@ def batch_advisories(self, advisories: List[Advisory]) -> Set[Advisory]:
 
         while advisories:
             b, advisories = advisories[: self.batch_size], advisories[self.batch_size:]
-            yield set(b)
+            yield b
```
pombredanne, line 203: Add TODO or ticket to revisit this cryptic code

pombredanne: Also do not use incorrect type hints as in `Set[Advisory]`... nothing is better than wrong.

vulnerabilities/data_source.py
```diff
@@ -78,19 +78,10 @@ class Advisory:
     """
 
     summary: str
-    impacted_package_urls: Iterable[PackageURL]
+    vulnerability_id: Optional[str] = None
+    impacted_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
```
pombredanne, line 82: IMHO you should have  a way to compare and sort Advisory.
This would make the tests super simpler. See https://docs.python.org/3/library/functools.html?highlight=functools#functools.total_ordering

vulnerabilities/tests/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        vuln_references=vuln_references,
+    )
+
+
+# This is not entirely correct, but enough for testing purpose.
+def advisory_sort_key(advisory):
+    return advisory.vulnerability_id
+
+
+def advisories_are_equal(expected_advisories, found_advisories):
```
pombredanne, line 58: See above. You should implmenent https://docs.python.org/3/library/functools.html?highlight=functools#functools.total_ordering for an Advisory for simpler and clearer code.

With that, the `advisories_are_equal` would become:
`sorted(expected_advisories) == sorted(found_advisories)`
and could be removed

vulnerabilities/tests/utils.py
```diff
@@ -0,0 +1,68 @@
...
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+
+from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import Reference
+
+
+def normalized_reference(reference):
```
pombredanne, line 27: IMHO make this a method of Reference

vulnerabilities/tests/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    severities = sorted(
+        reference.severities, key=lambda severity: (severity.value, severity.system.identifier)
+    )
+
+    return Reference(reference_id=reference.reference_id, url=reference.url, severities=severities)
+
+
+def normalized_advisory(advisory):
```
pombredanne, line 35: IMHO move as a method of Advisory

vulnerabilities/tests/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        vulnerability_id=advisory.vulnerability_id,
+        impacted_package_urls=impacted_package_urls,
+        resolved_package_urls=resolved_package_urls,
+        vuln_references=vuln_references,
+    )
+
+
+# This is not entirely correct, but enough for testing purpose.
+def advisory_sort_key(advisory):
+    return advisory.vulnerability_id
```
pombredanne, line 55: See below using attrgetter

vulnerabilities/tests/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def advisories_are_equal(expected_advisories, found_advisories):
+
+    expected_advisories = list(filter(lambda x: isinstance(x, Advisory), expected_advisories))
+    expected_advisories.sort(key=advisory_sort_key)
```
pombredanne, line 61: Using `from operator import attrgetter` you could use:
```suggestion
    expected_advisories.sort(key=attrgetter('vulnerability_id`))
```
Or...
```suggestion
    expected_advisories.sort(key=lambda advisory: advisory.vulnerability_id)
```
But IMHO it is better to have that in the Advisory class as suggested here.

pombredanne (approved):
Looking good, see a few nits for your consideration. And merge!

PR #370, opened 2021-03-06
Update pypi deps db to 2021-03-06.

rolfschr:
Simple update of the pypi index.
With a working [Nix]() setup, all you have to do is running ``./etc/nix/get-latest-pypi-deps-db.sh`` and use its output to update ``/etc/nix/flake.nix``

### DIFF
--- a/etc/nix/flake.lock
+++ b/etc/nix/flake.lock
@@ -22,16 +22,16 @@
         "pypi-deps-db": "pypi-deps-db"
       },
       "locked": {
-        "lastModified": 1606460537,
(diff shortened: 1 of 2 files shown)


rolfschr:
I realize something is not working. I didn't check my local output before pushing. Sorry for that, I'll have to look into this.

rolfschr:
Seems to be an issue with the version of ``black``. See here https://github.com/DavHau/mach-nix/issues/250.

sbs2001:
@rolfschr  Thanks, merging this

Issue #348, opened 2021-02-12
Update deps  and their structure


sbs2001:
We are using pretty outdated deps. These should be upgraded, after all we don't want to be "vulnerable" . 

Unrelated: have requirement file separate dev dependencies.

sbs2001:
@rolfschr  heads up ! Btw the nix tests are failing at https://github.com/nexB/vulnerablecode/pull/347 due to update in  deps

Hritik14:
Repaste from #379 
>The pytest bug is weird. @Hritik14 could you try updating the venv pytest to latest and try running python -m pytest . Just to determine if issue is caused due to system pytest invocation or just some version incompability.

@sbs2001 I think you're right. It is failing because the tests are only compatible with pytest version shipped with the project requirements. 
Here is an asciinema reproducing my results:
[![asciicast](https://asciinema.org/a/WhwxQT9ANJQtvOo976mnqws6z.svg)](https://asciinema.org/a/WhwxQT9ANJQtvOo976mnqws6z)

As you can see, after first installation, `which pytest` points to the system pytest.
I have `pytest 6.2.2` installed and the project requires `pytest==5.3.2`.    
So, it is both due to system pytest invocation and version incompatibility.

Maybe we can look into this here

prathamaror:
I have a pretty good idea about this should I start working on this I have some questions prior

Hritik14:
This has been fixed by https://github.com/nexB/vulnerablecode/pull/379

PR #379, opened 2021-03-15
Correct API docs path and fix pytest invocation

Hritik14:
The earlier location http://127.0.0.1:8000/api/docs was a 404. As the
urls.py suggests, the README now reflects http://127.0.0.1:8000/api/schema/swagger-ui/

After the first install, activating the venv is done prior to installing
pytest, thus the system default pytest is being called. It is either
required to reactive venv or run pytest via python -m.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -245,7 +245,7 @@ Start the webserver::
 
 For full documentation about API endpoints use this URL::
 
-    http://127.0.0.1:8000/api/schema/swagger-ui/
+    http://127.0.0.1:8000/api/docs
 
 
 
--- a/README.rst
+++ b/README.rst
@@ -206,7 +206,7 @@ Run Tests
 Use these commands to run code style checks and the test suite::
 
     black -l 100 --check .
-    DJANGO_DEV=1 pytest
+    DJANGO_DEV=1 python -m pytest
 
 
 Data import
@@ -245,7 +245,7 @@ Start the webserver::
 
 For full documentation about API endpoints use this URL::
 
-    http://127.0.0.1:8000/api/docs
+    http://127.0.0.1:8000/api/schema/swagger-ui/
 
 
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
           pip install -r requirements.txt
     
       - name: Run tests
-        run: pytest
+        run: python -m pytest
         env:
           # The hostname, username used to communicate with the PostgreSQL service container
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -206,7 +206,7 @@ Run Tests
 Use these commands to run code style checks and the test suite::
 
     black -l 100 --check .
-    DJANGO_DEV=1 pytest
+    DJANGO_DEV=1 python -m pytest
```
pombredanne, line 209: If you are in an activate virtualenv, pytest should work. Did you experience an issue here?

Hritik14: Yes. As mentioned [here](https://stackoverflow.com/questions/35045038/how-do-i-use-pytest-with-virtualenv), it fails on the first run. I've generated an [asciinema](https://asciinema.org/a/k7gQuPDlDxSU3iPCl8k3sVEi7) to reproduce the issue as well.

sbs2001:
The pytest bug is weird. @Hritik14  could you try updating the venv pytest to latest and try running `python -m pytest` . Just to determine if issue is caused due  to system pytest invocation or just some version incompability.

Hritik14:
@sbs2001 I think you're right. It is failing because the tests are only compatible with pytest version shipped with the project requirements. 
As you can see, after first installation, `which pytest` points to the system pytest.
I have `pytest 6.2.2` installed and the project requires `pytest==5.3.2`.    
So, it is both due to system pytest invocation and version incompatibility.

Maybe we can look into this in #348.

In any way, for now, I think the pytest invocation should always be the one shipped with the project itself so as to not _just break_ due to any system update.

sbs2001:
@Hritik14  thanks for confirming. Do you mind repasting your comment at https://github.com/nexB/vulnerablecode/issues/348, that'll be handy.

We'll use `python -m pytest` as you suggest. It probably doesn't matter much, but I want this to be consistent with the CI. You'll need to update to https://github.com/nexB/vulnerablecode/blob/ff08fd6044110bbd0b7b57010e3622b2fb25c4b6/.travis.yml#L21  and https://github.com/nexB/vulnerablecode/blob/ff08fd6044110bbd0b7b57010e3622b2fb25c4b6/.github/workflows/main.yml#L41

As for the api doc related changes. That's a good check. IMHO the docs should be at http://127.0.0.1:8000/api/docs, the `urls.py` need some changes.

Hritik14:
@sbs2001 Your wish is my command. :)

sbs2001 (approved):
@Hritik14  thanks for the PR, I'm merging this

PR #382, opened 2021-03-17
Explicity provide lxml parser to beautifulsoup

Hritik14:
BeautifulSoup would generate warnings without an explicit parser which
shows up in the test results as well.
It is always better to look at clean test results than one filled with
unnecessary warnings.


sbs2001:
@Hritik14 thanks for this :) , merging this

PR #389, opened 2021-03-18
Make sure vulnerability id is_cve or is_vulcoid

Hritik14:
This makes sure that vulnerability id supplied in alpine_linux importer
is either a cve, vulcoid or empty so as to stand on the definition of
vulnerability id.
It could be possible to introduce a validator at the model level for the
same as well using these functions.
I would propose that we create a validators class where we would keep some custom validators which could be used inside the models to keep the database protected against invalid input. One such validator for `vulnerability id` would look like the following
```python
    def vulnerability_id_validator(id: str):
        if id != "" and not (DataSource.is_cve(id) or DataSource.is_vulcoid(id)):
            raise ValidationError("{}: not a valid cve or volcoid".format(id))
```

### DIFF
--- a/vulnerabilities/data_source.py
+++ b/vulnerabilities/data_source.py
@@ -228,6 +228,26 @@ class DataSource(ContextManager):
             b, advisories = advisories[: self.batch_size], advisories[self.batch_size :]
             yield b
 
+    @staticmethod
+    def is_cve(id: str):
+        c = id.split("-")
+        if len(c) == 3 and c[0].lower() == "cve" and c[1].isdigit() and c[2].isdigit():
+            return True
+        return False
+
+    @staticmethod
+    def is_vulcoid(id: str):
+        c = id.split("-")
+        if (
+            len(c) == 4
+            and c[0].lower() == "vulcoid"
+            and c[1].isdigit()
+            and c[2].isdigit()
+            and c[3].isdigit()
+        ):
+            return True
+        return False
+
 
 @dataclasses.dataclass
 class GitDataSourceConfiguration(DataSourceConfiguration):
--- a/vulnerabilities/data_source.py
+++ b/vulnerabilities/data_source.py
@@ -44,6 +44,7 @@ from packageurl import PackageURL
 
 from vulnerabilities.oval_parser import OvalParser
 from vulnerabilities.severity_systems import ScoringSystem
+from vulnerabilities.helpers import is_cve
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sbs2001:
@Hritik14  detecting these when we are at last step, ie during model creation won't be as effective as detection on creation of `Advisory` instances. 

I would rather add the validation to the `Advisory` dataclass. Doing lot of work then failing is lot worse then failing before doing much in this case ;)

vulnerabilities/data_source.py
```diff
@@ -228,6 +228,26 @@ def batch_advisories(self, advisories: List[Advisory]) -> Set[Advisory]:
...
+        if (
+            len(c) == 4
+            and c[0].lower() == "vulcoid"
+            and c[1].isdigit()
+            and c[2].isdigit()
+            and c[3].isdigit()
+        ):
+            return True
+        return False
+
```
sbs2001, line 250: These are not required. `is-cve`  is duplicated we already a method for doing that validation at https://github.com/nexB/vulnerablecode/blob/3d66b4e82ee31422ab907d3388f739b768ffd2ac/vulnerabilities/helpers.py#L81

`is_vulcoid` is not needed since `vulcoids` are not assigned manually. The advisories without vulnerability id are assigned vulcoid, see https://github.com/nexB/vulnerablecode/blob/3d66b4e82ee31422ab907d3388f739b768ffd2ac/vulnerabilities/models.py#L65

vulnerabilities/importers/alpine_linux.py
```diff
@@ -193,7 +193,9 @@ def _load_advisories(
                         impacted_package_urls=[],
                         resolved_package_urls=resolved_purls,
                         references=references,
-                        vulnerability_id=vuln_ids[0] if vuln_ids[0] != "CVE-????-?????" else "",
+                        vulnerability_id=vuln_ids[0]
+                        if self.is_cve(vuln_ids[0]) or self.is_vulcoid(vuln_ids[0])
```
sbs2001, line 197: As said above, validate for only cve using the helper

sbs2001 (changes requested):

Hritik14:
I've applied the requested changes. Please review. Also, I think it would be better to squash and merge the PR to avoid intermediate commits.

sbs2001:
LGTM ! could you squash and force push the commits from your side ?

Hritik14:
I would do it but it would just mess up the git history (I'm not really in favor of --force). If you could squash like it's shown here I guess it would be lot easier https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-a-pull-request

sbs2001 (approved):
Thanks @Hritik14  merging this

PR #390, opened 2021-03-19
Resolves #380: Added severity in the view at http://127.0.0.1:8000/vulnerabilities/<…

### LINKED ISSUE #380: Display severities in the vulnerability view
In the view at http://127.0.0.1:8000/vulnerabilities/<vulnerability_id> we need to display the severity obtained from various data sources.

The severity must be nested inside the reference to show the connection.

Pushpit07:
Added severity in the view at http://127.0.0.1:8000/vulnerabilities/<vulnerability_id> by calling the scores() function for the particular reference and getting its value and scoring_system

Signed-off-by: Pushpit <pushpit07@gmail.com>


Pushpit07:
Now all the severity values and scoring_systems related to the particular reference is being shown

Screenshot 2021-03-19 at 16 25 30G
Screenshot 2021-03-19 at 16 26 01
Screenshot 2021-03-19 at 16 26 27




Screenshot 2021-03-19 at 16 24 53

Hritik14:
I guess if we could have a generalized version of severity (for ex, the one like  purl), it would be much better.

sbs2001:
@Hritik14  nope unifying them is the last thing we want to do. Severity changes according to the vendor. A score provided by redhat, would be given more importance when the found vulnerability is affecting a builtin redhat  package but archlinux would rate it low if it is a external package.

See the bottom section at https://access.redhat.com/security/updates/classification/

Hritik14:
By generalisation, I didn’t actually mean generalised to one number/score. I meant that we can use a general format that could be applicable everywhere.
A very rudimentary example would be ‘ASA:7.6’ or ‘RHSA:3.5’.

sbs2001:
@Pushpit07 I think continuing to use tables leaves us with lot of unused whitespace and the scores look too crammed up.

I think  we should switch using a list instead, with nested severities ? 

Something like 
References:-
  * Reference ID value  | Reference URL
     * Severity scoring system {scoring_system}  |  {scoring_value}
 
A practical example:-

```
References:-
  *AVG-853| https://security.archlinux.org/AVG-853
    *CVSS_3 | 8.5
     *CVSS_3_vector| CVSS:3/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:H   
```

Pushpit07:
@sbs2001 I feel that tables represent data in a much more organised and readable format. Implementing the items as a list would actually make the experience cumbersome. 
Instead, what I suggest is that we can provide a list of severities and scoring_system inside the table itself. Something like-


AVG-853  | https://security.archlinux.org/AVG-853 | * CVSS_3<br />*8.5  | * cvssv3  <br />  * cvssv3_vector  <br/> * cvssv2_vector 
--- | --- | --- | ---
AVG-652  | https://security.archlinux.org/AVG-652 | * CVSS_2<br />*7.0  | * cvssv2  <br />  * cvssv2_vector


What say?

sbs2001:
@Pushpit07  that's a interesting approach. The example you provided doesn't help though, why am I seeing `CVSS_3` in two columns ? How can I differentiate between what's the severity system and the value ? Please provide a clearer example :) 

>Implementing the items as a list would actually make the experience cumbersome.

Other than duplication of metadata, I don't see any other problem with this approach, maybe I'm missing something here ?

Pushpit07:
> @sbs2001 I feel that tables represent data in a much more organised and readable format. Implementing the items as a list would actually make the experience cumbersome.
> Instead, what I suggest is that we can provide a list of severities and scoring_system inside the table itself. Something like-
> 
> AVG-853	https://security.archlinux.org/AVG-853	* CVSS_3
> *8.5	* cvssv3 
> * cvssv3_vector 
> * cvssv2_vector
> AVG-652	https://security.archlinux.org/AVG-652	* CVSS_2
> *7.0	* cvssv2 
> * cvssv2_vector
> What say?

Re: 

--- | --- | --- | ---
AVG-853  | https://security.archlinux.org/AVG-853 | * CVSS_3<br />*8.5  | * cvssv3  <br />  * cvssv3_vector  <br/> * cvssv2_vector 
--- | --- | --- | ---
AVG-652  | https://security.archlinux.org/AVG-652 | * CVSS_2<br />*7.0  | * cvssv2  <br />  * cvssv2_vector

Pushpit07:
Reference ID | URL | Severity Value | Scoring system
--- | --- | --- | ---
AVG-853  | https://security.archlinux.org/AVG-853 | * CVSS_3<br />* 8.5  | * cvssv3  <br />  * cvssv3_vector  <br/> * cvssv2_vector 
AVG-652  | https://security.archlinux.org/AVG-652 | * CVSS_2<br />* 7.0  | * cvssv2  <br />  * cvssv2_vector

Pushpit07:
> @Pushpit07 that's a interesting approach. The example you provided doesn't help though, why am I seeing `CVSS_3` in two columns ? How can I differentiate between what's the severity system and the value ? Please provide a clearer example :)

Table headers would be there to differentiate though

Reference ID | URL | Severity Value | Scoring system
--- | --- | --- | ---
AVG-853  | https://security.archlinux.org/AVG-853 | * CVSS:3/AV:N/AC:L/PR:N/UI:N/S:U/C:H<br />* 8.5  | * cvssv3  <br />  * cvssv3_vector  <br/> * cvssv2_vector 
AVG-652  | https://security.archlinux.org/AVG-652 | * AV:N/AC:M/Au:N/C:C/I:C/A:C <br />* 7.0  | * cvssv2  <br />  * cvssv2_vector

Pushpit07:
> Other than duplication of metadata, I don't see any other problem with this approach, maybe I'm missing something here ?

Yeah, duplication is the main reason due to which I'm suggesting to avoid it

Pushpit07:
@sbs2001 Any updates on this? What to do?

sbs2001:
@Pushpit07  sorry for the late reply.  IMHO tables is the way to go, but we need some way to  denote the informational hierarchy. In simple terms we want to denote that  `severity_system -> severity_value`  pairs are child of the reference. 

The current approach brings  them to the same level as that of the reference's main attributes which is kinda misleading. 

Sorry for taking this long,I think this is not a good first issue, but you've been doing a good job. We need further design to fix the hierarchy thing and this will be done. I'm open to ideas here :)

Pushpit07:
If I understand correctly, is this right?

Screenshot 2021-03-24 at 16 43 53

sbs2001:
@Pushpit07  the severity value should be the child/sibling of the each individual of each scoring system. See the example at https://github.com/nexB/vulnerablecode/pull/390#issuecomment-802888700.

Pushpit07:
Oh, okay. This is what you mean?

Screenshot 2021-03-25 at 08 26 56

sbs2001:
@Pushpit07 yep that's what I mean't. I realized that we would want to nest them further. 
Something like 

```
References:-
  *AVG-853| https://security.archlinux.org/AVG-853
    *Severities
       System    |     Value
       *CVSS_3 | 8.5
       *CVSS_3_vector| CVSS:3/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:H   
```

That'd be easy to  be implement right ?

Pushpit07:
Yup. Piece of cake

AmitGupta7580:
Can we use this type of nested rows for severity represntation ?

image

sbs2001:
@AmitGupta7580  that would work in case most of the references have a nested severity. But in reality that is not the case. Most of references don't provide a score. So we would end up with lots of wasted whitespace, and the references would get cramped unnecessarily

AmitGupta7580:
image
image

What about this Tree structure View where user can expand particular reference and view its severity extracted from that reference.

sbs2001:
@AmitGupta7580  why are there multiple `references` nodes ?

AmitGupta7580:
> @AmitGupta7580 why are there multiple `references` nodes ?

@sbs2001 No these 2 images are same, I just expanded the 2nd reference.

Pushpit07:
@sbs2001  Apologies for the delay. Been ill for the last 4 days and running behind schedule. Will give updates in a few hours

AmitGupta7580:
@Pushpit07 Can you please brief me what are you trying to implement.

Pushpit07:
@sbs2001 Updated design

Screenshot 2021-04-01 at 15 58 04
Screenshot 2021-04-01 at 15 58 31

vulnerabilities/templates/base.html
```diff
@@ -6,6 +6,7 @@
 <link rel="stylesheet" href="{% static 'css/bulma.css' %}"/>
 <link rel="stylesheet" href="{% static 'css/custom.css' %}"/>
 <link rel="stylesheet" href="{% static 'css/font-awesome.css' %}"/>
+<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-eOJMYsd53ii+scO/bJGFsiCZc+5NDVN2yr8+0RDqr0Ql0h+rP48ckxlpbzKgwra6" crossorigin="anonymous">
```
sbs2001, line 9: We don't want to use bootstrap. We already use bulma. See https://bulma.io/documentation/elements/table/

vulnerabilities/static/css/custom.css
```diff
@@ -6,4 +6,8 @@
 
 .Site-content {
 flex: 1;
+}
+
+table {
```
sbs2001, line 11: I'm guessing this won't  be needed if we use bulma ?

sbs2001 (changes requested):
@Pushpit07 I like your design. 

Just don't use bootstrap, we already use bulma, using 2 css frameworks is a bad idea especially for a  small feature.

vulnerabilities/static/css/custom.css
Pushpit07: This was necessary. I removed the `is-bordered` class from the table so that cells are not divided by a line. That was causing difficulty in understanding the structure of severity

Pushpit07:
@sbs2001 Updated design using Bulma

Screenshot 2021-04-02 at 16 39 18
Screenshot 2021-04-02 at 16 39 31

Pushpit07:
Have a look @sbs2001

sbs2001:
@Pushpit07  the typography is all messed up 

image


The reference Id table is bigger than it's parent header. 

Also the table headers are now misleading. We have a `url` column which contains severities. 

IMHO just stick with simple lists for now  https://github.com/nexB/vulnerablecode/pull/390#issuecomment-806344755 . Don't do anything fancy.

Pushpit07:
With fixed alignment

Screenshot 2021-04-05 at 12 48 09

Pushpit07:
> IMHO just stick with simple lists for now [#390 (comment)](https://github.com/nexB/vulnerablecode/pull/390#issuecomment-806344755) . Don't do anything fancy.

Ok

Pushpit07:
Updated table

Screenshot 2021-04-05 at 13 32 02
Screenshot 2021-04-05 at 13 32 35
Screenshot 2021-04-05 at 13 33 29

Pushpit07:
@sbs2001 ping

sbs2001:
@Pushpit07  this is still using table.

Pushpit07:
It looks good though. And isn't causing any confusion now

Pushpit07:
Screenshot 2021-04-07 at 18 56 19

Pushpit07:
Removed table structure

Screenshot 2021-04-08 at 20 41 16

Pushpit07:
@sbs2001 ping

sbs2001:
@Pushpit07  This looks weird when there's no reference id. 

I think, we should make this simpler.

Let's keep the references as they are, not touch them. 

Instead we'll create another section called scores. 

This section would have a table(same like the references one) with 3 columns 
system | value | reference

the reference would have value equal to the parent url of the score.

Because really, going lists is making things complicated wrt css and presentation. I especially don't want to maintain the custom css.  

Again sorry for wasting your time, design is tough.

Pushpit07:
> This looks weird when there's no reference id.

Yes, it's hard to understand this way


> Instead we'll create another section called scores.
> 
> This section would have a table(same like the references one) with 3 columns
> system | value | reference
> 
> the reference would have value equal to the parent url of the score.

Can you explain this by an example?

sbs2001:
@Pushpit07  see https://jsfiddle.net/0rv6we7p/5/

sbs2001:
@Pushpit07  ping

Pushpit07:
oh, probably missed your comment. Will change it to how it is in the jsfiddle. Give me a day's time

Pushpit07:
@sbs2001 

Screenshot 2021-04-28 at 21 07 46

Screenshot 2021-04-28 at 21 07 36

vulnerabilities/templates/vulnerability.html
```diff
@@ -20,33 +20,72 @@ <h3>Summary:</h3>
...
+                    <th> System </th>
+                    <th> Score </th>
+                    <th> Found At </th>
+                </tr>
+            {% for ref in object_list %}
+                {% for obj in ref.scores %}
                 <tr>
-                {% if ref.reference_id %}
-                    <td>{{ref.reference_id}}</td>
+                    {% if ref.scores %}
```
sbs2001, line 61: IMHO ` if ref.scores ` is redundant since we are already iterating from it earlier.

vulnerabilities/templates/vulnerability.html
```diff
@@ -20,33 +20,72 @@ <h3>Summary:</h3>
...
+                        <td>{{obj.scoring_system}}</td>
 
-                {% else %}
-                    <td>-</td>
+                    {% else %}
```
sbs2001, line 64: Every score always has a scroing system, no need for this `else`

vulnerabilities/templates/vulnerability.html
```diff
@@ -20,33 +20,72 @@ <h3>Summary:</h3>
...
+                        <td>-</td>
 
-                {% endif %}
+                    {% endif %}
 
-                {% if ref.url %}
-                    <td><a href="{{ ref.url }}">{{ref.url}}</a></td>
+                    {% if ref.scores %}
```
sbs2001, line 69: This `if` is redundant as said before

vulnerabilities/templates/vulnerability.html
```diff
@@ -20,33 +20,72 @@ <h3>Summary:</h3>
...
+                        <td>{{obj.value}}</td>
 
-                {% else %}
-                    <td>-</td>
+                    {% else %}
```
sbs2001, line 72: Every score always has a `value`, this `else` is not required

vulnerabilities/templates/vulnerability.html
```diff
@@ -20,33 +20,72 @@ <h3>Summary:</h3>
...
 
-                {% endif %}
-                </tr>
+                    {% endif %}
+
+                    {% if ref.url %}
+                        <td><a href="{{ ref.url }}">{{ref.url}}</a></td>
+
+                    {% else %}
+                        <td>-</td>
```
sbs2001, line 81:
```python
  {% else %}
   <td>{{ ref.reference_id }}</td>
```

would be better

sbs2001 (changes requested):
Thanks @Pushpit07 , see my suggestions inline, and this will be good to merge then.

sbs2001:
@Pushpit07  This branch is stale, could you rebase this with latest `main`

sbs2001 (approved):
Thanks merging, this

Issue #338, opened 2021-02-07
Fix batched_advisories method 


sbs2001:
The current implementation of the `batched_advisories` at https://github.com/nexB/vulnerablecode/blob/6609bcfdebb4c28bd30dfd9d42011323d238ee4c/vulnerabilities/data_source.py#L195 needs to be simplified

1. Don't use sets, the hash method for the advisory dataclass is broken. 
2. Simplify the logic.

Eventually we want to move to non-batch proccessing so this low priority

PR #393, opened 2021-03-20
Collect Mozilla

### LINKED ISSUE #78: Collect Mozilla
See https://www.mozilla.org/en-US/security/known-vulnerabilities/

Hritik14:
Fixes #78. ~Most of~ the work is done. Here's the checklist
- [x] Parsing for mozilla data source
- [x] Working mozilla importer
- [x] Migrate to GitDataSource
- [x] Write test cases
- [x] More verbose comments
- [x] Update dependencies

Currently, it requires ~two~ extra dependencies: ~PyGithub~ and markdown
~I would eliminate the PyGithub dependency next but~ I think markdown has
to stay.
~Final dependencies would be comitted later.~

I would also really like advices on 
- The FIXMEs present in current code.
- Should we mark all the pks <= resolved pkg as vulnerable as mozilla doesn't ship that information

This is how it looks for now
image


vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
+from typing import Set, List, Generator
+
+import re
+import asyncio
+from bs4 import BeautifulSoup
```
sbs2001, line 5: Group your imports into 

1. Stdlib imports
2. 3rd party imports
3. project imports

Separate them by one line and sort them.

sbs2001 (changes requested):
@Hritik14  I've done a light review for now, will do another review once changes are made.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
...
+
+        advisories = []
+        for path in files:
+            for advisory in self._to_advisories(path):
+                advisories.append(advisory)
+                if len(advisories) >= self.batch_size:
+                    yield advisories
+                    advisories = []
+
+        # In case batch size is too high
```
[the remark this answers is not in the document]
Hritik14: Please have a look at the above lines
```py
        advisories = []
        for path in files:
            for advisory in self._to_advisories(path):
                advisories.append(advisory)
                if len(advisories) >= self.batch_size:
                    yield advisories
                    advisories = []
```
It only yields `if len(advisories) >= self.batch_size:` which would never be true in case the batch size is greater than the total number of advisories present. The loop would just complete and reach the end of the function. So, I've added a final yield.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        files = [
+            f for f in files if f.endswith(".md") or f.endswith(".yml")
+        ]  # skip irrelevant files
+
+        advisories = []
+        for path in files:
+            for advisory in self._to_advisories(path):
+                advisories.append(advisory)
+                if len(advisories) >= self.batch_size:
```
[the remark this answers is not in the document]
Hritik14: I'm unsure if I should just fetch all the advisories then pass them to `batch_advisories` effectively nullifying the benefit of python generators or should I keep generating them according to batch size and yield them.
I've implemented the later in this case.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
...
+                    yield advisories
+                    advisories = []
+
+        # In case batch size is too high
+        yield advisories
+
+    def _to_advisories(self, path: str) -> List[Advisory]:
```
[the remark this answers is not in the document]
Hritik14: tbh, the overuse made me add `_`. Initially they didn't have underscores. I'd go ahead and remove them.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
...
+                    vulnerability_id=cve if is_cve(cve) else "",
+                    impacted_package_urls=[],
+                    resolved_package_urls=fixed_package_urls,
+                    references=references,
+                )
+            )
+
+        return advisories
+
+    def _parse_md(self, mfsa_id, lines) -> List[Advisory]:
```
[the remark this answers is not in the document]
Hritik14: Ok, changing.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,208 @@
...
+        p = ""
+        if h3tag:
+            for tag in h3tag.next_siblings:
+                if tag.name:
+                    if tag.name != "p":
+                        break
+                    p += tag.get_text()
+        return p
+
+    def _parse_md_front_matter(self, lines):
```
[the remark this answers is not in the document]
Hritik14: I guess you meant https://github.com/nexB/vulnerablecode/blob/3d66b4e82ee31422ab907d3388f739b768ffd2ac/vulnerabilities/importers/istio.py#L86
Yes. It would make more sense as a helper. Should I name it `extract_markdown_front_matter` or let me know if you have something  better in mind.

vulnerabilities/importer_yielder.py
```diff
@@ -234,6 +234,16 @@
         "data_source": "IstioDataSource",
         "data_source_cfg": {"repository_url": "https://github.com/istio/istio.io"},
     },
+    {
+        "name": "mozilla",
+        "license": "mpl-2.0",
+        "last_run": None,
+        "data_source": "MozillaDataSource",
+        "data_source_cfg": {
+            "branch": None,
```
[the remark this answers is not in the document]
Hritik14: I actually got this from the rust importer. I guess I should modify both to just fall on defaults.

vulnerabilities/importers/mozilla.py
Hritik14: On it

Hritik14:
@sbs2001 any update ?

vulnerabilities/importers/mozilla.py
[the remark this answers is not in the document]
sbs2001: Use `batch_advisories` for now. It has it's own problems and there's https://github.com/nexB/vulnerablecode/issues/338 for that. The generator thing won't do much, since we are importing like 10-20 MBs of data.  

We'll be able to fix the  https://github.com/nexB/vulnerablecode/issues/338 with some minor changes quite easily, if every importer uses the same thing. I don't really want the batching logic repeated in multiple places.

sbs2001:
@Hritik14  are you sure, you pushed the changes ?

Hritik14:
@sbs2001 ah, not yet. I actually replied on the requested changes. I'm not sure if you get notifications for those replies. Do let me know and I'll leave a comment next time as well.

Hritik14:
I've refactored as suggested and also added a helper functions to split markdown and the front matter. I'd add it to the istio importer after this PR is merged.

Hritik14:
rebased

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+        return p
+
+    def mfsa_id_from_filename(self, filename):
+        match = MFSA_FILENAME_RE.search(filename)
+        if match:
+            return "mfsa" + match.group(1)
+
+        return None
+
+    def get_package_urls(self, pkgs: List[str]) -> List[PackageURL]:
```
sbs2001, line 129: What if pkgs is None ? You are using `data.get("fixed_in") above

sbs2001, line 129: This doesn't require the instance, why don't you make it a staticmethod ?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+        severities = ["critical", "low", "high", "medium", "none"]
+        severity = [{severity in data.get("impact"): severity} for severity in severities][0].get(
+            True
+        )
+
+        return [
+            Reference(
+                reference_id=data["mfsa_id"],
+                url="https://www.mozilla.org/en-US/security/advisories/{}".format(data["mfsa_id"]),
+                severities=[VulnerabilitySeverity(scoring_systems["cvssv3.1_qr"], severity)],
```
sbs2001, line 162: It's not a cvssv3.1_qr , create a system for these.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def get_yml_references(self, data: any) -> List[Reference]:
+        """
+        Returns a list of references
+        Currently only considers the given mfsa as a reference
+        """
+        # FIXME: Needs improvement
+        # Should we add 'bugs' section in references too?
+        # Should we add 'impact'/severity of CVE in references too?
+        # If yes, then fix alpine_linux importer as well
```
sbs2001, line 149: What's wrong at alpine linux ?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+        # Otherwise, do we need severity field for adversary as well?
+
+        # FIXME: Write a helper for cvssv3.1_qr severity detection ?
+        severities = ["critical", "low", "high", "medium", "none"]
+        severity = [{severity in data.get("impact"): severity} for severity in severities][0].get(
```
sbs2001, line 154: This is too much to unpack use simple `for` loop .

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+                        break
+                    p += tag.get_text()
+        return p
+
+    def mfsa_id_from_filename(self, filename):
+        match = MFSA_FILENAME_RE.search(filename)
+        if match:
+            return "mfsa" + match.group(1)
+
+        return None
```
sbs2001, line 127: Does this even happen ? If there's no code to handle that

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+            )
+        ]
+
+    def html_get_p_under_h3(self, html, h3: str):
+        soup = BeautifulSoup(html, features="lxml")
+        h3tag = soup.find("h3", text=lambda txt: txt.lower() == h3)
+        p = ""
+        if h3tag:
+            for tag in h3tag.next_siblings:
+                if tag.name:
```
sbs2001, line 116: Just curious, could you point me to an example where this is not true ?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,164 @@
...
+                    if tag.name != "p":
+                        break
+                    p += tag.get_text()
```
sbs2001, line 119: Would that collect the `strong` tag too ? Like the one at https://github.com/mozilla/foundation-security-advisories/blob/master/announce/2016/mfsa2016-14.md ? Not a big deal but we should avoid collecting those tags.

sbs2001 (changes requested):
Thanks @Hritik14  I've left some comments inline for your consideration. This looks good btw .

vulnerabilities/importers/mozilla.py
Hritik14: That comment is about should we add the severity of the CVE itself ? Severity is only available in references, so maybe we should consider adding the CVE Mitre as a reference (as done in this PR). 
In Alpine, the CVE is not added as a reference but only as id. It won't really be required there as no severity is mentioned in the advisory, I wrote it just to maintain uniformaty. 
This is what I was thinking

vulnerabilities/importers/mozilla.py
Hritik14: As per the [given format[(https://github.com/mozilla/foundation-security-advisories#markdown-format) it should never happen.
Just in case it happens, I've now added to skip that file.

vulnerabilities/importers/mozilla.py
Hritik14: Done. See #415

vulnerabilities/importers/mozilla.py
Hritik14: No.
For the linked document it returns
```
Security researcher Holger Fuhrmannek reported that a malicious\nGraphite....
```
Only consecutive `p` tags are collected and [get_text()](https://beautiful-soup-4.readthedocs.io/en/latest/#get-text) makes sure we're getting only textual representation of the content present in `p` tags.

Hritik14:
@sbs2001 any update ?

vulnerabilities/helpers.py
```diff
@@ -79,6 +80,41 @@ def create_etag(data_src, url, etag_key):
...
+    ['title: ISTIO-SECURITY-2019-001','description: Incorrect access control.'
+    ,'cves: [CVE-2019-12243]'],
+    ["# Markdown starts here"]
+    """
+
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
+
+    for index, line in enumerate(lines.split("\n")):
```
pombredanne, line 107: What about this instead?
```Python
import saneyaml

# normalize line endings just in case:
text = text.replace("\r\n", "\n")
front_matter, _, body = text.rpartition("\n---\n")
front_matter = saneyaml.load(front_matter)
```

For instance:
```
>>> import saneyaml
>>> text = """---
... title: ISTIO-SECURITY-2019-001
... subtitle: 安全公告
... description: 错误的权限控制。
... cve: [CVE-2019-12243]
... publishdate: 2019-05-28
... keywords: [CVE]
... skip_seealso: true
... aliases:
...     - /zh/blog/2019/cve-2019-12243
...     - /zh/news/2019/cve-2019-12243
... ---
... 
... {{< security_bulletin
...         cves="CVE-2019-12243"
...         cvss="8.9"
...         vector="CVSS:3.0/AV:A/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:N/E:H/RL:O/RC:C"
...         releases="1.1 to 1.1.6" >}}
... 
... ## 内容{#context}
... """
>>> text = text.replace("\r\n", "\n")
>>> front_matter, _, body = text.rpartition("\n---\n")
>>> front_matter = saneyaml.load(front_matter)
>>> print(front_matter)
{'title': 'ISTIO-SECURITY-2019-001', 'subtitle': '安全公告', 'description': '错误的权限控制。', 'cve': ['CVE-2019-12243'], 'publishda
te': '2019-05-28', 'keywords': ['CVE'], 'skip_seealso': True, 'aliases': ['/zh/blog/2019/cve-2019-12243', '/zh/news/2019/cve-2019-122
43']}

```

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def get_advisories_from_md(mfsa_id, lines) -> List[Advisory]:
+    yamltext, mdtext = split_markdown_front_matter(lines.read())
+    data = yaml.safe_load(yamltext)
+    data["mfsa_id"] = mfsa_id
+
+    fixed_package_urls = get_package_urls(data.get("fixed_in"))
+    references = get_yml_references(data)
+    cves = re.findall(r"CVE-\d+-\d+", yamltext + mdtext, re.IGNORECASE)
```
pombredanne, line 97: I would prefer to compile this ahead and give it a nice callable name

pombredanne, line 97: I would prefer to compile this ahead and give it a nice callable name

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+                p += tag.get_text()
+    return p
+
+
+def mfsa_id_from_filename(filename):
+    match = MFSA_FILENAME_RE.search(filename)
+    if match:
+        return "mfsa" + match.group(1)
+
+    return None
```
pombredanne, line 137: This is not needed, as this is the default

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+        for pkg in pkgs
+        if pkg
+    ]
+    return package_urls
+
+
+def get_yml_references(data: any) -> List[Reference]:
+    """
+    Returns a list of references
```
pombredanne, line 156: Please use `Return a list of references.` e.g. the imperative style.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+    Currently only considers the given mfsa as a reference
+    """
+    # FIXME: Needs improvement
+    # Should we add 'bugs' section in references too?
```
pombredanne, line 160: Yes

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+    # Should we add 'impact'/severity of CVE in references too?
+    # If yes, then fix alpine_linux importer as well
```
pombredanne, line 162: Please enter a ticket for that instead.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def get_yml_references(data: any) -> List[Reference]:
+    """
+    Returns a list of references
+    Currently only considers the given mfsa as a reference
+    """
+    # FIXME: Needs improvement
+    # Should we add 'bugs' section in references too?
+    # Should we add 'impact'/severity of CVE in references too?
```
pombredanne, line 161: In general yes.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+    # If yes, then fix alpine_linux importer as well
+    # Otherwise, do we need severity field for adversary as well?
+
+    severities = ["critical", "high", "medium", "low", "none"]
+    severity = "none"
+    if data.get("impact"):
```
pombredanne, line 167: Avoid "getting" multiple times, e.g. create a variable instead:
`impact = data.get("impact")`
and use it after

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+        impact = data.get("impact").lower()
+        for s in severities:
```
pombredanne, line 169: I am not sure I get why you do things this way. What would be data examples?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+            if s in impact:
+                severity = s
+                break
+
+    return [
+        Reference(
+            reference_id=data["mfsa_id"],
```
pombredanne, line 176: Create a variable for `data["mfsa_id"]` and reuse it. Do not get multiple times.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+            url="https://www.mozilla.org/en-US/security/advisories/{}".format(data["mfsa_id"]),
```
pombredanne, line 177: Since you would have a var, you can use an f string instead now

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
+import re
+from typing import List
+from typing import Set
+
+import yaml
```
pombredanne, line 5: Please use saneyaml instead which has some secure defaults (and that we maintain)

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+
+    return None
+
+
+def get_package_urls(pkgs: List[str]) -> List[PackageURL]:
+    package_urls = [
+        PackageURL(
+            type="mozilla",
+            # pkg is of the form "Firefox ESR 1.21" or "Thunderbird 2.21"
+            name=pkg.rsplit(None, 1)[0],
```
pombredanne, line 145: this split and the one below are rather cryptic, magic and likely error prone, please use a plain loop rather than a list comprehension and try to find a more readable and explicit way to do things. partition and rpartition are usually more robust and easier to grok.

Also avoid doing the multiple time the same op (here for the name and version)

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+            summary=description,
+            vulnerability_id="",
+            impacted_package_urls=[],
+            resolved_package_urls=fixed_package_urls,
+            references=references,
+        )
+    ]
+
+
+def html_get_p_under_h3(html, h3: str):
```
pombredanne, line 119: Find a better name and add a docstring. The name should be descriptive about what it returns or does, not how it does it. Here `html_get_p_under_h3` tells me nothing at all.
Think of it this way: 3 months from now, you are reading this code again. the function and docstring should make this obvious and self-evident.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+    if match:
+        return "mfsa" + match.group(1)
+
+    return None
+
+
+def get_package_urls(pkgs: List[str]) -> List[PackageURL]:
+    package_urls = [
+        PackageURL(
+            type="mozilla",
```
pombredanne, line 143: Are we sure we want to create a new type for this?
How are the CPEs at the NVD?

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+
+@patch("vulnerabilities.importers.MozillaDataSource._update_from_remote")
+class MozillaImportTest(TestCase):
+
+    tempdir = None
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.tempdir = tempfile.mkdtemp()
+        zip_path = os.path.join(TEST_DATA, "mozilla.zip")
```
pombredanne, line 25: Please avoid zip fixtures unless really needed, e.g. extremely rarely if ever

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+            data_source_cfg={
+                "repository_url": "https://example.git",
+                "working_directory": os.path.join(cls.tempdir, "mozilla_test"),
+                "create_working_directory": False,
+                "remove_working_directory": False,
+            },
+        )
+
+    @classmethod
+    def tearDownClass(cls) -> None:
```
pombredanne, line 44: Without zip, this is not needed

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.tempdir = tempfile.mkdtemp()
+        zip_path = os.path.join(TEST_DATA, "mozilla.zip")
+
+        with zipfile.ZipFile(zip_path, "r") as zip_ref:
+            zip_ref.extractall(cls.tempdir)
+
+        cls.importer = models.Importer.objects.create(
```
pombredanne, line 30: Why use a setup()? move this down to the one test that uses it instead

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        # Make sure no requests for unexpected package names have been made during the tests.
+        shutil.rmtree(cls.tempdir)
+
+    def test_import(self, _):
+        runner = ImportRunner(self.importer, 100)
+
+        # Remove if we don't need set_api in MozillaDataSource
```
pombredanne, line 51: Do not keep dead code around

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+        self.assert_for_package("Mozilla Suite", "mfsa2005-29", "1.7.6")
+
+    def assert_for_package(
+        self,
+        package_name,
+        mfsa_id,
+        resolved_version,
+        vulnerability_id=None,
+        impacted_version=None,
+    ):
```
pombredanne, line 78: Can you add a docstring? I cannot get exactly what this is doing otherwise.

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+        runner.run()
+
+        assert models.Vulnerability.objects.count() == 9
+        assert models.VulnerabilityReference.objects.count() == 10
+        assert models.VulnerabilitySeverity.objects.count() == 9
+        assert models.PackageRelatedVulnerability.objects.filter(is_vulnerable=False).count() == 16
+
+        assert models.Package.objects.count() == 12
+
+        self.assert_for_package("Firefox ESR", "mfsa2021-06", "78.7.1")
```
pombredanne, line 64: please used named keyword args for clarity and readability

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+        assert models.VulnerabilityReference.objects.get(url=ref_url, vulnerability=vuln)
+
+        assert models.PackageRelatedVulnerability.objects.filter(
+            package=pkg, vulnerability=vuln, is_vulnerable=False
+        )
+
+
+def test_categorize_versions_ranges():
+    # Populate if impacted version is filled
+    pass
```
pombredanne, line 97: I assume this is upcoming in the same PR? otherwise, please remove dead code.

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
...
+        assert models.Package.objects.count() == 12
+
+        self.assert_for_package("Firefox ESR", "mfsa2021-06", "78.7.1")
+        self.assert_for_package("Firefox ESR", "mfsa2021-04", "78.7", "CVE-2021-23953")
+        self.assert_for_package("Firefox for Android", "mfsa2021-01", "84.1.3", "CVE-2020-16044")
+        self.assert_for_package("Thunderbird", "mfsa2014-30", "24.4")
+        self.assert_for_package("Thunderbird", "mfsa2014-30", "24.4")
+        self.assert_for_package("Mozilla Suite", "mfsa2005-29", "1.7.6")
+
+    def assert_for_package(
```
pombredanne, line 71: What could be a better name?

vulnerabilities/tests/test_mozilla.py
```diff
@@ -0,0 +1,97 @@
+import os
```
pombredanne, line 1: Could you add add some true unit tests too for each of your functions?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+                url=f"https://cve.mitre.org/cgi-bin/cvename.cgi?name={cve}",
+            )
+        )
+
+    description = html_get_p_under_h3(markdown(mdtext), "description")
+
+    return [
+        Advisory(
+            summary=description,
+            vulnerability_id="",
```
pombredanne, line 111: Is this needed?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+            impacted_package_urls=[],
```
pombredanne, line 112: How can you get no impacted packages and fixed packages?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+                references=references,
+            )
+        )
+
+    return advisories
+
+
+def get_advisories_from_md(mfsa_id, lines) -> List[Advisory]:
+    yamltext, mdtext = split_markdown_front_matter(lines.read())
+    data = yaml.safe_load(yamltext)
```
pombredanne, line 92: Consider using saneyaml instead?

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+    with open(path) as lines:
+        if path.endswith(".md"):
+            return get_advisories_from_md(mfsa_id, lines)
+        if path.endswith(".yml"):
+            return get_advisories_from_yml(mfsa_id, lines)
+
+    return []
+
+
+def get_advisories_from_yml(mfsa_id, lines) -> List[Advisory]:
```
pombredanne, line 62: That part seems to be the shared when using a YAML front matter. IMHO there is duplicated code here between your two functions.

vulnerabilities/importers/mozilla.py
```diff
@@ -0,0 +1,180 @@
...
+from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import GitDataSource
+from vulnerabilities.data_source import Reference
+from vulnerabilities.data_source import VulnerabilitySeverity
+from vulnerabilities.helpers import is_cve
+from vulnerabilities.helpers import split_markdown_front_matter
+from vulnerabilities.severity_systems import scoring_systems
+
+REPOSITORY = "mozilla/foundation-security-advisories"
+MFSA_FILENAME_RE = re.compile(r"mfsa(\d{4}-\d{2,3})\.(md|yml)$")
```
pombredanne, line 19: IMHO this should be made as callable closing on .search or .match with an explicit name. And should use named groups for clarity.

pombredanne (changes requested):
Thanks!
see a few comments inline for you consideration.

Also, can you avoid having a test zip in vulnerabilities/tests/test_data/mozilla.zip  ? text is fine instead.
This will simplify the tests and help review too.

vulnerabilities/helpers.py
Hritik14: We have not been using `saneyaml` anywhere in the project. If that is preferred, I can add it to requirements and we can start using that but I don't see any specific reason to do so here.
Regarding `rpartition`. We cannot have that either as a markdown is allowed to have a `"\n---"\n`. 
Consider this 
```py
>>> text=""""---
... announced: February 4, 2014
... fixed_in:
... - Firefox 27
... - Firefox ESR 24.3
... - Thunderbird 24.3
... - Seamonkey 2.24
... impact: High
... reporter: Cody Crews
... title: Clone protected content with XBL scopes
... ---
... 
... <h3>Description</h3>
... 
... Text
... ---
... other text
... """
>>> text = text.replace("\r\n", "\n")
>>> front_matter, _, body = text.rpartition("\n---\n")
>>> 
>>> front_matter
'"---\nannounced: February 4, 2014\nfixed_in:\n- Firefox 27\n- Firefox ESR 24.3\n- Thunderbird 24.3\n- Seamonkey 2.24\nimpact: High\nreporter: Cody Crews\ntitle: Clone protected content with XBL scopes\n---\n\n<h3>Description</h3>\n\nText'
>>> 
>>> body
'other text\n'
>>> 
```
 The official validator by mozilla parses it something like this 
https://github.com/mozilla/foundation-security-advisories/blob/d43d09d204ab5da014e83b7d1743df289cefee92/check_advisories.py#L183-L208

Further, as it is a helper, we cannot have it mozilla specific. All it should do is to split the front matter and markdown.
I could have something like this
```py
import yaml

text="""---
announced: February 4, 2014
fixed_in:
- Firefox 27
- Firefox ESR 24.3
- Thunderbird 24.3
- Seamonkey 2.24
impact: High
reporter: Cody Crews
title: Clone protected content with XBL scopes
---

<h3>Description</h3>
---
other text
"""
# normalize line endings just in case:
text = text.replace("\r\n", "\n")
linezero,_, text = text.partition("---\n")
if not linezero: # nothing before first ---
    front_matter,_, body = text.partition("---")
    front_matter = yaml.safe_load(front_matter)
else:
    front_matter = ""
    body = linezero + "---\n" + text
print(front_matter)
print(body)
```
which prints
```
{'announced': 'February 4, 2014', 'fixed_in': ['Firefox 27', 'Firefox ESR 24.3', 'Thunderbird 24.3', 'Seamonkey 2.24'], 'impact': 'High', 'reporter': 'Cody Crews', 'title': 'Clone protected content with XBL scopes'}


<h3>Description</h3>
---
other text
```
but doesn't look any better to me.

vulnerabilities/importers/mozilla.py
Hritik14: According to [python docs](https://docs.python.org/3/library/re.html?highlight=re#re.compile)
> Note

The compiled versions of the most recent patterns passed to re.compile() and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions. 

and we will be using this regex only until the loop is over, so it should suffice like this anyway.

If you'd still like to have it compiled, I'd like to open a PR to change the helpers.py to include a standard `cve_regex` that would be used by all the importers. Something like
```
cve_regex = re.compile(r"CVE-\d{4}-\d{4,7}", re.IGNORECASE)
is_cve = cve_regex.is_cve
```
then I use `cve_regex.find_all` above.

Hritik14: same as above

vulnerabilities/importers/mozilla.py
Hritik14: Fixed

vulnerabilities/importers/mozilla.py
Hritik14: Fixed

vulnerabilities/importers/mozilla.py
Hritik14: It was a personal reminder. Removed.

vulnerabilities/importers/mozilla.py
Hritik14: fixed

vulnerabilities/importers/mozilla.py
Hritik14: These are all unique impacts
```
mfsa2020-25:  high
mfsa2013-50:  Critical
mfsa2009-19:  High
mfsa2017-17:  critical
mfsa2009-15:  Low
mfsa2015-75:  Moderate
mfsa2020-07:  moderate
mfsa2007-06:  Critical (Firefox 2.0 not affected in default configuration)
mfsa2015-21:  Medium
mfsa2005-02:  Moderate (on a multiuser computer)
mfsa2005-28:  Critical (local)
mfsa2005-48:  Low (High for Mozilla Suite)
mfsa2005-59:  Severe
mfsa2010-55:  Low (Critical in Gecko 1.9.1 and earlier)
mfsa2005-10:  Moderate to Critical
```
I did that to make look like cvssv3.1_qr but now as we have generic_textual (https://github.com/nexB/vulnerablecode/pull/415) as well, perhaps I can just put the entire impact line over there.

vulnerabilities/importers/mozilla.py
Hritik14: I added the `mfsa_id` in `data` to make sure everything we need about the advisory is at one place, inside `data`. 
https://github.com/nexB/vulnerablecode/blob/5997b26ba35d3cff83c07ee6f6fd2c81b4f526ae/vulnerabilities/importers/mozilla.py#L65

Is there something wrong with this approach?

vulnerabilities/importers/mozilla.py
Hritik14: Fixed

vulnerabilities/importers/mozilla.py
Hritik14: Would we need to update all the importers to reflect this ?

vulnerabilities/importers/mozilla.py
Hritik14: I've updated it to use a for loop. I've used rsplit to be consistent with the upstream. 
Here's what the upstream uses: https://github.com/mozilla/bedrock/blob/bd0f50a1c0a7455115cc2bf9eb3ae9027586a024/bedrock/security/models.py#L28-L30
It is mentioned in the commit message here https://github.com/nexB/vulnerablecode/pull/393/commits/05d93b2e3a88b15bceb5675b7cb46569d3fc216b

vulnerabilities/importers/mozilla.py
Hritik14: Updated
```py
def html_get_section(html, heading, heading_tag="h3"):
    """
    Return the contents of all <p> tags below the provided header (by default h3 tag)
    eg:
    <h3>Description</h3>
    <p> Here goes the desc.</p>
    <p> More desc.</p>

    html_get_section("description") would return "Here goes the desc.More desc"
    """
    soup = BeautifulSoup(html, features="lxml")
    h3tag = soup.find(heading_tag, text=lambda txt: txt.lower() == heading.lower())
```

Does it look better?

vulnerabilities/tests/test_mozilla.py
Hritik14: This was inspired by npm and rust test cases. Removing it now.

vulnerabilities/tests/test_mozilla.py
Hritik14: Removed

vulnerabilities/tests/test_mozilla.py
Hritik14: Again, inspired by npm importer. Removing zip entirely.

vulnerabilities/tests/test_mozilla.py
Hritik14: Removed

vulnerabilities/tests/test_mozilla.py
Hritik14: A similar approach was taken here https://github.com/nexB/vulnerablecode/blob/f254b0d4ac54b70c648055a7e8eda16c05dce0f9/vulnerabilities/tests/test_npm.py#L104-L130

Replacing zip with tests for each function would remove this as well.

vulnerabilities/tests/test_mozilla.py
Hritik14: impacted versions are not provided by mozilla. This was for future PR. Removed.

vulnerabilities/tests/test_mozilla.py
Hritik14: Removed

vulnerabilities/tests/test_mozilla.py
Hritik14: OK. Replacing these with true unit tests

vulnerabilities/importers/mozilla.py
Hritik14: Removed

vulnerabilities/importers/mozilla.py
Hritik14: Upstream doesn't provide with a list of impacted package. We might consider all packages until this package as impacted but I'm not sure about this. There are many other importers that do this too. Eg: https://github.com/nexB/vulnerablecode/blob/f254b0d4ac54b70c648055a7e8eda16c05dce0f9/vulnerabilities/importers/alpine_linux.py#L194

vulnerabilities/importers/mozilla.py
Hritik14: Same as https://github.com/nexB/vulnerablecode/pull/393#discussion_r607036120

vulnerabilities/importers/mozilla.py
Hritik14: format for yaml and md differ for some fields, like description. I tried combining them but it didn't look any better to me. Do you have any suggestion ?

vulnerabilities/importers/mozilla.py
Hritik14: Is this better ? 
``search_mfsa_in_filename = re.compile(r"mfsa(\d{4}-\d{2,3})\.(md|yml)$").search``
There is no grouping, only mfsa id is returned so I've avoided named groups.

Hritik14:
@pombredanne Thank you for the through review. Please check unresolved conversations and I'll push the updated code.

vulnerabilities/importers/mozilla.py
Hritik14: The CVEs are already added as `vulnerablity_id`, do you mean we should repeat them as a Reference as well ?
The only reason I can see to repeat the CVE as a reference is to mention it's `impact` as stated by mozilla. That would go under `generic_textual`.

vulnerabilities/importers/mozilla.py
Hritik14: Added

vulnerabilities/importers/mozilla.py
Hritik14: Just noticed that entries like `Critical (Firefox 2.0 not affected in default configuration)` exceed maximum value length.

vulnerabilities/importers/mozilla.py
Hritik14: Currently, I'm adding the CVE as a reference if it "refers" to _another_ CVE.

vulnerabilities/helpers.py
pombredanne: > but doesn't look any better to me.

It feels a bit more readable may be? 

> The official validator by mozilla parses it something like this
https://github.com/mozilla/foundation-security-advisories/blob/d43d09d204ab5da014e83b7d1743df289cefee92/check_advisories.py#L183-L208

We could very much reuse it as-is as well. This is under an MPL license so we would have to do it in an orderly fashion with proper license tracking and code separation though. 
If we do not copy it, we cannot reuse code from it though.

vulnerabilities/importers/mozilla.py
pombredanne: ok, enter a ticket separately then and let's do that separately.

vulnerabilities/importers/mozilla.py
pombredanne: :+1:

vulnerabilities/importers/mozilla.py
pombredanne: nope, thank for the explanation.

vulnerabilities/importers/mozilla.py
pombredanne: In practice there are only 512 regex that can be cached https://github.com/python/cpython/blob/master/Lib/re.py#L264 which is why the practice is to compile in a variable 
And yes a separate PR for the CVE regex is welcomed too, but regex is a tool, so I would rather expose a function than a regex.

vulnerabilities/importers/mozilla.py
pombredanne: good point. We need a ticket though, as this may need to be interpreted as "all versions before this version are vulnerable" and it warrant some research and discussions.

vulnerabilities/importers/mozilla.py
pombredanne: Better but why having `heading_tag="h3"` as an argument?
Also if you collect paragraphs, then you should join them with a line return and glued not as in `"Here goes the desc.More desc"`

While you are at it, using a doctest in the docstring works too since this is mostly what you are doing already here

pombredanne: Also since this is only used to get a description, why not call this `get_description(html_text)` and hide "how the sausage is made" inside the function?

vulnerabilities/importers/mozilla.py
pombredanne: yes, the fact the CVE is an id is one thing. It still needs to be added as a reference too e.g. a CVE is nothing special beyond the fact we use this as a unique identifier when available.

vulnerabilities/importers/mozilla.py
pombredanne: rpartition is a more robust splitter than rsplit. Your rsplit will fail when there is no spaces 
```
>>> a, b = 'foo'.rsplit(None, 1)
ValueError: not enough values to unpack (expected 2, got 1)
>>> a, _, b = 'foo'.rpartition(' ')

```

vulnerabilities/importers/mozilla.py
pombredanne: https://nvd.nist.gov/vuln/detail/CVE-2021-23981 as a CPE with mozilla as a vendor, so it may make sense to use that as a type.
How much can you infer from a "mozilla" Package URL?
Can you find mechanically a download URL for the package? a homepage? and so on?

vulnerabilities/importers/mozilla.py
pombredanne: This is at best an artistic scale.... I would like to see what it looks like for recent values, say no less than 5 years old as this will weed out old things like FF 2.0
For the funky values I would rather have an explicit hardcoded mapping....and since they document it, we should IMHO create a Mozilla scoring system
@sbs2001 what's your take?

vulnerabilities/importers/mozilla.py
pombredanne: Why pass down a mapping when you can pass named arguments? 
It is much better to give names.
Something such as:
```Python
def get_reference(mfsa_id, impact):
    """
    Return a reference for tan mfsa and is impact value.
    """
    known_impacts = set(["critical", "high", "medium", "low", "none"])

    impact = impact or ''
    impact = impact.lower()
    severity = known_impacts.get(impact)

    severities=[]
    if severity:
        severities = [VulnerabilitySeverity(scoring_systems["generic_textual"], severity)]

    return Reference(
        reference_id=mfsa_id,
        url=f"https://www.mozilla.org/en-US/security/advisories/{mfsa_id}",
        severities=severities,
    )

```

Also do not return a list when you return a single value. I appreciate planning for the future, but that's usually not needed. We can always refactor later if needed.


And add other impacts that may be funky to `known_impacts` that you can turn in a mapping.
I would also raise an exception or some warning if there is an impact that is not known and is not in the mapping.

vulnerabilities/tests/test_mozilla.py
pombredanne: I meant add unit tests, not replace these entirely. Both are useful

vulnerabilities/tests/test_mozilla.py
pombredanne: If we have zip in npm and rust we need to replace them too then... Do you mind to create a ticket for this>

pombredanne:
@Hritik14 done :) Thank you for having looked at all these in details!

vulnerabilities/importers/mozilla.py
Hritik14: opened https://github.com/nexB/vulnerablecode/pull/439

vulnerabilities/importers/mozilla.py
Hritik14: Understood and implemented :)

vulnerabilities/importers/mozilla.py
Hritik14: Thanks.
I've modified it to look like this
```py
def get_yml_references(mfsa_id, impact, advisories) -> List[Reference]:
    """
    Return a list of references
    Considers the following as references:
        - mfsa
        - cve
        - bugzilla or direct reference
    """

    known_impacts = { 
            "severe": "critical",
            "critical": "critical",
            "critical (local)": "critical",
            "high": "high",
            "moderate": "moderate",
            "medium": "moderate",
            "moderate (on a multiuser computer)": "moderate",
            "low": "low",
            "critical (firefox 2.0 not affected in default configuration)": "critical",
            "low (critical in gecko 1.9.1 and earlier)": "low",
            "moderate to critical": "moderate",
            }


    impact = impact or ''
    impact = impact.lower()
    severity = known_impacts.get(impact)
    if severity:
        severities=[VulnerabilitySeverity(scoring_systems["generic_textual"], severity)]
    else:
        warnings.warn(f'{mfsa_id}: unknown severity "{severity}"')

    # Add mfsa as reference
    references = [
            Reference(
                reference_id=data["mfsa_id"],
                url=f"https://www.mozilla.org/en-US/security/advisories/{data['mfsa_id']}",
                severities=severities,
                )
            ]

    advisories = data.get("advisories", {})
    for advisory in advisories:
        # Add advisory cve as reference
    ...
    ...
```
I would be needing the references list as now I've also added the bugs and cve itself as a reference.

vulnerabilities/importers/mozilla.py
Hritik14: Done. Ticket here https://github.com/nexB/vulnerablecode/issues/440

vulnerabilities/importers/mozilla.py
Hritik14: If there's no space, `rpartition` puts the entire string on `b` which would be `version` here. imo, if `pkg="Mozilla"` it refers to the `name` and not the `version`.
To account for that, I came up with this. Not sure if it looks hacky.
```py
            name, _, version = pkg.rpartition(' ')
            if not name:
                name = version
```

vulnerabilities/importers/mozilla.py
Hritik14: Done

vulnerabilities/tests/test_mozilla.py
Hritik14: Created https://github.com/nexB/vulnerablecode/issues/442

vulnerabilities/helpers.py
Hritik14: As it is required by multiple importers, let's move this to it's own PR. Opened https://github.com/nexB/vulnerablecode/pull/443

vulnerabilities/importers/mozilla.py
Hritik14: @pombredanne Do you mean to use the following as the types or just `nvd`? 
 -   cpe:2.3:a:mozilla:firefox:*:*:*:*:*:*:*:*
 -   cpe:2.3:a:mozilla:thunderbird:*:*:*:*:*:*:*:*

I'm still unsure about how this will help. 
By `mozilla` as a type, I intend to point to https://hg.mozilla.org/.

vulnerabilities/importers/mozilla.py
sbs2001: Ouch, is this going to be redundant now, just like the suse_backport importer. We now have a improved notion of fixed/resolved/patched package now.  Due to https://github.com/nexB/vulnerablecode/pull/436

sbs2001:
@pombredanne  re zip files: see https://github.com/nexB/vulnerablecode/issues/442#issuecomment-822368195 . 

@Hritik14  IMHO mocking the whole GItDataSource's logic of collecting changes is redundant, since there are thorough tests of that already. There is some mock db things which IMHO is also not required . Instead you could just test the `to_advisories` and see whether correct `Advisory` objects are returned.  It's fast (no need to setup DB and such) and easy to implement.

Something like https://github.com/nexB/vulnerablecode/blob/main/vulnerabilities/tests/test_retiredotnet.py . There are tests which do check what ends up in db, but I think those should be limited and ideally should write small amounts of data.

vulnerabilities/importers/mozilla.py
Hritik14: @sbs2001 That is scary. I've opened https://github.com/nexB/vulnerablecode/issues/449 to discuss this further.

pombredanne:
@Hritik14 do you mind to rebase?

vulnerabilities/importers/mozilla.py
pombredanne: I would like to merge this sooner than later. ... See also https://github.com/nexB/vulnerablecode/issues/449#issuecomment-878336583

vulnerabilities/importers/mozilla.py
pombredanne: > By mozilla as a type, I intend to point to https://hg.mozilla.org/.

That would be part of what the `mozilla` would refer to. I would not use `nvd` ever as a package type.
So we are cool here. Marking this resolved.

vulnerabilities/importers/mozilla.py
pombredanne: this works fine. Or you could do this slightly more explicit:
```
>>> def name_ver(s):
...     if ' ' in s:
...         name, _, ver = s.rpartition(' ')
...     else:
...         name = s
...         ver = None
...     return name, ver
... 
>>> name_ver('foo')
('foo', None)
>>> name_ver('foo 23.34')
('foo', '23.34')
```

vulnerabilities/importers/mozilla.py
pombredanne: Let's create a scoring system for mozilla using these values:
```
critical
severe
high
medium
moderate
low
``` 
in https://github.com/nexB/vulnerablecode/blob/main/vulnerabilities/severity_systems.py

vulnerabilities/importers/mozilla.py
pombredanne: Have you pushed these updates?

vulnerabilities/tests/test_mozilla.py
pombredanne: @Hritik14 npm and rust are for git ... yours are just plain text, so please do not zip here.

pombredanne (approved):
Let's merge this first and fix it later.
Please enter an issue to track that

PR #400, opened 2021-03-25
Add documentation of Bulk Search API

AmitGupta7580:
image
image

**Response :** 

`{
  "result": [
    {
      "url": "string",
      "unresolved_vulnerabilities": [
        {
          "url": "string",
          "vulnerability_id": "string",
          "references": [
            {
              "source": "string",
              "reference_id": "string",
              "url": "string",
              "scores": [
                {
                  "value": "string",
                  "scoring_system": "cvssv2"
                }
              ]
            }
          ]
        }
      ],
      "resolved_vulnerabilities": [
        {
          "url": "string",
          "vulnerability_id": "string",
          "references": [
            {
              "source": "string",
              "reference_id": "string",
              "url": "string",
              "scores": [
                {
                  "value": "string",
                  "scoring_system": "cvssv2"
                }
              ]
            }
          ]
        }
      ],
      "purl": "string",
      "type": "string",
      "namespace": "string",
      "name": "string",
      "version": "string",
      "subpath": "string",
      "qualifiers": {}
    }
  ]
}`

**POSTMAN**

image


sbs2001 (changes requested):
@AmitGupta7580  This is much much better.  However could you not nest the response inside `results` ? 

And you need to fix the DCO. Write decent commit messages, `Update abcd` is not very helpful.

AmitGupta7580:
> However could you not nest the response inside `results` ?

@sbs2001 I have tried lots of things. But unable to remove that nesting. Actually some name has to be given here.

 class PackageBulkResponseSerializer(serializers.Serializer):
 &nbsp;&nbsp;&nbsp;&nbsp; **result** = serializers.ListField(child=PackageSerializer())

Is there any problem with this representation ?


> Write decent commit messages, `Update abcd` is not very helpful.

Actually I am just want to pass all these checks thats why I forget to name commits properly. This will not happen in future.

sbs2001:
This is still nesting the results.

vulnerabilities/api.py
```diff
@@ -168,8 +170,8 @@ def bulk_search(self, request):
                 purl_response["resolved_vulnerabilities"] = []
                 purl_response["purl"] = purl_string
             response.append(purl_response)
-
-        return Response(response)
+        res = {"result": response}
```
sbs2001, line 173: This is still nested.

sbs2001 (changes requested):
we don't want to nest the results

vulnerabilities/api.py
AmitGupta7580: > This is still nesting the results.

As mentioned here https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema
to listing object we can use `Serializer(many=True)`

`@extend_schema(request=PackageBulkRequestSerializer, responses=PackageSerializer(many=True))`

When I am trying to return a list of Packages then it automatically convert it into paginator view.

Output : 

image

And I also tried to remove pagination from `setting.py` file and add a custom paginator in other API's but then it adds unwanted parameters in bulk_api as it fetches directly from the Package Model.

Preview :

image

Still trying some other options..

pombredanne:
@AmitGupta7580 gentle ping... do you think you can finish this?

AmitGupta7580:
Hello @pombredanne, I have tried lots of things on this but unable to remove that nesting and also suggested some other possible alternatives to it. I think I will not be able to do this task in the current version of code.

pombredanne:
In the end this is more of an API structure change than a documentation issue..... this is still valid and need to be visited in light of what we want as API principles... See also https://github.com/nexB/vulnerablecode/issues/454 as DRF spectacular should be retired IMHO.

So I am closing this now. See https://github.com/nexB/vulnerablecode/issues/511 for a follow up.

@AmitGupta7580 Thank you ++ for all your efforts here even if this is not merged... this ends up being a very valuable conversation.

tfranzel:
@pombredanne i tried to explain to you that you are not forced to use a CDN if you are concerned about security [(comment)](https://github.com/tfranzel/drf-spectacular/issues/389#issuecomment-843384777) . You are free to host any dist package of SwaggerUI on your side and then point to that with the `SWAGGER_UI_DIST` and `SWAGGER_UI_FAVICON_HREF` URL settings. It is simply bad form (for a variety of reasons) to copy&paste an upstream project into our code base. The fact that you did not rebut my statements made me assume you agree.

@AmitGupta7580 the fix for your problem is pretty easy actually. the `pagination_class` is inherited from the view to `bulk_search`. you just need to turn off the pagination on the action like so: 

```python
    @extend_schema(request=PackageBulkRequestSerializer, responses=PackageBulkResponseSerializer)
    @action(detail=False, methods=["post"], pagination_class=None)
    def bulk_search(self, request):
```

AmitGupta7580:
image
image

Thank You @tfranzel It workes for me. But I am facing a problem regarding Parameters, Can you guide me how can I remove the default parameters of a Serializer.

image

tfranzel:
@AmitGupta7580 the same mechanics apply also to`django-filter`. so you need to clear the corresponding (derived) classes that invoke those mechanics (tfranzel/drf-spectacular#399). 

```python
@action(detail=False, methods=["post"], pagination_class=None, filter_backends=[])
...
```

AmitGupta7580:
Thank you very much for your help @tfranzel. @pombredanne, @sbs2001 take a look on it. should I make a new PR?

image

image

PR #404, opened 2021-03-26
enable configuration of allowed host

tardyp:
when deploying this outside of heroku, we get strange 400 errors, which turn out to be this variable that needs to be configured.

### DIFF
--- a/vulnerablecode/settings.py
+++ b/vulnerablecode/settings.py
@@ -27,9 +27,7 @@ if not DEV_MODE:
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
 
-ALLOWED_HOSTS = [
-    ".herokuapp.com",
-]
+ALLOWED_HOSTS = os.environ.get("VC_ALLOWED_HOSTS", "*").split(":")
 
 # Application definition
 
--- a/README.rst
+++ b/README.rst
@@ -160,6 +160,11 @@ for this purpose::
 
     SECRET_KEY=$(python -c "from django.core.management import utils; print(utils.get_random_secret_key())")
 
+You will also need to setup the VC_ALLOWED_HOSTS environment variable to match the hostname where the app is deployed::
+
+    VC_ALLOWED_HOSTS=vulnerablecode.your.domain.example.com
+
+You can specify several host by separating them with a colon `:`
 
 Using Nix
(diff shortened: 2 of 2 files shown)


Hritik14:
This is strange. Why do we even have heroku hardcoded in there?

sbs2001:
@Hritik14 
There was a demo thing up after gsoc 2017 on heroku, hence it stayed there I guess.

sbs2001:
@tardyp  thanks for the PR ! I'm passing up this to @tdruez  to review.

Hritik14:
@sbs2001 I guess we should remove heroku as well. It could turn out to really unwanted for some installations.

vulnerablecode/settings.py
```diff
@@ -28,7 +28,7 @@
 DEBUG = False
 
 ALLOWED_HOSTS = [
-    ".herokuapp.com",
+    os.environ.get("VC_ALLOWED_HOSTS", ".herokuapp.com"),
```
tdruez, line 31: `VC_ALLOWED_HOSTS` here is not great since you are providing a single host through the environ and not a list of hosts.
I would suggest a simple `VC_HOST` for this env variable.
Also, we probably want to remove the ".herokuapp.com" default.

tdruez (changes requested):

vulnerablecode/settings.py
tardyp: I think it is better to keep it near the django semantics, so I switched the variable to be splitable.

I removed herokuapp to switch to '*' for the default, as I am not sure we really want to get it not work by default.

Hritik14:
Is it not better to let the let the user configure their `settings.py` by themselves ? 
We could document it further and arrange it better and point the users to configure their settings.py in the readme.

tardyp:
In my case, I am deploying vulnerablecode via docker in a kubernetes cluster, so I would prefer to have a default configuration which does not require to modify the app code.

vulnerablecode/settings.py
sbs2001: Now that `VC_ALLOWED_HOSTS` can be used to specify multiple hosts separated via a colon,  it solves both of the issues you guys pointed out ?

sbs2001 (approved):

tdruez (approved):

PR #407, opened 2021-03-28
Removed repeated dependency in reqirements.txt.

vignesh-cloud-prog:
Signed-off-by: vignesh-cloud-prog <vigneshun80@gmail.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,7 +37,6 @@ pygit2==1.5.0
 Pygments==2.6.1
 pyparsing==2.4.5
 pytest==5.3.2
-django-widget-tweaks==1.4.8 
 pytest-dependency==0.4.0
 pytest-django==3.7.0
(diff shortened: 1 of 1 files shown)


sbs2001:
What is repeated ? Could you elaborate ? Your commit message is also misleading, you are adding deps not removing them.

requirements.txt
```diff
@@ -37,7 +37,6 @@ pygit2==1.5.0
 Pygments==2.6.1
 pyparsing==2.4.5
 pytest==5.3.2
-django-widget-tweaks==1.4.8
```
sbs2001, line 40: This is required for the UI

sbs2001 (changes requested):
This is required for the UI

requirements.txt
vignesh-cloud-prog: I am sorry about that

Issue #343, opened 2021-02-10
Deal with  duplicate reference urls


sbs2001:
![image](https://user-images.githubusercontent.com/28975399/107518768-66b0c900-6bd5-11eb-8ca3-df640266a824.png)

sbs2001:
Explore storing canonical form of urls

savish28:
Hey @sbs2001 regarding this issue I am thinking of overriding the save method of `VulnerabilityReference` model and overriding the instance URL with `str(urlpy.parse(instance_url).canonical())`.
Secondly, Do we want something to be done regarding http/https?

sbs2001:
@savish28 thanks for looking into this,  your approach would indeed make things better. 

As for dealing with `http vs https`  we could either 

1. Have a script which removes the duplicates ?
or 
2.We could also override the save method to check for the same canonical url except with `https`/`http` and do things accordingly. 
If provided url is http but we already have the url with https we ignore the `http` url. 
If provided url is https but we already have the url with `http` we replace the `http` url with `https` one 

This approach would increase db hits tho

But in any case `str(urlpy.parse(instance_url).canonical())` would do only good.

vignesh-cloud-prog:
Can i take up this

sbs2001:
@savish28  are you working on this ?  

@vignesh-cloud-prog  this is a non trivial issue btw (not a good first )

savish28:
@sbs2001 Yes, I would open a PR in some time.

TG1999:
With the latest models the references can not be duplicated, closing this now.

PR #408, opened 2021-03-29
Backend: Fix duplicate reference urls(#343)

savish28:
Fix: #343.

The following changes are been proposed in this PR:
- [x] Saving URLs in canonical form ('http://foo.com/?b=2&a=1&d=3' -> 'http://foo.com/?a=1&b=2&d=3'
- [x] If provided url is http but we already have the url with https we ignore the http url.
- [x] If provided url is https but we already have the url with http we replace the http url with https one

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -23,6 +23,7 @@
 import importlib
 from datetime import datetime
 from time import sleep
+import urlpy
 
 from django.db import models
 from django.db import IntegrityError
@@ -109,6 +110,38 @@ class VulnerabilityReference(models.Model):
     @property
     def scores(self):
         return VulnerabilitySeverity.objects.filter(reference=self.id)
+
+    def save(self, *args, **kwargs):
+        if self.id:
+            super(VulnerabilityReference, self).save(*args, **kwargs)
+        else:
+            url_parsed = urlpy.parse(self.url)
+            self.url = str(url_parsed.canonical())
+            url_scheme = url_parsed.scheme
+            scheme_independent_url = self.url[len(url_scheme) :]
+            if url_scheme == "http":
+                similar_instance = VulnerabilityReference.objects.filter(
+                    vulnerability=self.vulnerability,
+                    source=self.source,
+                    reference_id=self.reference_id,
+                    url="https" + scheme_independent_url,
+                ).first()
+                if not similar_instance:
+                    super(VulnerabilityReference, self).save(*args, **kwargs)
+            elif url_scheme == "https":
+                similar_instance = VulnerabilityReference.objects.filter(
+                    vulnerability=self.vulnerability,
+                    source=self.source,
+                    reference_id=self.reference_id,
+                    url="http" + scheme_independent_url,
+                ).first()
+                if similar_instance:
+                    similar_instance.url = self.url
+                    similar_instance.save()
+                else:
+                    super(VulnerabilityReference, self).save(*args, **kwargs)
+            else:
+                super(VulnerabilityReference, self).save(*args, **kwargs)
 
     class Meta:
         unique_together = ("vulnerability", "source", "reference_id", "url")
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -112,7 +112,7 @@ class VulnerabilityReference(models.Model):
         return VulnerabilitySeverity.objects.filter(reference=self.id)
 
     def save(self, *args, **kwargs):
-        if self.id:
+        if self.id or not self.url:
             super(VulnerabilityReference, self).save(*args, **kwargs)
         else:
             url_parsed = urlpy.parse(self.url)
--- a/requirements.txt
+++ b/requirements.txt
@@ -56,3 +56,4 @@ requests==2.23.0
 toml==0.10.2
 PyYAML==5.4
 freezegun==1.1.0
+urlpy==0.5


savish28:
@sbs2001 ping. Could you please review the PR.

sbs2001:
@savish28  I'm not really sure how much of impact this is going to have on the db side things, since we are doing more queries per `save`.  The imports are already slow enough. 

So here's what I suggest. Keep the url normalization part in the `save` method. The scheme thing, should be moved into a separate job/command, https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/ .

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
     def scores(self):
         return VulnerabilitySeverity.objects.filter(reference=self.id)
 
+    def save(self, *args, **kwargs):
+        if self.id:
+            super(VulnerabilityReference, self).save(*args, **kwargs)
+        else:
+            url_parsed = urlpy.parse(self.url)
+            self.url = str(url_parsed.canonical())
```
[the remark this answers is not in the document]
sbs2001: Or `self.url=""`

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
...
+            url_scheme = url_parsed.scheme
+            scheme_independent_url = self.url[len(url_scheme) :]
+            if url_scheme == "http":
```
sbs2001, line 122: This block should be broken down into a separate function.

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
...
+                similar_instance = VulnerabilityReference.objects.filter(
+                    vulnerability=self.vulnerability,
+                    source=self.source,
+                    reference_id=self.reference_id,
+                    url="https" + scheme_independent_url,
+                ).first()
+                if not similar_instance:
+                    super(VulnerabilityReference, self).save(*args, **kwargs)
+            elif url_scheme == "https":
```
sbs2001, line 131: Same as above, This block should be broken down into a separate function.

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
...
+                similar_instance = VulnerabilityReference.objects.filter(
+                    vulnerability=self.vulnerability,
+                    source=self.source,
+                    reference_id=self.reference_id,
+                    url="http" + scheme_independent_url,
+                ).first()
+                if similar_instance:
+                    similar_instance.url = self.url
+                    similar_instance.save()
```
sbs2001, line 140: `return similar_instance.save()` would be better here for readability

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
...
+                else:
+                    super(VulnerabilityReference, self).save(*args, **kwargs)
```
sbs2001, line 142: Use a `return` here

vulnerabilities/models.py
```diff
@@ -110,6 +111,38 @@ class VulnerabilityReference(models.Model):
...
+            else:
+                super(VulnerabilityReference, self).save(*args, **kwargs)
```
sbs2001, line 144: Same as above, for the computers it doesn't make difference, but when reading I don't have to verify whether the return value is correct.

savish28: > @savish28 I'm not really sure how much of impact this is going to have on the db side things, since we are doing more queries per `save`. The imports are already slow enough.
> 
> So here's what I suggest. Keep the url normalization part in the `save` method. The scheme thing, should be moved into a separate job/command, https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/ .

@sbs2001 So, a nice idea, Does this mean that the command/job `python manage.py job_name` have to be called explicitly by the admin to run the following job and it would go through all the entries and it would update the db. I think it can be extended in two ways:-
1. If no arguments are passed then all the entries are looked upon and are checked.
2. If arguments like id/Vulnerability/source etc are passed and only filtered objects are processed.

I think that in this case, each run would cost O(N) queries and would not be automated.

But in the save method update, it would cost 1 extra query per update and would be automated. So I think query wise the override save method is more time-efficient and more automated.

What would be your suggestion in the case and if we are going with the job idea, does the proposed method would work?

vulnerabilities/models.py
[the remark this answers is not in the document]
savish28: Updated the code. It returned "/" for the same but keeping "" for "" would make more sense so updated the changes.

vulnerabilities/models.py
sbs2001: @savish28  

>called explicitly by the admin

We could get clever by running the job after each import, so this is a non-issue, I guess. 

>1. If no arguments are passed then all the entries are looked upon and are checked.
  2. If arguments like id/Vulnerability/source etc are passed and only filtered objects are processed.
    
Btw I want to remove the source thing when I get time :) it's not populated anywhere and is a bad way to keep track of source(s) .  With that out of the way, what would be the use case of (2) ? In almost all cases (1) would be used. 

> each run would cost O(N) queries

IMHO that can be done `1 + (no of duplicates) `. One query to fetch all references (just the id and url columns) . Iterate over all records from server side, check for dups in the already fetched records. If a dup is found follow the procedure along the lines of what you're doing here, so that's gonna cost 1 `DELETE`  query per dup. 

Dups are rare anyways so it won't be O(N) . 

> it would cost 1 extra query per update

IMHO that'd be 1 extra query per insert/ (`save` method invocation in most cases )

savish28: Ohh okay, I get the idea. We are importing all the data and then doing this job. This would result in more efficiency rather than doing operation at each save.

So I guess these are the deliverables now,
- [ ] Create Job for removing duplicates.
- [ ] Add functionality to run the job after each import.
Does this seem right? @sbs2001

sbs2001:
@savish28  I just realized there's a third way to do this which would be even better. 

Let's move your logic in a wrapper method like `_get_or_create_keep_secure` (this a bad name, need to discuss this more). We don't use save method explicitly anyways. https://github.com/nexB/vulnerablecode/blob/dfb1ab5558c597240573988cb981b498f98a964c/vulnerabilities/import_runner.py#L127 that's where refs get created. A `get_or_create` results in 2 queries, we can replace that with `_get_or_create_keep_secure` .Your logic is essentially  `create_keep_secure` part, need to implement the `get` part . 

So as far as queries are concerned `_get_or_create_keep_secure` would atmost require 2 queries which is same as that of `get_or_create`.

sbs2001:
@savish28  ping :)

pombredanne:
@savish28 gentle ping.

pombredanne:
@savish28 do you think you want to complete and bring this to a mergeable state?
Otherwise we may close for now and you can reopen later.

savish28:
> @savish28 do you think you want to complete and bring this to a mergeable state?
> Otherwise we may close for now and you can reopen later.

@pombredanne Sure, closing this PR for now.

Issue #403, opened 2021-03-26
run vulnerablecode behind proxy


tardyp:
Looking at deploying this in a proxy protected environment I figured that the import will not work behind proxy

the rootcause is pygit2 which does not support proxy. There is experimental support in the mainline, but it is not yet released.
https://github.com/libgit2/pygit2/pull/1063

In general, I would respectfully question the use of pygit2 in favor gitpython. 

Being backed by standard git makes it in my experience more reliable in all environments.

Would you accept a PR switching this dependency?

sbs2001:
@tardyp  

> the rootcause is pygit2 which does not support proxy

That's a nice find. Needs to be fixed soon.

> respectfully question the use of pygit2 in favor gitpython

I'm not super familiar with the pygit2 code we have here, but I guess at the time  pygit2 was used due to existing familiarity and sketched out a quick working prototype. We can always iterate and improve. 

> Would you accept a PR switching this dependency?

Yes !! That would be great. 

@pombredanne  ping to you

tardyp:
PR done, with unit test working for now.  Pubishing for early review

I'll be working this afternoon in running it with some real data

pombredanne:
> In general, I would respectfully question the use of pygit2 in favor gitpython.
> 
> Being backed by standard git makes it in my experience more reliable in all environments.
> 
> Would you accept a PR switching this dependency?

Sure thing. The alternative would be to possibly look into FetchCode https://github.com/nexB/fetchcode/ which is using the extracted guts of pip for the parts that deals with version control systems (it but cannot deal much with checkouts though, but may not be an issue) e.g. it uses the git CLI directly, but either way is fine.

tardyp:
The data_source.py module actually has incremental fetch ability and change detection code.

However, it looks like the caller actually do not care, and working directories are always recreated from scratch for all runs of import.

This may not be a huge issue.

FWYI there is another proxy issue with the package code, which I will send a PR soon.

Hritik14:
>  it looks like the caller actually do not care, and working directories are always recreated from scratch for all runs of import.

This is configurable here 
https://github.com/nexB/vulnerablecode/blob/3688d89a88014f2e04475fda0179362e9ab54eae/vulnerabilities/data_source.py#L241-L243

tardyp:
> This is configurable here

right, but nobody uses it except for tests. Am I wrong?

Hritik14:
[...]

PR #409, opened 2021-03-29
Switch data_source dependency to GitPython

tardyp:
While neither pygit2 or GitPython have a formidable API and documentation

GitPython has the advantage of being pure python and support
same proxy feature as git.

I removed some of the tests, which to me are much over mocked, and do not really test much.

I also changed the diff algorithm so that it only diff the whole thing instead of considering diff commit by commits.

Fix: #403


vulnerabilities/data_source.py
```diff
@@ -319,34 +319,38 @@ def _collect_file_changes(
...
+        cutoff_commit = None
+        for commit in self._repo.iter_commits(self._repo.head):
+            if commit.committed_date < self.cutoff_timestamp:
                 break
+            cutoff_commit = commit
 
-            if previous_commit is None:
-                previous_commit = commit
-                continue
+        if cutoff_commit is None:
```
pombredanne, line 331: Could `if cutoff_commit` be enough here?

vulnerabilities/data_source.py
```diff
@@ -319,34 +319,38 @@ def _collect_file_changes(
...
+        for d in cutoff_commit.diff(self._repo.head.commit):
+            if not _include_file(d.b_path, subdir, recursive, file_ext) or _is_binary(d):
+                continue
 
-            previous_commit = commit
+            abspath = os.path.join(self.config.working_directory, d.b_path)
+            if d.new_file:
+                added_files.add(abspath)
+            elif d.a_blob and d.b_blob and d.a_path != d.b_path:
+                added_files.add(abspath)  # consider moved files as added
```
pombredanne, line 351: If you do not mind, I like to avoid end of line comments when possible
```suggestion
                # consider moved files as added
                added_files.add(abspath) 
```

vulnerabilities/data_source.py
```diff
@@ -319,34 +319,38 @@ def _collect_file_changes(
...
-                    updated_files.add(abspath)
+        for d in cutoff_commit.diff(self._repo.head.commit):
+            if not _include_file(d.b_path, subdir, recursive, file_ext) or _is_binary(d):
+                continue
 
-            previous_commit = commit
+            abspath = os.path.join(self.config.working_directory, d.b_path)
+            if d.new_file:
+                added_files.add(abspath)
+            elif d.a_blob and d.b_blob and d.a_path != d.b_path:
```
pombredanne, line 350: Would it help if the part `d.a_blob and d.b_blob` here and below was refactored in one test followed by two sub-tests: one on path and one blob?

vulnerabilities/tests/test_data_source.py
```diff
@@ -28,8 +28,8 @@
 from unittest.mock import MagicMock
 from unittest.mock import patch
 import xml.etree.ElementTree as ET
+import git
```
pombredanne, line 31: nit: add a line to separate from std. library imports
```suggestion

import git
```

vulnerabilities/tests/test_data_source.py
```diff
@@ -51,8 +51,34 @@ def load_oval_data():
...
+def clone_url(tmp_path):
+    git_dir = tmp_path / "git_dir"
+    repo = git.Repo.init(str(git_dir))
+    new_file_path = str(git_dir / "file")
+    open(new_file_path, "wb").close()
+    repo.index.add([new_file_path])
+    repo.index.commit("Added a new file")
+
+    yield str(git_dir)
+    shutil.rmtree(git_dir)
```
pombredanne, line 64: What about putting the `shutil.rmtree(` in a `try/finally` block?

pombredanne (changes requested):
Thank you++ :+1: 
I added  a few nitpickings for your consideration.
Also you may want to add the pinned deps of gitpython to the requirements file?

vulnerabilities/data_source.py
tardyp: I usually restrict to None comparaison when possible (explicit is better than implicit) and I can see that most of this file is also using this principle

tardyp:
@pombredanne I think I have answer to your remarks.

i fixed the remaining tests, and nix build.

vulnerabilities/data_source.py
```diff
@@ -319,35 +319,41 @@ def _collect_file_changes(
...
-                if not _include_file(d.new_file.path, subdir, recursive, file_ext) or d.is_binary:
-                    continue
+        def _is_binary(d: DiffIndex):
+            if not d.b_blob:
+                return False
+            try:
+                d.b_blob.data_stream.read().decode()
+            except UnicodeDecodeError:
+                return True
+            return False
```
sbs2001, line 341: What is this function trying to do ? Is it used to check whether the diff blob consists of binary ? This would return true if the diff is about something like a tar ball ?

tardyp: Indeed, it checks by trying to decode it with default encoding settings.

The previous code had some logic of checking if the data was binary to not count it as a diff.

The is_binary semantic was provided by libgit2. The logic is quite complicated:
https://github.com/libgit2/libgit2/blob/508361401fbb5d87118045eaeae3356a729131aa/src/buf_text.c#L268

pombredanne: @sbs2001 as you suggested in the Gitter chat I guess we could use binaryornot which does a decent job there and is used extensively in scancode-toolkit:
- https://github.com/audreyfeldroy/binaryornot
- https://github.com/nexB/thirdparty-packages/blob/main/pypi/binaryornot-0.4.4.tar.gz.ABOUT
- https://github.com/nexB/typecode/blob/157b7be15c6cb4be754f915bde7974a3e0284961/src/typecode/contenttype.py#L926

tardyp: I changed the PR to use binaryornot. I am not sure why the nix CI is failing :-/

vulnerabilities/data_source.py
```diff
@@ -319,35 +319,35 @@ def _collect_file_changes(
...
+        def _is_binary(d: DiffIndex):
+            return is_binary_string(d.b_blob.data_stream.read(1024))
 
-                abspath = os.path.join(self.config.working_directory, d.new_file.path)
-                # TODO
-                # Just filtering on the two status values for "added" and "modified" is too
-                # simplistic. This does not cover file renames, copies &
-                # deletions.
-                if d.status == pygit2.GIT_DELTA_ADDED:
+        for d in cutoff_commit.diff(self._repo.head.commit):
```
sbs2001, line 337: This errors out with older versions of git (mine was `2.25.1`).  See https://github.com/gitpython-developers/GitPython/issues/1016. 

We could just add a note about this in the README I guess ?

sbs2001:
@tardyp  this looks good to me aside from the issue with git version . IMHO the nix test failure has nothing to do with your changes. Those are probably caused to due to some sync issues. @rolfschr  ping ! . 

Thanks for this, and sorry for being late.

etc/nix/flake.nix
```diff
@@ -74,7 +74,7 @@
             name = "vulnerablecode-${version}";
             src = vulnerablecode-src;
             dontConfigure = true; # do not use ./configure
-            propagatedBuildInputs = [ pythonEnv postgresql ];
+            propagatedBuildInputs = [ pythonEnv postgresql gitMinimal];
```
sbs2001, line 77: @rolfschr   is this fine on your side ?

rolfschr: Look good to me. @tardyp  I gues you could remove the ssl cert file fix for pygit2 in the nix test if you remove pygit2 anyways.

```diff
diff --git a/etc/nix/flake.nix b/etc/nix/flake.nix
index 1a65045..84ed672 100644
--- a/etc/nix/flake.nix
+++ b/etc/nix/flake.nix
@@ -119,10 +119,6 @@
 
             buildInputs = [ wget vulnerablecode ];
 
-            # Used by pygit2.
-            # See https://github.com/NixOS/nixpkgs/pull/72544#issuecomment-582674047.
-            SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
-
             unpackPhase = "true";
 
             buildPhase = ''
```

sbs2001 (approved):
Merging this, I'll take care of nix tests and docs in another PR. 
Thanks @tardyp  :)

tardyp:
Sorry I was in vacation.. thanks for merging.

pombredanne:
> Sorry I was in vacation.. thanks for merging.


Sure thing. Thanks for the contrib!

PR #413, opened 2021-03-29
Misc fixes for deploying vulnerablecode on a container platform

tardyp:
This is my last few fixes for today. 

We can now call vulnerable code enterprise ready ;)

### DIFF
--- a/vulnerablecode/settings.py
+++ b/vulnerablecode/settings.py
@@ -26,6 +26,7 @@ if not DEV_MODE:
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
+DEBUG_PROPAGATE_EXCEPTIONS = True
 
 ALLOWED_HOSTS = [
     ".herokuapp.com",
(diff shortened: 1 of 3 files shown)


vulnerablecode/settings.py
```diff
@@ -26,6 +26,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = False
+DEBUG_PROPAGATE_EXCEPTIONS = True
```
sbs2001, line 29: Enabling this when the `DJANGO_DEV=1` env variable is also set is probably better idea.

sbs2001 (changes requested):

vulnerablecode/settings.py
tardyp: well, the whole point is to stil enable this when not using DJANGO_DEV.

If we don't have this in prod, we have absolutely no visibility in case of error 500.

This setting allows to propagate the exception to the base logger, and thus to display them in runserver's stdout.
I dont think there are security implication to this.

If we are in DJANGO_DEV, then DEBUG is true, and exception is printed in the browser along with debug UI which is really insecure.

PR #415, opened 2021-03-30
Add unspecified scoring system

Hritik14:
This could be used where the scale of scoring system is unspecified and do not correspond to a numeric representation. This must be used as the last resort. 
Required in https://github.com/nexB/vulnerablecode/pull/397#discussion_r603774595 and https://github.com/nexB/vulnerablecode/pull/393#discussion_r603964317


sbs2001 (changes requested):
@Hritik14  it makes sense to have something when there's no doc for severities.

But in this case I think we are losing information  by calling it "unspecified" . 
IMHO we should call it  "generic_textual". That way the downstream projects would be still able to interpret these.  And we should give the url a default value of empty string.

Hritik14:
@sbs2001 done

sbs2001 (approved):
LGTM

PR #418, opened 2021-04-01
Fix redhat import failure

### LINKED ISSUE #398: Investigate and fix redhat import failure
The redhat importer is still failing after the fix at https://github.com/nexB/vulnerablecode/pull/387. 

From  the logs at https://pipelines.actions.githubusercontent.com/Ep9prel8NYELDl3abhBbfgiORCbX0JuViN8i8esKzVS8DdQt5i/_apis/pipelines/1/runs/151/signedlogcontent/3?urlExpires=2021-03-24T04%3A20%3A33.2100559Z&urlSigningMethod=HMACV1&urlSignature=7VXaMaJelaCEUG6EDzDbZKLAw%2BfWbGUypuHe164RCjE%3D

It seems there is problem at https://github.com/nexB/vulnerablecode/blob/bfa005ece230ada9032c3c0b4a7bab7b59673e0b/vulnerabilities/importers/redhat.py#L93. 

To fix this:

1.Find the cause of 443 
2.Add a workaround on non-json response .

Hritik14:
This fixes #398. 
This PR is mainly comprises of following two commits.

---
### requests_with_5xx_retry: Retry on 5xx errors

Introduced and used a helper function for retries on 5xx errors. This is
important and some servers like bugzilla.redhat.com return 502 Proxy Error
which was the cause of #398

A ticket has been raised in RedHat here https://redhat.service-now.com/help?id=rh_ticket&table=sc_req_item&sys_id=278239541b1ba010477e43fccd4bcb4a

Here is the response from bugzilla-owner@redhat.com
> Yes, just back off a bit and retry when this happens. It's been a long-term issue and we're working on a way to address it but it will still take time. There's also a time-out on individual queries - if you're hitting that sometimes on other requests then you'll need to optimise them or break them up into smaller ones
---
### Not all RHSA errata have a CVRF document

This is mentioned in the NOTE of "2.1 List all CVRFs" of
https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/cvrf

Such a case would lead to a crash before this commit.
Eg: https://access.redhat.com/hydra/rest/securitydata/cvrf/RHSA-2005:835.json
No cvrfdoc would be found in the statement
                    value = rhsa_data["cvrfdoc"]["aggregate_severity"]

### DIFF
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -25,6 +25,7 @@ import re
 
 import requests
 import toml
+import urllib3
 import yaml
 
 # TODO add logging here
@@ -79,3 +80,21 @@ def create_etag(data_src, url, etag_key):
 
 
 is_cve = re.compile(r"CVE-\d+-\d+", re.IGNORECASE).match
+
+
+def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
+    """
+    Returns a requests sessions which retries on 5xx errors with
+    a backoff_factor
+    """
+    retries = urllib3.util.Retry(
+        total=max_retries,
+        backoff_factor=backoff_factor,
+        raise_on_status=True,
+        status_forcelist=range(500, 600, 1),
+    )
+    adapter = requests.adapters.HTTPAdapter(max_retries=retries)
+    session = requests.Session()
+    session.mount("https://", adapter)
+    session.mount("http://", adapter)
+    return session
--- a/vulnerabilities/importers/redhat.py
+++ b/vulnerabilities/importers/redhat.py
@@ -20,16 +20,19 @@
 #  VulnerableCode is a free software code from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
-from packageurl import PackageURL
 import requests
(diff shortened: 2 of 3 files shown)


vulnerabilities/helpers.py
```diff
@@ -79,3 +80,21 @@ def create_etag(data_src, url, etag_key):
 
 
 is_cve = re.compile(r"CVE-\d+-\d+", re.IGNORECASE).match
+
+
+def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
```
sbs2001, line 85: You said there were 104 errors,  how are they addressed ?

sbs2001 (approved):
LGTM, though I hate we need to do this low level requests things ;)

vulnerabilities/helpers.py
Hritik14: 104 came after retrying with short delays for multiple times. I increased the `backoff_factor` to 1 in the redhat importer.

Hritik14:
@sbs2001 Perhaps we would be able to revert back after redhat fixes the problem on their side. Until then, this is what we got.

Issue #318, opened 2021-01-27
manage.py import should provide some kind of progress feedback


pombredanne:
Right now we do not know what happens and what was done.
Design TBD

AmitGupta7580:
I added progress bar in NVD importer . Using `tqdm` package

![image](https://user-images.githubusercontent.com/57043342/113254007-a8134a00-92e3-11eb-895d-86be0762fc32.png)

AmitGupta7580:
I face some problems while adding progress bar in other importers. 
As you know data is importing in 2 different processes.

1. Downloading data from resources and cast into Advisory class.
2. Adding advisory to postgresql database.

Some importers do these task at same time e.g : NVD
While other importers first collect all the data then save it to database e.g : archlinux

[Solution] To add a common progress bar we has to implement all importers in similar fashion.

Hritik14:
> Eventually we want to move to non-batch proccessing
I think this can be better done after https://github.com/nexB/vulnerablecode/issues/338

unibik:
Hi! I was looking this issue#318 and was wondering if it is still open or if it has already been resolved. If it's still open, I'd be happy to work on it. Please let me know. Thank you!

pombredanne:
As @keshav-space pointed to, we are moving to all-background importer and improvers running as pipelines.

Therefore a command line progress bar does not make sense in the general case.

What we need instead is something the same as ScanCode.io loop progress https://github.com/aboutcode-org/scancode.io/blob/e1607c7f9684dbca9cf67bbbbf3145e0aaf06fdf/aboutcode/pipeline/__init__.py#L260 and this is now part of https://pypi.org/project/aboutcode.pipeline/ which we have already in VCIO and all the new pipelines are using this progress indicator.

I am closing this as completed.

pombredanne:
The main issue for pipelines is:
- https://github.com/aboutcode-org/vulnerablecode/issues/1509

PR #419, opened 2021-04-01
Add progress bar while importing data from importers

Abhra303:
This fixes the issue #318 . Here I used tqdm package to visualize the process of importing with a progress bar .  Here is the screenshot below - 
 
import_runner py - vulnerablecode  Container vulnerablecode_web (_vulnerablecode)  - Visual Studio Code 01-04-2021 15_21_11

### DIFF
--- a/vulnerabilities/import_runner.py
+++ b/vulnerabilities/import_runner.py
@@ -30,11 +30,13 @@ import traceback
 from typing import Set
 from typing import Tuple
 from typing import Optional
+from tqdm import tqdm
 
 import packageurl
 from django.db import DataError
 from django.core import serializers
 
+from django.core.management.base import BaseCommand
 from vulnerabilities import models
 from vulnerabilities.data_source import Advisory, DataSource
 from vulnerabilities.data_source import PackageURL
@@ -78,7 +80,7 @@ class ImportRunner:
         self.importer = importer
         self.batch_size = batch_size
 
-    def run(self, cutoff_date: datetime.datetime = None) -> None:
+    def run(self, cutoff_date: datetime.datetime = None, command: BaseCommand = None) -> None:
         """
         Create a data source for the given importer and store the data retrieved in the database.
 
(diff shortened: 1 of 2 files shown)


AmitGupta7580:
@Abhra303 Try your code on NVD importer. As I mentioned https://github.com/nexB/vulnerablecode/issues/318#issuecomment-811688097 . You are only showing progress bar for saving data to database. In case of archlinux the downloading time is less (5280 vulnerabilites) thats why it does not make any difference.

Abhra303:
@AmitGupta7580  you're right. I didn't think about it. Trying if that can be managed.

sbs2001:
@Abhra303  any issues ?

Abhra303:
Actually no severe issue.
When we ran the import command, it first fetches the data from the source then store them in the database.
The progress bar that I have  implemented is for process_advisory function and this function is used to store data into database. So the progress bar is showing the progress of storing the data into the database not the progress of fetching the data from the source.
As both of these functions use for loops , they takes time to fetch the data or store the data.
So for importers such as nvd , the data they fetch is huge. As a result, when I run import command the progress bar is showing after a significant time.

Abhra303:
User experience for import progress is improved. 
Fetching vulnerability data from a source can be time consuming based on user's internet speed. We have nothing to do with it. But the "Collecting data from source..." statement will let the user know that the importer is now fetching vulnerability data from source . After fetching all data , user will see the progress bar that denotes the download of those data into database. Here is a screenshot - 

import py - vulnerablecode  WSL_ Ubuntu-18 04  - Visual Studio Code 06-04-2021 14_40_49

Abhra303:
@sbs2001 can you please review the changes?

AmitGupta7580:
@Abhra303 Instead of just writing `Collecting the data from the source`, we should also have a progress bar which display the status of downloaded data, As importer is still not fully verbose.

Abhra303:
@AmitGupta7580  I was thinking of adding progress bar for collecting the data also. To do this , we have to add tqdm function in every importer file manually. Not only this, if in future other contributors make new importer files then they also have to take care of tqdm. 
Another point is that if in future any issue generate related to the way importers collect data,  then developers also may have to change some code related to tqdm to keep the behaviour of cli same and this have to be done in every importer file. 
For these reasons, I thought that it will be better to add the progress bar only for the procress_advisory function.
But yes, If reviewers like @AmitGupta7580  's idea, I will surely add progress bar for collecting the data also..

sbs2001:
@Abhra303 it seems to me there would be some overhead for importer implementation to get the progress bar working. 

IMHO this ticket would be best resolved after we remove the batch processing things and have a uniform way for importer implementation.

Abhra303:
Thank you for the clarification.

PR #421, opened 2021-04-02
Drop dephell specifier

sbs2001:
Finally ! 

We need to  publish univers to pypi IMHO before merging this. Currently this uses the suboptimal way of fetching univers via git repo.


sbs2001:
@rolfschr  the nix tests are failing because this PR is changing some deps.

Issue #423, opened 2021-04-03
Update Apache_httpd importer from XML to JSON Advisory


AmitGupta7580:
Fetching URL : `https://httpd.apache.org/security/vulnerabilities-httpd.xml` no longer exists.

![image](https://user-images.githubusercontent.com/57043342/113480631-fb87c280-94b2-11eb-961e-41a169d2311e.png)

PR #425, opened 2021-04-04
Update Apache_httpd importer from XML to JSON Advisory

AmitGupta7580:
Resolve https://github.com/nexB/vulnerablecode/issues/423

Tasks : 

- [x] Fetch data into JSON format.
- [x] Differentiate impacted and resolved packages.
- [x] Add severity of vulnerability.
- [x] Create Test Case class for Apache_httpd importer.


vulnerabilities/importers/apache_httpd.py
```diff
@@ -21,14 +21,18 @@
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 import dataclasses
-from xml.etree import ElementTree
 
+from bs4 import BeautifulSoup
```
sbs2001, line 25: Nit: Sort your imports

vulnerabilities/importers/apache_httpd.py
```diff
@@ -21,14 +21,18 @@
...
 import requests
+import re
```
sbs2001, line 27: Nit: Group your imports into 3 groups 
1. stdlib imports at top
2. 3rd party imports at middle (eg requests)
3. project imports (eg "from vulnerabiities.......")

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +53,58 @@ def updated_advisories(self):
...
-def fetch_xml(url):
-    resp = requests.get(url).content
-    return ElementTree.fromstring(resp)
+def fetch_links(url):
+    links = []
+    data = requests.get(url).content
+    soup = BeautifulSoup(data, features="lxml")
+    for tag in soup.find_all("a"):
+        link = tag.get("href")
+        if not re.search("^CVE.*json$", link):
```
sbs2001, line 107: Using Regex here is overkill. Also you can chop off one line of code by doing something lik 

Use 
```python
link = tag.get("href")
if link.endswith("json"):
  .....do something 

```

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +53,58 @@ def updated_advisories(self):
...
+            continue
+        links.append(url + link)
```
sbs2001, line 109: A robust option would be to use 
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin instead of  string concat.

vulnerabilities/tests/test_apache_httpd.py
```diff
@@ -0,0 +1,96 @@
...
+from packageurl import PackageURL
+
+from unittest import TestCase
+
+from vulnerabilities.data_source import Reference
+from vulnerabilities.data_source import Advisory
+from vulnerabilities.data_source import VulnerabilitySeverity
+from vulnerabilities.severity_systems import scoring_systems
+from vulnerabilities.importers.apache_httpd import ApacheHTTPDDataSource
+from vulnerabilities.importers.apache_httpd import fetch_links
```
sbs2001, line 35: As mentioned before, sort and group your imports

vulnerabilities/tests/test_apache_httpd.py
```diff
@@ -0,0 +1,96 @@
...
+TEST_DATA = os.path.join(BASE_DIR, "test_data/apache_httpd/CVE-1999-1199.json")
+
+
+class TestApacheHTTPDDataSource(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        data_source_cfg = {"etags": {}}
+        cls.data_src = ApacheHTTPDDataSource(1, config=data_source_cfg)
+        with open(TEST_DATA) as f:
+            cls.nvd_data = json.load(f)
```
sbs2001, line 47: This is not nvd_data anymore.

vulnerabilities/tests/test_apache_httpd.py
```diff
@@ -0,0 +1,96 @@
...
+                "rather than increasing at a constant rate. This makes a denial of service "
+                "attack based on this method more effective than methods which cause Apache"
+                " to use memory at a constant rate, since the attacker has to send less data.",
+                impacted_package_urls=[
+                    PackageURL(
+                        type="apache",
+                        namespace=None,
+                        name="httpd",
+                        version="1.3.1",
+                        qualifiers={},
```
sbs2001, line 65: You could get rid of 
qualifiers={}, namespace=None, subpath=None 

This iws all over the codebase and not required since those are defaults

sbs2001 (changes requested):
Thanks, this looks good to be merged with minor changes I suggested

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +53,58 @@ def updated_advisories(self):
...
         return []
 
-
-def to_advisories(data):
-    advisories = []
-    for issue in data:
+    def to_advisory(self, data):
+        cve = data["CVE_data_meta"]["ID"]
+        description = data["description"]["description_data"]
+        summary = next((item["value"] for item in description if item["lang"] == "eng"), "")
```
pombredanne, line 68: Using next here is a little cryptic. Would there be a more readable and explicit way to do this?
I reckon that `description` has this shape (and are you sure this is always present?):

```JSON
  "description": {
    "description_data": [
      {
        "lang": "eng",
        "value": "For configurations using proxying...."
      }
    ]
  }
```
so what about this?
```
descriptions = data.get("description", {}).get("description_data", [])
description = None
for desc in descriptions:
    if desc.get("lang") == "eng":
        description = desc.get("value")
        break
```

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +53,58 @@ def updated_advisories(self):
...
+        severity = VulnerabilitySeverity(
+            system=scoring_systems["apache_httpd"],
+            value=data["impact"][0]["other"],
```
pombredanne, line 71: This feels as being magical with `[0]`... Let's try something more explicit and easier to read possibly like the suggestion above. In particular if there is no impact, then no `VulnerabilitySeverity` should be created. As it the code may be brittle and crash easily if this happens.

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +53,58 @@ def updated_advisories(self):
...
+        )
+        reference = Reference(reference_id=cve, url=self.url + cve + ".json", severities=[severity])
```
pombredanne, line 73: Use a template and an f-string instead of  concatenating strings or a urljoin

pombredanne (changes requested):

vulnerabilities/importers/apache_httpd.py
AmitGupta7580:
```
impacts = data.get("impact", [])
impact = None
for imp in impacts:
     value = imp.get("other")
     if value is not None:
          impact = value
          break

if impact is not None:
     severity = VulnerabilitySeverity(
          system=scoring_systems["apache_httpd"],
          value=impact,
     )
     reference = Reference(reference_id=cve, url=self.url + cve + ".json", severities=[severity])
else:
     reference = Reference(reference_id=cve, url=self.url + cve + ".json")
```
please ignore concatenation of url I am working on it.

sbs2001 (changes requested):
I guess this needs to rebased and refactored according to new model (which would 1-2 line changes IMHO) See the new importer now for reference

AmitGupta7580:
image

Done 👍

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +54,81 @@ def updated_advisories(self):
...
+            reference = Reference(
+                reference_id=cve,
+                url=self.ref_url.format(cve),
+                severities=[severity],
+            )
+        else:
+            reference = Reference(
+                reference_id=cve,
+                url=self.ref_url.format(cve),
             )
```
sbs2001, line 97: This big code block can be avoided by doing something like : 

```python
severities = []
if impact:
  severities.append( VulnerabilitySeverity(......))

reference = Reference(...... severities=severities)

```

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +54,81 @@ def updated_advisories(self):
...
+            for products in vendor["product"]["product_data"]:
+                for version in products["version"]["version_data"]:
+                    version_value = version["version_value"]
+
+                    if version["version_affected"] == "<":
+                        fixed_packages.append(
+                            PackageURL(type="apache", name="httpd", version=version_value)
+                        )
+
+                    else:
```
sbs2001, line 112: This is probably right, but could you be more explicit with expected values of version range operator. 
Something like:
```python
elif version["version_affected"] == "=" or version["version_affected"] == "<=" .... : 
..... 
```

would be more helpful.

sbs2001: We would also want to resolve that version  range into discrete versions, that way this data would be more rich. 

For fetching all version of httpd you'd need https://github.com/nexB/vulnerablecode/blob/deb44c68345315085fbab07d2f8b58e93327d08b/vulnerabilities/package_managers.py#L345 and https://github.com/apache/httpd . Anyways let me know if you need additional hep here.

vulnerabilities/tests/test_apache_httpd.py
```diff
@@ -0,0 +1,92 @@
...
+            Advisory(
+                summary="A serious problem exists when a client sends a large number of "
+                "headers with the same header name. Apache uses up memory faster than the "
+                "amount of memory required to simply store the received data itself. That "
+                "is, memory use increases faster and faster as more headers are received, "
+                "rather than increasing at a constant rate. This makes a denial of service "
+                "attack based on this method more effective than methods which cause Apache"
+                " to use memory at a constant rate, since the attacker has to send less data.",
+                affected_packages=[
+                    AffectedPackage(
```
sbs2001, line 59: If we start resolving the version ranges, I'm guessing the patched_package field of these instances would also get populated.

sbs2001 (changes requested):

vulnerabilities/importers/apache_httpd.py
AmitGupta7580:
```python
severities = []
if impact is not None:
    severities.append(
        VulnerabilitySeverity(
            system=scoring_systems["apache_httpd"],
            value=impact,
        )
    )
reference = Reference(
    reference_id=cve,
    url=self.ref_url.format(cve),
    severities=severities,
)
```

vulnerabilities/tests/test_apache_httpd.py
AmitGupta7580: Sorry Didn't understand it. Please explain bit more.

vulnerabilities/importers/apache_httpd.py
```diff
@@ -21,15 +21,23 @@
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
 import dataclasses
-from xml.etree import ElementTree
+import asyncio
```
sbs2001, line 24: Nit: Sort your imports

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
         # skips processing the response further to avoid duplicate work
 
         if create_etag(data_src=self, url=self.url, etag_key="ETag"):
```
sbs2001, line 64: IMHO this won't work becuase `self.url == "https://httpd.apache.org/security/json/{}"`

vulnerabilities/importers/apache_httpd.py
```diff
@@ -40,7 +48,12 @@ class ApacheHTTPDDataSourceConfiguration(DataSourceConfiguration):
 class ApacheHTTPDDataSource(DataSource):
 
     CONFIG_CLASS = ApacheHTTPDDataSourceConfiguration
-    url = "https://httpd.apache.org/security/vulnerabilities-httpd.xml"
+    url = "https://httpd.apache.org/security/json/{}"
```
sbs2001, line 51: Maybe use urllib here  instead ? That would be more concise in the code. 

eg you could have something like:
```python
base_url = "https://httpd.apache.org/security/json/"
```
Then in `fetch_links` you could obtain full urls by doing something like
```
links.append(urllib.parse.urljoin(self.base_url, link))
```

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
+        if impact is not None:
+            severities.append(
+                VulnerabilitySeverity(
+                    system=scoring_systems["apache_httpd"],
+                    value=impact,
+                )
+            )
+        reference = Reference(
+            reference_id=cve,
+            url=self.ref_url.format(cve),
```
sbs2001, line 101: Continuation of above `base_url` suggestion.

ref_url can be eliminated. Instead here do something like
`url= urllib.parse.urljoin(self.base_url, cve, ".json")`

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
+            for link in links:
+                data = requests.get(link).json()
+                advisories.append(self.to_advisory(data))
             return self.batch_advisories(advisories)
 
         return []
 
+    def to_advisory(self, data):
+        cve = data["CVE_data_meta"]["ID"]
+        descriptions = data.get("description", {}).get("description_data", [])
```
sbs2001, line 77: Just curious, does this cascade all the way down. Could you point me to an example where `descriptions == []` .

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
+        description = None
+        for desc in descriptions:
+            if desc.get("lang") == "eng":
```
sbs2001, line 80: When does this condition not get satisifed, ie does `desc.get("lang") == None` happen ?

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
+            if value is not None:
+                impact = value
+                break
+        if impact is not None:
+            severities.append(
+                VulnerabilitySeverity(
+                    system=scoring_systems["apache_httpd"],
+                    value=impact,
+                )
+            )
```
sbs2001, line 98: This block can be made more concise via 

```python
impacts = data.get("impact", [])
for impact in impacts:
  value = impact.get("other")
  if value:
    severities.append(
                VulnerabilitySeverity(
                    system=scoring_systems["apache_httpd"],
                    value=value,
                )
     break  
    

```

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
 
-def to_advisories(data):
-    advisories = []
-    for issue in data:
-        resolved_packages = []
-        impacted_packages = []
-        for info in issue:
-            if info.tag == "cve":
-                cve = info.attrib["name"]
+        versions = []
```
sbs2001, line 105: This name is misleading because really, it stores some version related data (which includes ranges and such).

Maybe rename this to `version_data`

vulnerabilities/importers/apache_httpd.py
```diff
@@ -49,47 +62,110 @@ def updated_advisories(self):
...
+
+        return Advisory(
+            vulnerability_id=cve,
+            summary=description,
+            affected_packages=nearest_patched_package(affected_packages, fixed_packages),
+            references=[reference],
         )
 
-    return advisories
+    def to_version_ranges(self, versions):
```
sbs2001, line 141: :+1:

vulnerabilities/importers/apache_httpd.py
sbs2001: This won't be an issue if we go with `base_url` approach.

sbs2001 (changes requested):
@AmitGupta7580  thanks, this looks good with minor changes.

vulnerabilities/tests/test_apache_httpd.py
sbs2001: Looking from the code, I think you understood it well :)

vulnerabilities/importers/apache_httpd.py
AmitGupta7580: > `url= urllib.parse.urljoin(self.base_url, cve, ".json")`

this results to only `https://httpd.apache.org/security/json/CVE-1999-1199`.
Can I use `urllib.parse.urljoin(self.base_url, "{}.json".format(cve))` or 
`urllib.parse.urljoin(self.base_url, f"{cve}.json")` ?

vulnerabilities/importers/apache_httpd.py
AmitGupta7580: No

vulnerabilities/importers/apache_httpd.py
AmitGupta7580: No, currently there is no example where `descriptions == []`.

I ran this code but got nothing.
``` python
descriptions = data.get("description", {}).get("description_data", [])
if len(descriptions) == 0:
    print("empty descriptions : {}".format(cve))
description = None
for desc in descriptions:
    if desc.get("lang") == "eng":
         description = desc.get("value")
    else:
         print("other language : {}".format(cve))
```

sbs2001: This is too defensive then, may mask issues, ideally we want to fail loudly. 

Does  `descriptions = data["description"]["description_data"]` work ? If yes then use that, otherwise try 
`descriptions = data["description"].get("description_data", [])`

vulnerabilities/importers/apache_httpd.py
sbs2001: I'm guessing that `desc.get("lang")` is never `None` (`"lang"` key is always present in `desc`) . If yes then use `if desc["lang"] == "eng" : .....`

vulnerabilities/importers/apache_httpd.py
```diff
@@ -40,56 +49,121 @@ class ApacheHTTPDDataSourceConfiguration(DataSourceConfiguration):
...
+        for desc in descriptions:
+            if desc.get("lang") == "eng":
+                description = desc.get("value")
+                break
+
+        severities = []
+        impacts = data.get("impact", [])
+        for impact in impacts:
+            value = impact.get("other")
+            if value is not None:
```
sbs2001, line 88: Use `if value:` instead, they are the same but concise and easy to read.

sbs2001 (changes requested):

vulnerabilities/importers/apache_httpd.py
sbs2001: The f string one works.

vulnerabilities/importers/apache_httpd.py
sbs2001: `description` is always present, @AmitGupta7580  checked it.

vulnerabilities/importers/apache_httpd.py
```diff
@@ -40,56 +49,121 @@ class ApacheHTTPDDataSourceConfiguration(DataSourceConfiguration):
...
     def updated_advisories(self):
         # Etags are like hashes of web responses. We maintain
         # (url, etag) mappings in the DB. `create_etag`  creates
         # (url, etag) pair. If a (url, etag) already exists then the code
         # skips processing the response further to avoid duplicate work
 
-        if create_etag(data_src=self, url=self.url, etag_key="ETag"):
-            data = fetch_xml(self.url)
-            advisories = to_advisories(data)
+        if create_etag(data_src=self, url=self.base_url, etag_key="ETag"):
```
sbs2001, line 64: This isn't valid now because the provided endpoint doesn't provide etags in the headers. Remove this if statement and the emtpy list return below

sbs2001: Also the comments related to  docs

vulnerabilities/importers/apache_httpd.py
```diff
@@ -20,16 +20,25 @@
...
+from univers.version_specifier import VersionSpecifier
 
 from vulnerabilities.data_source import Advisory
 from vulnerabilities.data_source import DataSource
 from vulnerabilities.data_source import DataSourceConfiguration
+from vulnerabilities.data_source import Reference
+from vulnerabilities.data_source import VulnerabilitySeverity
+from vulnerabilities.package_managers import GitHubTagsAPI
+from vulnerabilities.severity_systems import scoring_systems
 from vulnerabilities.helpers import create_etag
```
sbs2001, line 40: Nit this isn't used anymore

sbs2001 (approved):

sbs2001:
@AmitGupta7580  I'm squashing the commits and merging it, consider squashing from your side next time .

AmitGupta7580:
> consider squashing from your side next time .

@sbs2001 didn't understand  :( . Did I do something wrong?

PR #427, opened 2021-04-04
Fixes #312 : Added solution for ModuleNotFoundError

### LINKED ISSUE #312: Run error after pull latest, "ModuleNotFoundError: No module named 'drf_spectacular'".
Hello,

I pulled latest and ran with Docker Compose, getting the error below:

image

Pushpit07:
Added commands to run in case of running into a ModuleNotFoundError

Signed-off-by: Pushpit <pushpit07@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,6 +106,12 @@ compose. For this you need to have the following installed.
 Use ``sudo docker-compose up`` to start VulnerableCode. Then access
 VulnerableCode at http://localhost:8000/ or at http://127.0.0.1:8000/
 
+If you run into a ``ModuleNotFoundError: No module named '<module>'``, run::
+
+    sudo docker-compose exec web bash
+    pip install -r requirements.txt
+
+
 Use ``sudo docker-compose exec web bash`` to access the VulnerableCode
 container. From here you can access ``manage.py`` and run management commands
 to import data as specified below.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -106,6 +106,12 @@ compose. For this you need to have the following installed.
 Use ``sudo docker-compose up`` to start VulnerableCode. Then access
 VulnerableCode at http://localhost:8000/ or at http://127.0.0.1:8000/
 
+If you run into a ``ModuleNotFoundError: No module named '<module>'``, run::
```
sbs2001, line 109: let's make this more general. 

The root cause of problem here is that after upgrading the code, the docker setup is not synced ie new migrations, deps are not applied. 

So a better solution to fix all such problems would be to do 
`sudo docker-compose up -d --no-deps --build web`

So I would change this to something like 
```
Don't forget to run `sudo docker-compose up -d --no-deps --build web` to sync your instance after every `git pull`.
```

sbs2001 (changes requested):
There's a better solution to solve all such problems, see my comment inline.

README.rst
Pushpit07: oh, great. Running this will sync the docker environment each time the codebase updates

README.rst
```diff
@@ -106,6 +106,9 @@ compose. For this you need to have the following installed.
 Use ``sudo docker-compose up`` to start VulnerableCode. Then access
 VulnerableCode at http://localhost:8000/ or at http://127.0.0.1:8000/
 
+**Important**: Don't forget to run `sudo docker-compose up -d --no-deps --build web` to sync your instance after every `git pull`.
```
sbs2001, line 109: You need to enclose the command in 2 tildes

PR #428, opened 2021-04-04
Resolves #399: Rectify invalid id in msr2019

### LINKED ISSUE #399: Invalid ids appearing as vulnerability id in msr2019
Ids such as `CVE-2011-4461-JETTY` are appearing as vulnerability id which are clearly not valid CVEs.
It can be mitigated by either
- Using `is_cve` helper and the `.group(0)` to get the matching cve 
- Adding `$` in is_cve helper making it `r"CVE-\d+-\d+$"` in order to avoid any invalid matches.

Pushpit07:
Added $ in is_cve helper

Earlier: r"CVE-\d+-\d+" 
Now: r"CVE-\d+-\d+$" 

This would avoid any invalid matches that were happening earlier

### DIFF
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -79,7 +79,7 @@ def create_etag(data_src, url, etag_key):
     return True
 
 
-is_cve = re.compile(r"CVE-\d+-\d+$", re.IGNORECASE).match
+is_cve = re.compile(r"CVE-\d{4}-\d{4,7}", re.IGNORECASE).match
 
 
 def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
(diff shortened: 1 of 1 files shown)


vulnerabilities/helpers.py
```diff
@@ -79,7 +79,7 @@ def create_etag(data_src, url, etag_key):
     return True
 
 
-is_cve = re.compile(r"CVE-\d+-\d+", re.IGNORECASE).match
+is_cve = re.compile(r"CVE-\d+-\d+$", re.IGNORECASE).match
```
sbs2001, line 82: IMHO the regex `CVE-\d{4}-\d{4,7} ` is more explicit and accurate.  Let's use that instead.

sbs2001 (changes requested):
Thanks, see my comment inline

Pushpit07:
Done

sbs2001 (approved):

PR #432, opened 2021-04-06
Fix ubuntu_usn importer for invalid CVE

AmitGupta7580:
Error:

image

Fixed:

image


vulnerabilities/importers/ubuntu_usn.py
```diff
@@ -74,7 +75,7 @@ def to_advisories(usn_db):
                 # The db sometimes contains entries like
                 # {'cves': ['python-pgsql vulnerabilities', 'CVE-2006-2313', 'CVE-2006-2314']}
                 # This `if` filters entries like 'python-pgsql vulnerabilities'
-                if not cve.startswith("CVE-"):
+                if not is_cve(cve):
                     continue
```
sbs2001, line 79: skipping entirely would be bad idea as we'll be losing other data. 

Instead I suggest 

```
  if not is_cve(cve):
                    cve = ""
```

sbs2001 (changes requested):

vulnerabilities/importers/ubuntu_usn.py
AmitGupta7580: Yes, It's a good idea as empty CVE will automatically takes the VULCOID-X-X-X.

AmitGupta7580:
Around 600-700 vulnerabilities increased now.

image

image

sbs2001 (approved):
Thanks @AmitGupta7580  I'm merging this

Issue #375, opened 2021-03-12
Get rid of bazillions of resolved/patched packages


sbs2001:
Security advisories usually provide a version range of packages which are vulnerable to some vulnerability. Currently we iterate over every available version of a package and check if it satisfies any of the given version range. If  it satisfies this range then  we treat this package version as vulnerable.

The versions which don't satisfy any range are treated as patched/resolved packages. This is wrong on many levels. 
eg. consider today I found `CVE-2021-xxx` affecting package `foo v2020` via the version range `<=2020` . 10 years later I will be stating that  `foo v2030` is patched to  `CVE-2021-xxx` . At that point this info is kind of garbage and redundant.  

Instead for `{package, vulnerability}`  pair there should be only one resolving/patched package, which completely fixes this vulnerability.

`resolved_package`  should be turned into a `resolving_package`. This `resolving_package` should be related to `{package, vulnerability}`  pair(s).

One rough design for new `PackageRelatedVulnerability`
https://github.com/nexB/vulnerablecode/blob/d8ce30e29c0d3257a1d00d89feb0afa753ba3f69/vulnerabilities/models.py#L176

could be 

```python
class PackageRelatedVulnerability(models.Model):

    package = models.ForeignKey(Package, on_delete=models.CASCADE)
    vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
    resolving_package = models.ForeignKey(Package, on_delete=models.CASCADE)
    

    def __str__(self):
        return f"{self.package.package_url} {self.vulnerability.vulnerability_id}"

    class Meta:
        unique_together = ("package", "vulnerability")
        verbose_name_plural = "PackageRelatedVulnerabilities"

```

This also makes https://github.com/nexB/vulnerablecode/issues/207#issuecomment-783317662 more approachable

PR #436, opened 2021-04-09
Add patched package 

sbs2001:
Fixes https://github.com/nexB/vulnerablecode/issues/375

I've done a terrible job at naming things here(see review comments), need help there. 

We would want to move all the intermediate data structures to one place. That will be done in another PR. 

See high level overview at https://github.com/nexB/vulnerablecode/pull/436#issuecomment-823137427


vulnerabilities/tests/test_models.py
```diff
@@ -66,3 +64,39 @@ def test_vulnerability_save_without_vulnerability_id(self):
                 ).count()
                 == 1
             )
+
+
+# FIXME: The fixture code is duplicated. setUpClass is not working with the pytest mark.
+@pytest.mark.django_db
+class TestPackageRelatedVulnerablity(TestCase):
```
sbs2001, line 71: @tdruez  these are the tests

vulnerabilities/data_source.py
```diff
@@ -87,17 +88,16 @@ class Advisory:
 
     summary: str
     vulnerability_id: Optional[str] = None
-    impacted_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
-    resolved_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
+    affected_packages_with_patched_package: List[
```
sbs2001, line 91: Name problem 1

vulnerabilities/helpers.py
```diff
@@ -20,16 +20,29 @@
...
+from packageurl import PackageURL
+from univers.versions import version_class_by_package_type
 
 # TODO add logging here
 
+is_cve = re.compile(r"CVE-\d{4}-\d{4,7}", re.IGNORECASE).match
+
+
+@dataclasses.dataclass(order=True, frozen=True)
+class AffectedPackageWithPatchedPackage:
```
sbs2001, line 42: This will be moved somewhere else in another PR

vulnerabilities/helpers.py
```diff
@@ -100,9 +118,41 @@ def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
     return session
 
 
-def contains_alpha(string):
-    """
-    Return True if the input 'string' contains any alphabet
-    """
+def nearest_patched_package(vulnerable_packages, resolved_packages):
```
sbs2001, line 121: This is where the magic happens, again it has poor naming. Help.

pombredanne (changes requested):
Some commits message is just "WIP"... may be there is a better message to reword? :angel:

vulnerabilities/admin.py
```diff
@@ -51,7 +51,7 @@ class PackageAdmin(admin.ModelAdmin):
 
 @admin.register(PackageRelatedVulnerability)
 class PackageRelatedVulnerabilityAdmin(admin.ModelAdmin):
-    list_filter = ("is_vulnerable", "package__type", "package__namespace")
+    list_filter = ("package__type", "package__namespace")
```
pombredanne, line 54: Is there anything that could replace is_vulnerable?

vulnerabilities/data_source.py
pombredanne: Could you survey the words use to depict affected/patched in other tools? 
- kaybee/vulas
- dependencycheck and dependency track (and OWASP)
- in general all the data source we use
...
with these listed here we can make a good pick together!

vulnerabilities/helpers.py
```diff
@@ -79,7 +92,12 @@ def create_etag(data_src, url, etag_key):
     return True
 
 
-is_cve = re.compile(r"CVE-\d{4}-\d{4,7}", re.IGNORECASE).match
+def contains_alpha(string):
+    """
+    Return True if the input 'string' contains any alphabet
```
pombredanne, line 97:
```suggestion
    Return True if the input 'string' contains any alphanumeric character```

File: vulnerabilities/helpers.py [hunk 7]
```diff
@@ -100,9 +118,41 @@ def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
     return session
 
 
-def contains_alpha(string):
-    """
-    Return True if the input 'string' contains any alphabet
-    """
+def nearest_patched_package(vulnerable_packages, resolved_packages):
+    # This class is used to  get around bisect module's lack of supplying custom
```
pombredanne, line 122: This should be a docstring

vulnerabilities/helpers.py
```diff
@@ -100,9 +118,41 @@ def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
...
+
+    vulnerable_packages = sorted(
+        [PackageURLWithVersionComparator(package) for package in vulnerable_packages]
+    )
+    resolved_packages = sorted(
+        [PackageURLWithVersionComparator(package) for package in resolved_packages]
+    )
+
+    resolved_package_count = len(resolved_packages)
+    affected_package_with_patched_package_objects = []
```
pombredanne, line 144: Why the _objects suffix?
```suggestion
    affected_package_with_patched_packages = []
```

vulnerabilities/importer_yielder.py
```diff
@@ -97,13 +97,13 @@
...
         "data_source_cfg": {"repository_url": "https://github.com/RetireNet/Packages.git"},
     },
-    {
-        "name": "suse_backports",
-        "license": "",
-        "last_run": None,
-        "data_source": "SUSEBackportsDataSource",
-        "data_source_cfg": {"url": "http://ftp.suse.com/pub/projects/security/yaml/", "etags": {}},
-    },
+    # {
```
pombredanne, line 100: Why is this disabled?

pombredanne (commented):
Thanks!
We need to find better names alright and settle on this once and for all IMHO.
vulnerable/affected/resolved/patched are used and I feel we are not consistent.

Also can you explain the high level design here?

sbs2001:
This PR removes the notion of `resolved_package` essentially only storing packages which are vulnerable. It also links these vulnerable packages with nearest(version-wise) package which is safe.

On a lower level, this required :  

1. Removing PackageRelatedVulnerability's `is_vulnerable` flag, since now we only store vulnerable packages primarily.

2. Adding a field called `patched_package` to `PackageRelatedVulnerability` . This is a foreign key to the nearest(wrt `PackageRelatedVulnerability.package` version wise) safe package .

3. Implementing a method called `nearest_patched_package` which accepts 2 lists of vulnerable packages and safe packages (wrt to some same vulnerablity) respectively. This method does the following:

 a. Sort both the lists , ie the packages are now in ascending order wrt their versions. `univers` is used here.
 b. Iterate over each vulnerable package, let's call the package being iterated `pkg`. For each `pkg` we find a package 
               which is just greater(version-wise ) than itself in the safe packages list. See the code [here](https://github.com/sbs2001/vulnerablecode/blob/8e85ff4e75174bbfe0e5169171eafdb81cd2165c/vulnerabilities/helpers.py#L147) 

c. If there exists such a package in (b) we bind this safe package with `pkg` by creating a [AffectedPackageWithPatchedPackage instance](https://github.com/sbs2001/vulnerablecode/blob/8e85ff4e75174bbfe0e5169171eafdb81cd2165c/vulnerabilities/helpers.py#L42)

vulnerabilities/importer_yielder.py
sbs2001: We probably want to get rid of that importer, since it only contains `(vulnerability, fixed/backported package)`

vulnerabilities/admin.py
sbs2001: that should be `patched_package` adding it.

vulnerabilities/helpers.py
sbs2001: I'm aware of the principle of not specifying the type. The problem here is this list contains instances of `AffectedPackageWithPatchedPackage` . I don't know how to pluralise that, 

`affected_package_with_patched_packages` -> is it one one affected package with multiple patched packages
`affected_packages_with_patched_package`  -> vice versa 

This is a mess :(

vulnerabilities/helpers.py
sbs2001: Use `AffectedPackageWithPatch`

vulnerabilities/data_source.py
pombredanne: The general renaming should be tracked in a separate ticket IMHO, not here.

Beyond this, what about `affected_packages` or just `impacted_packages`?

vulnerabilities/helpers.py
pombredanne: Or just `AffectedPackage` ?

vulnerabilities/helpers.py
```diff
@@ -20,16 +20,29 @@
...
 
 # TODO add logging here
 
+is_cve = re.compile(r"CVE-\d{4}-\d{4,7}", re.IGNORECASE).match
+
+
+@dataclasses.dataclass(order=True, frozen=True)
+class AffectedPackageWithPatchedPackage:
+    vulnerable_package: PackageURL
+    patched_package: Optional[PackageURL] = None
```
pombredanne, line 44: What about: 
```suggestion
    patched_version: Optional[str] = None
```

vulnerabilities/helpers.py
pombredanne: we will fix that in a separate ticket.

vulnerabilities/helpers.py
```diff
@@ -100,9 +118,41 @@ def requests_with_5xx_retry(max_retries=5, backoff_factor=0.5):
...
 
-def contains_alpha(string):
-    """
-    Return True if the input 'string' contains any alphabet
-    """
+def nearest_patched_package(vulnerable_packages, resolved_packages):
+    # This class is used to  get around bisect module's lack of supplying custom
+    # compartor. Get rid of this once we use python 3.10 which supports this.
+    # See https://github.com/python/cpython/pull/20556
+    class PackageURLWithVersionComparator:
```
pombredanne, line 125: Are you missing a @total_ordering there?

Also you should mention this is only to compare packages of the same type OR you could implement comparison and sorting across types.

pombredanne: You may want to check that the type/ns/name are the same  too

vulnerabilities/helpers.py
pombredanne: Based on other review, `affected_package` would be a better name for now.

pombredanne (approved):
I added  a few more comments for your consideration and please merge.
Let's have a separate ticket for the naming issue.

PR #443, opened 2021-04-18
helper: split_markdown_front_matter

Hritik14:
helper for istio and mozilla importers
The following two alternatives are also proposed:
```py
def split_markdown_front_matter(lines: str) -> Tuple[str, str]:
    fmlines = []
    mdlines = []
    splitter = mdlines

    for index, line in enumerate(lines.split("\n")):
        if index == 0 and line.strip().startswith("---"):
            splitter = fmlines
        elif line.strip().startswith("---"):
            splitter = mdlines
        else:
            splitter.append(line)

    return "\n".join(fmlines), "\n".join(mdlines)
```

and the one present [here](https://github.com/mozilla/foundation-security-advisories/blob/d43d09d204ab5da014e83b7d1743df289cefee92/check_advisories.py#L183-L208) under MPL license.

### DIFF
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -26,6 +26,7 @@ import json
 import re
 from typing import Optional
 from typing import List
+from typing import Tuple
 
 import requests
 import saneyaml
@@ -164,3 +165,30 @@ def nearest_patched_package(
         )
 
     return affected_package_with_patched_package_objects
+
+
+def split_markdown_front_matter(text: str) -> Tuple[str, str]:
+    r"""
+    Split text into markdown front matter and the markdown body
+    Returns ("", text) for text with non existing front matter
+
+    >>> text='''---
+    ... title: DUMMY-SECURITY-2019-001
+    ... description: Incorrect access control.
+    ... cves: [CVE-2042-1337]
+    ... ---
+    ... # Markdown starts here
+    ... '''
+    >>> split_markdown_front_matter(text)
+    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]\n', '\n# Markdown starts here\n')
+    """
+
+    front_matter = ""
+    body = text
+    text = text.replace("\r\n", "\n")
+    linezero,_, text = text.partition("---\n")
+
+    if not linezero: # nothing before first ---
+        front_matter,_, body = text.partition("---")
+
+    return front_matter, body
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -167,7 +167,7 @@ def nearest_patched_package(
     return affected_package_with_patched_package_objects
 
 
-def split_markdown_front_matter(text: str) -> Tuple[str, str]:
+def split_markdown_front_matter(lines: str) -> Tuple[str, str]:
     r"""
     Split text into markdown front matter and the markdown body
     Return ("", text) for text with non existing front matter
@@ -180,15 +180,19 @@ def split_markdown_front_matter(text: str) -> Tuple[str, str]:
     ... # Markdown starts here
     ... '''
     >>> split_markdown_front_matter(text)
-    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]\n', '\n# Markdown starts here\n')
+    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]', '# Markdown starts here\n')
     """
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
 
-    front_matter = ""
-    body = text
-    text = text.replace("\r\n", "\n")
-    linezero, _, text = text.partition("---\n")
+    lines = lines.replace("\r\n", "\n")
+    for index, line in enumerate(lines.split("\n")):
+        if index == 0 and line.strip().startswith("---"):
+            splitter = fmlines
+        elif line.strip().startswith("---"):
+            splitter = mdlines
+        else:
+            splitter.append(line)
 
-    if not linezero:  # nothing before first ---
-        front_matter, _, body = text.partition("---")
-
-    return front_matter, body
+    return "\n".join(fmlines), "\n".join(mdlines)
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -24,8 +24,8 @@ import bisect
 import dataclasses
 import json
 import re
-from typing import Optional
 from typing import List
+from typing import Optional
 from typing import Tuple
 
 import requests
@@ -182,6 +182,7 @@ def split_markdown_front_matter(text: str) -> Tuple[str, str]:
     >>> split_markdown_front_matter(text)
     ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]', '# Markdown starts here')
     """
+    # The doctest contains \n and for the sake of clarity I chose raw strings than escaping those.
     lines = text.splitlines()
     if lines[0] == "---":
         lines = lines[1:]
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -186,9 +186,9 @@ def split_markdown_front_matter(text: str) -> Tuple[str, str]:
     front_matter = ""
     body = text
     text = text.replace("\r\n", "\n")
-    linezero,_, text = text.partition("---\n")
+    linezero, _, text = text.partition("---\n")
 
-    if not linezero: # nothing before first ---
-        front_matter,_, body = text.partition("---")
+    if not linezero:  # nothing before first ---
+        front_matter, _, body = text.partition("---")
 
     return front_matter, body
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -170,7 +170,7 @@ def nearest_patched_package(
 def split_markdown_front_matter(text: str) -> Tuple[str, str]:
     r"""
     Split text into markdown front matter and the markdown body
-    Returns ("", text) for text with non existing front matter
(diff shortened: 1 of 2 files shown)


sbs2001:
This looks good to me, could you swap this with the thing we have in istio and run tests. That way we'll cover some more testcases and the refactor would also be done.

sbs2001:
As far as the alternative approaches are concerned, I would stick with something dumb and simple (implementation wise). 
The mozilla one, and the one in comments seems the simplest. I'm not sure that the  one in the comments  would work in case the frontmatter  doesn't start from first line. Does that even happen ?

Hritik14:
> swap this with the thing we have in istio and run tests

Done. 

> The mozilla one, and the one in comments seems the simplest

This PR actually originally offers the same code with a minor improvement.

> I'm not sure that the one in the comments would work in case the frontmatter doesn't start from first line. Does that even happen ?

I'm not aware of any instance where front matter doesn't start at the very beginning.  I looked at some implementations and it appears that it is rather a strict requirement. (Opening fence must begin on the first line of the markdown string/file [ref](https://github.com/ParkSB/markdown-it-front-matter))
Also, as `---` is a valid markdown, I don't think it is allowed for it to mark the beginning of front matter without being at the very top.
I've verified this assumption with github gists and it appears to be true.
Here are the results:  https://gist.github.com/Hritik14/7d88ed5d9789c0f5d726746d654590dd (github shows front matter in a nice table)

sbs2001:
@Hritik14  thanks for the research and the refactor.

Let's use the one you suggested in the comments then. Though the current implementation is concise, it's not super obvious.

Hritik14:
@sbs2001 I've updated it with the one in comments though I'd really appreciate @pombredanne's insight here as he was really in favor of the `partition` (https://github.com/nexB/vulnerablecode/pull/443/commits/d8ea80151a71e3742e50fe843e0af2ddfaabe1b7) approach.

Hritik14:
@sbs2001 @pombredanne ping

vulnerabilities/importers/istio.py
```diff
@@ -210,10 +174,10 @@ def process_file(self, path):
         return advisories
 
     def get_data_from_md(self, path):
-        """Return a mapping of vulnerability data from istio . The data is
-        in the form of yaml_lines inside a .md file.
+        """Return a mapping of vulnerability data from istio. The data is
+        in the form of yaml objects found inside front matter of the .md file.
```
pombredanne, line 178: You are telling too much about how the sausage is made here in your documentation.
 `Return a mapping of vulnerability data extracted from an advisory.` may be plenty enough.

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    ... cves: [CVE-2042-1337]
+    ... ---
+    ... # Markdown starts here
+    ... '''
+    >>> split_markdown_front_matter(text)
+    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]', '# Markdown starts here\n')
+    """
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
```
pombredanne, line 187: I personally find this a little bit hard to understand and read.
What about no splitter and just `frontmatter = []` and `markdown = []` ? and then beow append directly.

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
         )
 
     return affected_package_with_patched_package_objects
+
+
+def split_markdown_front_matter(lines: str) -> Tuple[str, str]:
```
pombredanne, line 170: I would call this `text` rather than `lines` which suggests we have a list of lines and we have not.

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    r"""
```
pombredanne, line 171: Why a raw string? 
```suggestion
    """
```

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    Split text into markdown front matter and the markdown body
```
pombredanne, line 172:
```suggestion
    Return a tuple of (front matter, markdown body) strings split from ``text``.
    Each can be an empty string.
```
(I use text rather than lines as a variable name)

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    """
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
+
+    lines = lines.replace("\r\n", "\n")
+    for index, line in enumerate(lines.split("\n")):
+        if index == 0 and line.strip().startswith("---"):
+            splitter = fmlines
+        elif line.strip().startswith("---"):
```
pombredanne, line 193: You should test equality, not start.

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    >>> split_markdown_front_matter(text)
+    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]', '# Markdown starts here\n')
+    """
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
+
+    lines = lines.replace("\r\n", "\n")
+    for index, line in enumerate(lines.split("\n")):
+        if index == 0 and line.strip().startswith("---"):
```
pombredanne, line 191: I understand what you are doing here with your "splitter" variable and this is sleek. But this may not be easy to read and understand at least it was not obvious to me at first. 

What could you do to make this easier to read?
1. the first test can be done outside of the loop since if the document starts with `---` it must be on the first line which can be safely discarded (the first line).
2. with that done, then the only thing left is that this is frontmatter until the line is `---` and at this stage you switch to markdown/body. Or you could partition the remaining joined text on `---`?

Or may be at least finding a better name than `splitter` would help? something like bucket?

Here are some alternative ways for your consideration:
```Python
frontmatter = []
markdown = []
lines = text.splitlines(False)
if lines[0] == "---":
    lines = lines[1:]
current_bucket = frontmatter
for line in lines:
    if line == "---":
        current_bucket = markdown
        continue
    current_bucket.append(line)
```
or this one not using any temp lists:
```Python
lines = text.splitlines(False)
# discard the first separator
if lines[0] == "---":
    lines = lines[1:]
text = "\n".join(lines)
frontmatter, _, markdown = text.partition("---\n")
```

The point I am trying to make is that the important thing is to make the code readable and easy to understand such that 6 months from now you or anyone can get what you meant easily. So naming and being explicit and clear matters a lot IMHO.

pombredanne (changes requested):
Thanks! See some comments inline for your consideration.

vulnerabilities/helpers.py
```diff
@@ -164,3 +165,34 @@ def __lt__(self, other):
...
+    ... '''
+    >>> split_markdown_front_matter(text)
+    ('title: DUMMY-SECURITY-2019-001\ndescription: Incorrect access control.\ncves: [CVE-2042-1337]', '# Markdown starts here\n')
+    """
+    fmlines = []
+    mdlines = []
+    splitter = mdlines
+
+    lines = lines.replace("\r\n", "\n")
+    for index, line in enumerate(lines.split("\n")):
```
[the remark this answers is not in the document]
Hritik14: Thanks. fixed

vulnerabilities/importers/istio.py
```diff
@@ -210,10 +174,10 @@ def process_file(self, path):
...
         """
 
         with open(path) as f:
-            yaml_lines = self.get_yaml_lines(f)
-            return self.get_data_from_yaml_lines(yaml_lines)
+            yaml_lines, _ = split_markdown_front_matter(f.read())
```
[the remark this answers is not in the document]
Hritik14: Fixed

vulnerabilities/importers/istio.py
Hritik14: fixed

vulnerabilities/helpers.py
Hritik14: The doctest contains `\n` and for the sake of clarity I chose raw strings than escaping those.

vulnerabilities/helpers.py
Hritik14: fixed

vulnerabilities/helpers.py
Hritik14: fixed

vulnerabilities/helpers.py
Hritik14: fixed

vulnerabilities/helpers.py
Hritik14: I on some implementations it says the separator could be 3 dashes or more. For now, let's settle with 3. We'll look at it if/when some anomaly arises.

vulnerabilities/helpers.py
Hritik14: I took the second implementation and now it looks like 
```py
    lines = text.splitlines()
    if lines[0] == "---":
        lines = lines[1:]
        text = "\n".join(lines)
        frontmatter, _, markdown = text.partition("\n---\n")
        return frontmatter, markdown

    return "", text
```

Lem me know what you think

Hritik14:
Thanks @pombredanne. I've incorporated the latest changes. Do have a look.

vulnerabilities/helpers.py
pombredanne: To avoid future question, it would be great to keep this comment `The doctest contains \n and for the sake of clarity I chose raw strings than escaping those.` below the docstring.

pombredanne (approved):
See a minor nit FYI. Otherwise LGTM!

vulnerabilities/helpers.py
Hritik14: done

PR #444, opened 2021-04-19
Fix nix Github workflow

rolfschr:
This PR automatically fetches the latest [pypi-deps-db](https://github.com/DavHau/pypi-deps-db) before running the automated Nix tests (Github Action). This does *not* commit the changes. However, this should make the nix tests keep running (on Github) when the ``requirements.txt`` change and the obvious fix is to fetch a new version of pypi-deps-db.

### DIFF
--- a/.github/workflows/test-import-using-nix.yml
+++ b/.github/workflows/test-import-using-nix.yml
@@ -25,5 +25,6 @@ jobs:
       - name: run checks & test import
         run: |
           cd etc/nix
+          ./get-latest-pypi-deps-db.sh --in-place
           nix --print-build-logs flake check
           ./test-import-using-nix.sh alpine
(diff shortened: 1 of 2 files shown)


sbs2001:
Hi @rolfschr  thanks :) . 

So I guess this won't need any intervention from your side when deps change.

rolfschr:
> So I guess this won't need any intervention from your side when deps change.

Well, yes, this was my intention :) I expect that some help may needed if case things break for other reasons than a simple missing (bleeding edge version of a) dependency.

sbs2001 (approved):

sbs2001:
Btw @rolfschr  we have a PR that removes the pygit dep which AFAIK caused some pain on your side. 

They are doing some changes to nix tests, could you review that PR ? https://github.com/nexB/vulnerablecode/pull/409

Issue #140, opened 2020-01-14
Store version ranges


pombredanne:
To support vulnerabilities that impact or fix a package version range, we would need to store that data.
In addition to that, we would also store the relationship between vuln and package for every known packages at the time we create or update a vuln.
This is follow up of #119

pombredanne:
From dupe 248:

> #65 and #84 don't provide concrete versions hence we need to store version ranges for that we must first improve the data models

sbs2001:
**Context**

Security advisories as a part of convention provide patched/vulnerable packages using a version range. For eg ruby provides this https://github.com/rubysec/ruby-advisory-db/blob/6efbdb053cbe41e55f435ddee25a1562fb73f3f2/gems/actionpack/CVE-2012-1099.yml#L23

When an advisory provides such version ranges we in ideal case want to convert the version range into a discrete set of versions, ie "resolve the range". 

We do this by 

1. Obtaining all the version released till date of the given package by calling some API. In this case the endpoint will be https://rubygems.org/api/v1/versions/actionpack.json . 

2.  Now we classify the versions into two groups. a. Satisfies the range b. Does not satisfy any range.  And then using some logic `a` group packages are either `vulnerable` or `patched`  while `b` group packages have opposite vulnerability status than the `a` group ones.

**The actual problem** : 

In some cases there is no API to carry out step 1 . ie Version ranges can't be resolved into discrete versions.  We want some way to store such data in a similar manner when we have discrete versions.  Eg https://security.gentoo.org/glsa/202009-18

copernico:
Maybe it is not necessary to require that version ranges should be resolved a-priori (that is, resolved to generate a finite set of *existing* artifacts); they could be used to determine a-posteriori whether or not a given artifact is in the range or not. The advantage of the latter method is that you would not need the API to determine what exists.
In practice though, one needs to enforce some limits on how to interpret intervals that are open-ended (for example, version 2.2 and up).

One way could be to allow only the last segment to vary; for example:

2.2 and up --> 2.2, 2.3, 2.4,....... (but not 3.0)
2.2.1 and up --> 2.2.1, 2.2.2, 2.2.3 (but not 2.3.x)
3.0 and up --> any 3.x (but not 4.x, 5.x etc)

Would this make sense? It's a compromise, this still requires enumerating a few intervals to achieve the correct semantics, but at least it is simple, hopefully easy to grasp, and quite flexible to cover the large majority of cases.

Your thoughts?

[...]

PR #467, opened 2021-05-30
Time travel to the date of advisory publish time when importing

sbs2001:
This PR gets us one step closer to https://github.com/nexB/vulnerablecode/issues/140#issuecomment-847098020

High level working:
1. While parsing any advisory it obtains the publish date of it. Almost every importer does this atm.
2. Since each advisory contains version ranges, we need to resolve it. Normally we would fetch all versions of the package from respective API and resolve it. This PR instead changes improves `*VersionAPI` classes to instead accept a cutoff date, to partition the versions accordingly. Here's a simple snippet
```python
In [3]: import asyncio 
   ...: import pytz 
   ...: from dateutil.parser import parse 
   ...:  
   ...: from vulnerabilities.package_managers import NpmVersionAPI, PypiVersionAPI 
   ...:  
   ...: api = PypiVersionAPI() 
   ...: asyncio.run(api.load_api(["django"])) 
   ...: cutoff_date = parse("2018").replace(tzinfo=pytz.UTC) 
   ...: api.get("django", cutoff_date) 
   ...:  
   ...:                                                                                                                        
Out[3]: 
{'new': { #These are after cutoff_date 2018
.........
  '1.11.14',
  '3.2.4',
  '3.2a1',
  '3.2b1',
  '3.2rc1'
},
 'valid': { # This are before 2018
  '1.2.6'
.......
  '2.0b1',
  '2.0rc1',
  '2.1a1'}}

```

Things to do 
1. Don't use dictionary with `new` and `valid` keys, use a class. @pombredanne  please help in naming these. 
2. Ubuntu and Debian VersionAPI need a fix, the current source doesn't give ALL versions.


vulnerabilities/package_managers.py
```diff
@@ -97,7 +122,16 @@ async def fetch(self, pkg, session):
         try:
             response = await session.request(method="GET", url=url)
             response = await response.json()
-            versions = set(response["releases"])
+            for version in response["releases"]:
```
pombredanne, line 125: What about this to avoid iterating twice:
```
for version, released_downloads in response["releases"].items():
    if not released_downloads:
        continue
    version.add(....

```

otherwise you are making three times a lookup in `response["releases"][version]`

vulnerabilities/package_managers.py
```diff
@@ -97,7 +122,16 @@ async def fetch(self, pkg, session):
...
+                if response["releases"][version]:
+                    versions.add(
+                        Version(
+                            value=version,
+                            release_date=parser.parse(
+                                response["releases"][version][-1]["upload_time_iso_8601"]
```
pombredanne, line 131: Let's use some intermediate variable that makes this clear:
1. we can have several downloads (a wheel, source dist, etc.)
2. we are picking the last one which is entirely arbitrary
3. we are getting the date

Another approach could have been to pick the most recent one or the oldest one.... in all cases this needs to be clarified

vulnerabilities/package_managers.py
```diff
@@ -166,7 +208,13 @@ async def fetch(self, pkg, session):
...
-            versions = {v for v in response.get("versions", [])}
+            for version in response.get("versions", []):
+                release_date = response.get("time", {}).get(version)
+                if release_date:
+                    release_date = parser.parse(release_date)
+                    versions.add(Version(value=version, release_date=release_date))
+                else:
+                    versions.add(Version(value=version, release_date=None))
 
         except ClientResponseError:
```
pombredanne, line 219: See https://github.com/nexB/vulnerablecode/issues/484 for a follow up

vulnerabilities/package_managers.py
```diff
@@ -217,12 +265,15 @@ class MavenVersionAPI(VersionAPI):
     async def load_api(self, pkg_set):
         async with client_session() as session:
             await asyncio.gather(
-                *[self.fetch(pkg, session) for pkg in pkg_set if pkg not in self.cache]
+                *[
+                    self.fetch(pkg, session)
+                    for pkg in pkg_set
+                    if pkg not in self.cache and "camel" not in pkg
```
pombredanne, line 271: Why special case "camel"? I can accept that there could be problem, but this should NOT be handled this way. It should be super explicit and explained thoroughly.

vulnerabilities/package_managers.py
```diff
@@ -217,12 +265,15 @@ class MavenVersionAPI(VersionAPI):
...
+                ]
             )
 
     async def fetch(self, pkg, session) -> None:
-        artifact_comps = pkg.split(":")
-        endpoint = self.artifact_url(artifact_comps)
+        endpoint = self.artifact_url(pkg)
```
pombredanne, line 276: Do not stop using the `maven-metadata.xml` file to get versions. Instead add a new function to parser the HTTP directory listing HTML page if available. (It could very well be absent). Do not try to fetch metadata at the groupid level

vulnerabilities/package_managers.py
```diff
@@ -327,15 +403,21 @@ def composer_url(pkg_name: str) -> str:
...
-        all_versions = {
-            version.replace("v", "") for version in all_versions if "dev" not in version
-        }
+        all_versions = set()
+        for version in resp["packages"][pkg_name]:
+            if "dev" in version:
+                continue
+            all_versions.add(
+                Version(
+                    value=version.replace("v", ""),
```
pombredanne, line 417: Are we stripping the leading v? we should use `lstrip()` instead

vulnerabilities/package_managers.py
```diff
@@ -327,15 +403,21 @@ def composer_url(pkg_name: str) -> str:
...
+                    release_date=parser.parse(resp["packages"][pkg_name][version]["time"]),
+                )
+            )
```
pombredanne, line 420: The comment below should be above the code it talks about, not after

vulnerabilities/package_managers.py
```diff
@@ -347,22 +429,51 @@ class GitHubTagsAPI(VersionAPI):
     package_type = "github"
 
     async def load_api(self, repo_set):
-        async with client_session() as session:
+        session = client_session()
+        async with session as session:
```
pombredanne, line 433: This code looks hard to read... we should have intermediate variable to avoid the nested complexity, possibly using generator expressions.

vulnerabilities/package_managers.py
```diff
@@ -347,22 +429,51 @@ class GitHubTagsAPI(VersionAPI):
...
-                    self.fetch(owner_repo.lower(), session)
+                    self.fetch(owner_repo.lower())
                     for owner_repo in repo_set
                     if owner_repo.lower() not in self.cache
                 ]
             )
 
-    async def fetch(self, owner_repo: str, session) -> None:
+    async def fetch(self, owner_repo: str, endpoint=None) -> None:
         # owner_repo is a string of format "{repo_owner}/{repo_name}"
```
pombredanne, line 443: Please create a proper docstring

vulnerabilities/package_managers.py
```diff
@@ -347,22 +429,51 @@ class GitHubTagsAPI(VersionAPI):
...
+        async with client_session() as session:
+            resp = await session.get(endpoint)
+            resp = await resp.read()
+
+        soup = BeautifulSoup(resp, features="lxml")
+        for release_entry in soup.find_all("div", {"class": "commit"}):
+            version = None
+            for links in release_entry.find_all("a"):
+                if f"/{owner_repo}/releases/tag/" in links["href"].lower():
+                    version = links["href"].split("/")[-1]
```
pombredanne, line 459: always prefer `partition` over split for robustness an clarity. For instance
```suggestion
                    _prefix, _slash, version = links["href"].rpartition("/")
```

vulnerabilities/package_managers.py
```diff
@@ -347,22 +429,51 @@ class GitHubTagsAPI(VersionAPI):
...
                 *[
-                    self.fetch(owner_repo.lower(), session)
+                    self.fetch(owner_repo.lower())
                     for owner_repo in repo_set
                     if owner_repo.lower() not in self.cache
                 ]
             )
 
-    async def fetch(self, owner_repo: str, session) -> None:
+    async def fetch(self, owner_repo: str, endpoint=None) -> None:
```
pombredanne, line 442: What about using a purl instead `owner_repo`

pombredanne (changes requested):
Thanks. See my comments inline for your consideration!

vulnerabilities/importers/apache_tomcat.py
```diff
@@ -62,7 +62,9 @@ def updated_advisories(self):
         return self.batch_advisories(advisories)
 
     def fetch_pages(self):
-        tomcat_major_versions = {i[0] for i in self.version_api.get("org.apache.tomcat:tomcat")}
+        tomcat_major_versions = {
+            i[0] for i in self.version_api.get("org.apache.tomcat:tomcat").valid_versions
```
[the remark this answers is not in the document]
sbs2001: fixing the i[0] thing, purl stuff would be done in another PR

vulnerabilities/package_managers.py
sbs2001: Tracking this in https://github.com/nexB/vulnerablecode/issues/492 . 

Could you elaborate an approach there ?

PR #476, opened 2021-06-10
Sort imports

Hritik14:
Sorting imports has been an unwritten coding style of our codebase. This PR makes it into a test rule.

There are some alternatives to this way of sorting 
- Use https://pypi.org/project/reorder-python-imports/
- Different multi-line-output mode https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes/#0-grid
- [force-single-line](https://pycqa.github.io/isort/docs/configuration/options/#force-single-line)


sbs2001:
Couple of things:
1. One import per line (@pombredanne  ^)
2. I dunno, but I  can see lot of imports which we don't use anymore, maybe we could eliminate those ? 
3. In any case this would be merged after we don't have so many  PRs, this could conflict with every other PR.

pombredanne:
@Hritik14 this is creating a merge "wall" for all other pending PR, so I would rather do this at once when all the pending PRs have been merged first.
Also we would want to add checks in the CI for this and black too.

pombredanne (approved):

PR #480, opened 2021-06-13
Remove old code and organize deps

sbs2001:


sbs2001:
@rolfschr  we are trying to seprate the dev deps and project deps into separate files. Could you modify the nix tests to install both of these ?

Hritik14:
Hey @sbs2001 
This commit removes most of the occurrences of `DJANGO_DEV`, yet does not remove all of them:
```bash
venv ❯ git grep -n "DJANGO_DEV"
.github/workflows/main.yml:56:          DJANGO_DEV: 1
.github/workflows/upstream_test.yml:52:          DJANGO_DEV: 1
Dockerfile:10:    DJANGO_DEV=1 python manage.py collectstatic
README.rst:269:    Environment="DJANGO_DEV=1"
docker-compose.yml:6:        - DJANGO_DEV=1
etc/nix/flake.nix:131:              export DJANGO_DEV=1
etc/nix/test-import-using-nix.sh:13:export DJANGO_DEV=${DJANGO_DEV:-1}
```

Further, as this commit totally removes enabling debug mode from the environment variables, I think we should ship with a development version of `vulnerablecode/settings.py` with DEBUG enabled as well.

pombredanne:
@Hritik14 Developers that know about debugging can enable it themselves. I do not see a point to make this easier. Instead a commented out setting is usually enough with a note.
Leaving some door open otherwise may not be a satisfying option and the benefits are low IMHO

sbs2001:
@Hritik14  yeah, seems like I missed some of em ;). 

@pombredanne  I don't think we need anything else other than django's built in `DEBUG=True`

Issue #492, opened 2021-07-03
Make package manager api code more readable


sbs2001:
See https://github.com/nexB/vulnerablecode/pull/467#discussion_r650844067

PR #495, opened 2021-07-03
Refactor package_managers

Hritik14:
Fixes: https://github.com/nexB/vulnerablecode/issues/492

Use uniform names and prototypes for `fetch` method.
Remove unnecessary checks. Re-use sessions.

I've removed `endpoint` argument from GitHubTagsAPI, I did not find a use for it at the current stage. Also, the `fetch` method is only called for the pkg which is not already in cache, so other checks appeared redundant. Please correct me if I am wrong.

---
Move load_api to parent class and refactor imports

The code for `load_api` was repetitive and used in all subclasses of
VersionAPI, it is better suited in the parent class. `fetch` method is
also now consistent and defined as an abstract method in the base class.
Python ABC is not used as different implementations of `load_api` and
`fetch` are allowed as done in DebianVersionAPI

### DIFF
--- a/vulnerabilities/package_managers.py
+++ b/vulnerabilities/package_managers.py
@@ -19,22 +19,20 @@
 #  for any legal advice.
 #  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
-
 import asyncio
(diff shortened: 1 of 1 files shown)


pombredanne (approved):
LGTM :+1:

PR #508, opened 2021-07-19
Use svn to collects tags in GitHubTagsAPI

Hritik14:
Surprisingly, GitHub allows svn requests to repositories. Now we can
have all the tags with a single request. This is much more efficient and
gentle to the APIs.

@pombredanne thank you so much for the suggestion


sbs2001 (approved):
This is much better

PR #523, opened 2021-08-02
Make upstream tests use makefile

Hritik14:
Refactor upstream tests to use makefile and standard installation
procedure. Also, it explicitly makes sure python version we test is
3.8.11


pombredanne (approved):
LGTM

PR #525, opened 2021-08-10
Separate import and improve operations

Hritik14:
## !! Breaking Changes !!

Entire database model is restructured.

- Get rid of added_advisory and batches
     - useless after a530627
- Introduce `improvers`
- See: https://github.com/nexB/vulnerablecode/pull/525#issuecomment-961778724 for a basic structure overview

Note: All changes in the importers are only reflected in the nginx importer while this PR is in draft stage.

### Knows defects (in current PR):
- [x] UI break (see above)
- [x] might crash in multiple imports / improves
- [x] No improver than default improver is implemented yet
- [x] normalized function of ``AdvisoryData`` has no body
- [x] nginx importer still has remains of set_api etc
- [x] Inference -> AdvisoryData encapsulation
- [x] Duplicated data in database
- [x] Drop migrations
- [x] ???

### Knows defects (to be solved in different PR):
- https://github.com/nexB/vulnerablecode/issues/593
- https://github.com/nexB/vulnerablecode/pull/476
- https://github.com/nexB/vulnerablecode/issues/277


Hritik14:
Feel free to review now. It's not perfect yet, so please consider these points
- Only the framework is there, the improvers/nginx.py improver doesn't actually improve anything. It needs to be written properly to fit this structure
- Many many names for patched/fixed/resolved packages are still there, please replace them in your mind, I'll refactor all of these in  one commit so as to not pollute current commit with irrelevant diffs
- There are a couple of fixmes, I'll try to figure them out, feel free to comment on those
- I have made no change in the importer logic yet (like refactoring importer_yielder, other potential improvements making it consistent with improvers), it is for the same reason that I don't want to pollute current commit. Current commit only focuses on improvers. I'll add a new commit after refactoring importers explicitly.
tl;dr : there's only improver framework (not improver itself), least possible irrelevant code is touched (although it needs to be refactored)

pombredanne:
@Hritik14 re:

> but then the importers would have to
mention that whatever they import have 100% confidence which might be
susceptible to typo errors making some importers not mention their
confidence thus zeroing on confidence. 

If you have a default of 100% this becomes a non-issue? I kinda like to avoid near-duplicated data structures.
In a way it makes sense to have a "100%" confidence in factual data we import... and there could be cases where we do not have 100% confidence in some imported factual data too?

pombredanne:
Note that using a common data structure does not mean that we have only importers.

vulnerabilities/models.py
```diff
@@ -203,6 +205,21 @@ class Meta:
...
+    vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
+    source = models.TextField(null=True)
+    confidence = models.PositiveIntegerField(default=1, validators=[MinValueValidator(1), MaxValueValidator(MAX_CONFIDENCE)])
+
+    def __str__(self):
+        return f"{self.package.package_url} {self.vulnerability.vulnerability_id}"
+
+    class Meta:
+        unique_together = ("package", "vulnerability")
+        verbose_name_plural = "PackageRelatedVulnerabilitiyFixes"
```
pombredanne, line 222: These are typically real names to use for display, not a camel case class name

pombredanne (changes requested):
Thank you!
... see some review comments inline.

vulnerabilities/data_inference.py
```diff
@@ -0,0 +1,37 @@
+import dataclasses
```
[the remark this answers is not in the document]
Hritik14: Noted. How about I add all the license in the last commit ? Might make things clearer. Let me know

vulnerabilities/data_source.py
```diff
@@ -88,14 +86,17 @@ class Advisory:
 
     summary: str
     vulnerability_id: Optional[str] = None
-    affected_packages: List[AffectedPackage] = dataclasses.field(default_factory=list)
+    affected_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
+    fixed_package_urls: Iterable[PackageURL] = dataclasses.field(default_factory=list)
```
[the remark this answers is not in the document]
Hritik14: We were not capturing version ranges earlier as well. Ranges were deflated into raw versions. As on https://github.com/nexB/vulnerablecode/blob/b6ef11d8c3552618f3ca4608ecd59611c6515df4/vulnerabilities/importers/nginx.py#L152-L191

Probably it will make sense to store version ranges _or_ concrete versions depending on the importer data source (as we don't want to tamper with the data source at all) . This will need further thought.

vulnerabilities/import_runner.py
```diff
@@ -107,84 +109,10 @@ def get_vuln_pkg_refs(vulnerability, package):
     )
 
 
-@transaction.atomic
-def process_advisories(data_source: DataSource) -> None:
```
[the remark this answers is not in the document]
Hritik14: I'll resolve this in favor of next comment.

vulnerabilities/importers/nginx.py
```diff
@@ -117,7 +118,8 @@ def to_advisories(self, data):
                 Advisory(
                     vulnerability_id=cve_id,
                     summary=summary,
-                    affected_packages=nearest_patched_package(vulnerable_packages, fixed_packages),
+                    affected_package_urls=vulnerable_packages,
```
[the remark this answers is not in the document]
Hritik14: Yes. Storing ranges v storing concrete versions should be separated and implemented (depending on the data source)

vulnerabilities/improvers/__init__.py
```diff
@@ -0,0 +1,11 @@
+from . import nginx
+
+IMPROVER_REGISTRY = [nginx.NginxTimeTravel]
+
+def find_class(class_name: str):
```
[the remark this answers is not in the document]
Hritik14: The strings are used for command line statements like `manage.py improve NginxTimeTravel`.
Perhaps having a str -> class mapping in the registry be better ? What do you think ?

vulnerabilities/improvers/nginx.py
```diff
@@ -0,0 +1,44 @@
+from packageurl import PackageURL
+
+from vulnerabilities.data_inference import Improver
+from vulnerabilities.data_inference import Advisory
+from vulnerabilities.data_inference import Inference
+from vulnerabilities.helpers import nearest_patched_package
+from vulnerabilities.models import Vulnerability
+from vulnerabilities.models import Package
+
+class NginxTimeTravel(Improver):
```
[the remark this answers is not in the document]
Hritik14: Sure. Though, this improver is highly experimental right now. I'll add the docstring.

vulnerabilities/improvers/nginx.py
```diff
@@ -0,0 +1,44 @@
...
+
+            time_traveller = nearest_patched_package(affected_packages, fixed_packages)
+            affected_packages = [ affected_package.vulnerable_package for affected_package in time_traveller]
+            fixed_packages = [ affected_package.patched_package for affected_package in time_traveller if affected_package.patched_package is not None]
+
+            inference = Inference(advisory = Advisory(
+                vulnerability_id=vulnerability.vulnerability_id,
+                summary=vulnerability.summary,
+                affected_package_urls=fixed_packages,
+            ), source="time travel", confidence=30)
```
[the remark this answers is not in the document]
Hritik14: Again, this is just a mock off. Calculating confidence would depend on the improver. I'll like to keep this for later.

vulnerabilities/improvers/nginx.py
```diff
@@ -0,0 +1,44 @@
...
+            inferences.append(inference)
+
+        return inferences
+
+
+def package_url(package):
```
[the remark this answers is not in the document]
Hritik14: tbh, I actually wanted to create an issue in package-url repo about this. Just delayed. 
This is a `Package` object, ultimately inherited from package-url. I need the conversion from a model returned `Package` object to a `PackageURL`. Let me know if there's already a better way.

vulnerabilities/management/commands/improve.py
```diff
@@ -0,0 +1,110 @@
...
+from vulnerabilities.models import Importer
+from vulnerabilities.import_runner import ImportRunner
+from vulnerabilities.importer_yielder import load_importers
+from vulnerabilities.improvers import IMPROVER_REGISTRY
+from vulnerabilities.improvers import find_class
+from vulnerabilities.improve_runner import ImproveRunner
+
+
+class Command(BaseCommand):
+    help = "Improve imported vulnerability data"
```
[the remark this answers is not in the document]
Hritik14: Fixed

vulnerabilities/management/commands/improve.py
```diff
@@ -0,0 +1,110 @@
...
+        parser.add_argument(
+            "--list",
+            action="store_true",
+            help="List available data inferences",
+        )
+        parser.add_argument(
+            "--all", action="store_true", help="Improve data from all available inferences"
+        )
+
+        parser.add_argument("sources", nargs="*", help="Data sources from which to import")
```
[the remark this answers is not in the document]
Hritik14: My mistake. Fixed

vulnerabilities/models.py
Hritik14: I could remove camel case with spaces... unless there's something better to go with ??

Hritik14:
@pombredanne 
> If you have a default of 100% this becomes a non-issue?

A typo could lead to inferences to have 100% confidence in that case... 

> I kinda like to avoid near-duplicated data structures

Importers and improvers look similar for now mostly because improvers don't really contain anything yet. There are various subclasess for importers like GitDataSource or OvalDataSource with corresponding helper functions to parse advisory data from an actual data source, whereas in case of improvers, they only get the entries from the existing data thus no advisory parsing would be required. Other than that an `Inference` could be seen as what we actually store in our system.

> In a way it makes sense to have a "100%" confidence in factual data we import... and there could be cases where we do not have 100% confidence in some imported factual data too

If it's a factual data then then by definition it's a fact thus 100% confidence, right?

vulnerabilities/models.py
sbs2001: nit: typo in  `PackageRelatedVulnerabilitiyFixes` -> `PackageRelatedVulnerabilityFixes`

sbs2001: Why is it not called `PackageRelatedVulnerabilitiyFix` ?

vulnerabilities/improvers/nginx.py
```diff
@@ -0,0 +1,44 @@
...
+from vulnerabilities.data_inference import Inference
+from vulnerabilities.helpers import nearest_patched_package
+from vulnerabilities.models import Vulnerability
+from vulnerabilities.models import Package
+
+class NginxTimeTravel(Improver):
+    def updated_inferences(self):
+        inferences = []
+
+        vulnerabilities = set(Vulnerability.objects.filter(vulnerable_packages__name="nginx"))
```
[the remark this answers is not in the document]
Hritik14: Please excuse the imperfection of the this particular improver.
> Only the framework is there, the improvers/nginx.py improver doesn't actually improve anything. It needs to be written properly to fit this structure

In the improver, there isn't the list of all packages to create a relationship, plus it's accessing the database directly plus a few more nits IMO, this is far from perfect right now.
About the weaker relationships, it's just not possible in this model. If there comes an Inference with a lower confidence, it is simply not updated in the database. 
Do have a look at ``improve_runner.py``
```py
def create_or_update_relation(relation, vulnerability, source, package, confidence):
    try:
        entry = relation.objects.get(
                vulnerability=vulnerability,
                package=package
                )
        if confidence > entry.confidence:
            entry.source = source
            entry.confidence = confidence
            entry.save()
        logger.debug("%s: Confidence improved for %s R %s, new confidence: %d", relation, package, vulnerability, confidence)

    except relation.DoesNotExist:
        relation.objects.create(
                vulnerability=vulnerability,
                source=source,
                package=package,
                confidence=confidence
                )

```
PS: Most probably we want to have the above function inside models.py and not here.

vulnerabilities/models.py
```diff
@@ -203,6 +205,21 @@ class Meta:
         unique_together = ("package", "vulnerability")
         verbose_name_plural = "PackageRelatedVulnerabilities"
 
+class PackageRelatedVulnerabilityFix(models.Model):
```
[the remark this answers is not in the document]
Hritik14: Can you point me where that is used ? 
Also, the names are not consistent yet, that'll come in a different PR as discussed in the gitter chat.

Hritik14:
Current commit brings breaking changes. Feel free to review but do ignore the TODO's and pointers mentioned below:
Knows defects:
- UI break (see above)
- inconsistent naming - will be resolved in a different PR
- might crash in multiple imports / improves
- No improver than default improver is implemented yet
- normalized function of ``AdvisoryData`` has no body
- nginx importer still has remains of set_api etc
- ???
(read more in commit message)

This is a step towards https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#meeting-on-tuesday-2021-08-17-at-1400-utc

Hritik14:
Trying out adding a real improver other than the default one now. It is not ready yet. Suggestions are welcome for the WIP commit but please avoid it in any strict review currently.

vulnerabilities/data_source.py
```diff
@@ -89,25 +115,43 @@ class Advisory:
     summary: str
     vulnerability_id: Optional[str] = None
     affected_packages: List[AffectedPackage] = dataclasses.field(default_factory=list)
+    fixed_packages: List[AffectedPackage] = dataclasses.field(default_factory=list)
```
pombredanne, line 118: We discussed the OSV design and fixed should be under affected_packages

vulnerabilities/data_source.py
```diff
@@ -89,25 +115,43 @@ class Advisory:
...
+        if isinstance(o, AffectedPackage):
+            return o.json()
+        if isinstance(o, Reference):
+            return vars(o)
+        if isinstance(o, datetime):
+            return o.isoformat()
+
+        return json.JSONEncoder.default(self, o)
+
+    def json(self):
```
pombredanne, line 135: What about instead something like this?
```Python
def to_dict(self):
    return {
        "summary": self.summary,
        "date_published": self.date_published.isoformat(),
        ...
    }
```

vulnerabilities/importers/nginx.py
```diff
@@ -178,19 +173,78 @@ def extract_vuln_pkgs(self, vuln_info):
...
         return [
-            PackageURL(type="generic", name="nginx", version=version, qualifiers=qualifiers)
-            for version in valid_versions
+            AffectedPackage(package=purl, version_specifier=version_range)
+            for version_range in version_ranges
         ]
 
 
+class NginxTimeTravel(Improver):
+    def inferences(self):
```
pombredanne, line 188: infer not inferences

vulnerabilities/importers/nginx.py
```diff
@@ -178,19 +173,78 @@ def extract_vuln_pkgs(self, vuln_info):
...
+            fixed_package_ranges = [
+                pkg.version_specifier for pkg in advisory_data.affected_packages
+            ]
+            fixed_packages = find_valid_versions(
+                self.version_api.get("nginx/nginx").valid_versions, fixed_package_ranges
+            )
+
+            pkgs = nearest_patched_package(affected_package_versions, fixed_package_ranges)
+            for pkg in pkgs:
+                print(pkg)
```
pombredanne, line 219: nit... you may want to guard these with a `if TRACE:` flag and keep them... otherwise drop them

pombredanne (changes requested):
Consider using `to_dict` rather than directly `to_json` ... a `to_json` becomes then a thin wrapper on the serialization with `to_dict`
See also my other comments inline.

vulnerabilities/data_source.py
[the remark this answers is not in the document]
Hritik14: New commits address this by using ``AffectedPackage`` class and having a VersionSpecifier

vulnerabilities/import_runner.py
```diff
@@ -107,84 +109,10 @@ def get_vuln_pkg_refs(vulnerability, package):
...
-def _package_url_to_package(purl: PackageURL) -> models.Package:
-    p = models.Package()
-    p.set_package_url(purl)
-    return p
+def process_advisories(source: str, advisories: Set[Advisory]) -> None:
+    """
+    Insert advisories into the database
+    Advisories are treated as full confidence infererences.
+    """
+    inferences = [ Inference(advisory, source, MAX_CONFIDENCE) for advisory in advisories ]
```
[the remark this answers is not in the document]
Hritik14: this structure has been revamped based on https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#meeting-on-tuesday-2021-08-17-at-1400-utc

vulnerabilities/importers/nginx.py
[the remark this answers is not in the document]
Hritik14: fixed by using ``AffectedPackage`` and ``VersionSpecifer``

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,132 @@
...
+
+    return models.Package.objects.get_or_create(**query_kwargs)
+
+
+def _package_url_to_package(purl: PackageURL) -> models.Package:
+    p = models.Package()
+    p.set_package_url(purl)
+    return p
+
+def create_or_update_relation(relation, vulnerability, source, package, confidence):
```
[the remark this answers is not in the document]
Hritik14: moved into models.py

vulnerabilities/improvers/__init__.py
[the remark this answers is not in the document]
Hritik14: Using fully qualified name prefixed with module names for this now.

vulnerabilities/data_source.py
```diff
@@ -74,12 +74,38 @@ def normalized(self):
         return Reference(reference_id=self.reference_id, url=self.url, severities=severities)
 
 
+@dataclasses.dataclass(order=True, frozen=True)
+class AffectedPackage:
```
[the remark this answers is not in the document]
Hritik14: Note to self: This also needs to be changed to something like:
```py
@dataclasses.dataclass(order=True, frozen=True)
class AffectedPackage:
    # this package MUST NOT have a version TODO: Enforce this
    package: PackageURL
    # the version specifier contains the version scheme as is: semver:>=1,3,4
    affected_version_specifier: VersionSpecifier
    fixed_version: str
```

vulnerabilities/data_inference.py
```diff
@@ -0,0 +1,55 @@
...
+    Only inferences with highest confidence for one vulnerability <-> package
+    relationship is to be inserted into the database
+    """
+
+    confidence: int
+    summary: Optional[str] = None
+    vulnerability_id: Optional[str] = None
+    affected_packages: List[PackageURL] = dataclasses.field(default_factory=list)
+    fixed_packages: List[PackageURL] = dataclasses.field(default_factory=list)
+    references: List[Reference] = dataclasses.field(default_factory=list)
```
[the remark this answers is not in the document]
Hritik14: We need to talk about when to use VULCOIDS

vulnerabilities/models.py
```diff
@@ -281,3 +317,13 @@ class VulnerabilitySeverity(models.Model):
 
     class Meta:
         unique_together = ("vulnerability", "reference", "scoring_system")
+
+
+class Advisory(models.Model):
+    date_published = models.DateField()
+    date_collected = models.DateField()
+    source = models.CharField(max_length=100)
+    improved_on = models.DateTimeField(null=True)
```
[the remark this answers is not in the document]
Hritik14: I'll go with ``date_improved``

vulnerabilities/data_source.py
Hritik14: Done. Thanks

vulnerabilities/importers/nginx.py
Hritik14: thanks!

vulnerabilities/data_source.py
Hritik14: Done!

vulnerabilities/importers/nginx.py
Hritik14: changed

vulnerabilities/data_source.py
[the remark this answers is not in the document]
Hritik14: Done

Hritik14:
Using the OSV design has some caveats. Given [data like](https://nginx.org/en/security_advisories.html):
```
Not vulnerable: 1.21.0+, 1.20.1+
Vulnerable: 0.6.18-1.20.0
```
It is impossible to directly get _one_ fixed version for a range of affected versions.

pombredanne:
@Hritik14 re: 

> Using the OSV design has some caveats. Given [data like](https://nginx.org/en/security_advisories.html):
> 
> ```
> Not vulnerable: 1.21.0+, 1.20.1+
> Vulnerable: 0.6.18-1.20.0
> ```
> It is impossible to directly get _one_ fixed version for a range of affected versions.

Is this a real case rather than some made up example?
I cannot fathom how you could get the two versions fixing the same bug.
In the example you gave, `1.21.0+` is impossible and `1.20.1+` can only be the correct one as this is the latest one.

Now things would be different if there were more than one range of course... but then that would be something entirely different?

pombredanne:
Here is the gist of our discussion:

1. nginx advisories are weird at best and likely not meant to be strictly parsable.
2. the corresponding CVEs CPE ranges seem to be also problematic in some cases
3. we should reach out to nginx to clarify
4. in the meantime and in general, where there are ambiguities like this we 
 4.1. make a safe judgment call in this case use the latest versions of multiple "fixed" version that apply to the same range. We may be overly cautious and there could be some configs that are not vulnerable that we may report as vulnerable in some cases
 4.2. log this and flag these for manual review

vulnerabilities/data_inference.py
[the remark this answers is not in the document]
Hritik14: Create a ticket to replace VULCOIDS by CVE if there already was a VULCOID and a new CVE has been found.
Plus we need to think about where do we want to generate vulcoids

Hritik14:
Turns out we can parse nginx versions smoothly. Here's more on it https://github.com/nexB/vulnerablecode/issues/553

vulnerabilities/improve_runner.py
```diff
@@ -77,9 +79,7 @@ def process_inferences(source: str, inferences: Set[Inference]):
     )
 
 
-def _get_or_create_vulnerability(
-    vulnerability_id, summary
-) -> Tuple[models.Vulnerability, bool]:
+def _get_or_create_vulnerability(vulnerability_id, summary) -> Tuple[models.Vulnerability, bool]:
```
pombredanne, line 82: Why this underscore prefix?

vulnerabilities/improvers/default.py
```diff
@@ -11,14 +12,15 @@
 from vulnerabilities.data_inference import MAX_CONFIDENCE
 from vulnerabilities.models import Advisory
 
+
 class DefaultImprover(Improver):
     def inferences(self) -> List[Inference]:
         advisories = Advisory.objects.all()
```
pombredanne, line 18: Add a docstring

pombredanne:
I do not like the batch processing of multiple Advisory at once in a single transaction... this may be an optimization we will need in the future, but not now.

Here is a suggested updated way:

1. each `Improver` has a method to process a single Advisory model instance such as `Improver.get_inferences(self, advisory): -> Inference` or something like that.

2. the framework then iterates on an `Improver`-provided query set such as `Improver.get_interesting_advisories(self): -> QuerySet` that has the `Advisories` it is interested in.
 
3. in the framework, there is an **atomic transaction** that updates both the Advisory (e.g. date and later a log of improvements with select for update) and whatever is updated or create from the `Inference`

vulnerabilities/models.py
```diff
@@ -281,3 +317,22 @@ class VulnerabilitySeverity(models.Model):
...
+    date_published = models.DateField(help_text="UTC Date of publication of the advisory")
+    date_collected = models.DateField(help_text="UTC Date on which the advisory was collected")
+    source = models.CharField(
+        max_length=100,
+        help_text="Fully qualified name of the importer prefixed with the module name importing the advisory. Eg: vulnerabilities.importers.nginx.NginxDataSource",
+    )
+    date_improved = models.DateTimeField(
+        null=True, help_text="Latest date on which the advisory was improved by an improver"
+    )
+    data = models.JSONField(
```
pombredanne, line 336:
```suggestion
    # we use a JSON field here to avoid creating a complete relational model for data that
    # is never queried directly; instead it is only retrieved and processed as a whole by
    # an improver 
    data = models.JSONField(
```

Hritik14:
The proposed structure is 
```
                                                      ┌────────────┐
┌──────────────────────┐                              │            │                                          ┌─────────────────────┐
│                      │                              │  Database  │ Has version ranges                       │                     │
│                      ├──────────────────────────────┤    in      │                                          │                     │
│                      │                              │  Advisory  │                                          │                     │
│                      │                              │    Model   │                                          │                     │
│                      │                              │            │                                          │   Frontend          │
│                      │                              ├────────────┘                                          │                     │
│        Importers     │                              │  As true                                              │                     │
│                      │                              │  as upstream                                          │                     │
│                      │                              │                                                       │                     │
│                      │                              │                                                       │                     │
│                      │                              │                                                       │                     │
│                      │                              │                                                       │                     │
└──────────────────────┘                              │                               ┌──────────────────────►│                     │
                                                      │                               │                       └─────────────────────┘
                                                      │                               │
                                                      │                               │
                                                      │                               │
                                                      │                               │
  ┌───────────────────────┐                           │                               │
  │                       │                           │                               │
  │                       │◄──────────────────────────┘                               │
  │                       │                                                           │
  │                       │                                                           │
  │                       │                                          ┌──────────────┐ │
  │    Improvers          │                                          │              │ │
  │                       │                                          │  Vuln id     │ │
  │     - DefaultImprover ├─────────────────────────────────────────►├──────────────┤ │
  │                       │                                          │              ├─┘
  │     - TimeTravel      │                                          │  Aff pkgs    │
  │                       │                                          ├──────────────┤
  │     - ...             │                                          │              │
  │                       │                                          │  Fixed pkgs  │
  └───────────────────────┘                                          ├──────────────┤
                                                                     │  Mapp & conf │
                                                                     └──────────────┘
                                                                             ▲
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                                                             │
                                         Only reads from Advisory model      │
                                                                             │
      ┌─────────────────────────┐        VersionRanges -> Concrete ranges    │
      │                         │                                            │
      │                         │ ───────────────────────────────────────────┘
      │                         │
      │                         │
      │                         │
      │   DefaultImprover       │
      │                         │
      │                         │
      │                         │
      │                         │
      └─────────────────────────┘
```

vulnerabilities/importers/nginx.py
```diff
@@ -106,89 +109,103 @@ def to_advisories(self, data):
...
+    versions = []
+    for rng in raw_ranges:
+        # Eg. "1.7.3+" gets converted to SemVersion(1.7.3)
+        # The way this needs to be interpreted is unique for nginx advisories
+        # More: https://github.com/nexB/vulnerablecode/issues/553
 
-        valid_versions = find_valid_versions(
-            self.version_api.get("nginx/nginx").valid_versions, version_ranges
-        )
+        versions.append(SemverVersion(rng.partition("+")[0].strip()))
```
pombredanne, line 157:
```suggestion
        # TODO: create a version scheme that's specific to nginx... because this is not exactly semver
        versions.append(SemverVersion(rng.partition("+")[0].strip()))
```

vulnerabilities/models.py
```diff
@@ -77,15 +89,15 @@ def vulnerable_to(self):
         """
         Returns packages which are vulnerable to this vulnerability.
```
pombredanne, line 90:
```suggestion
        Return packages that are vulnerable to this vulnerability.
```

vulnerabilities/models.py
```diff
@@ -122,33 +134,29 @@ class Package(PackageURLMixin):
...
-    resolved_vulnerabilities = models.ManyToManyField(
-        to="Vulnerability",
-        through="PackageRelatedVulnerability",
-        through_fields=("patched_package", "vulnerability"),
-        related_name="patched_packages",
+        related_name="packages",
     )
 
     @property
     def vulnerable_to(self):
```
pombredanne, line 148:
```suggestion
    # TODO: consider renaming to "affected_by"
    def vulnerable_to(self):
```

vulnerabilities/models.py
```diff
@@ -122,33 +134,29 @@ class Package(PackageURLMixin):
...
         """
         Returns vulnerabilities which are affecting this package.
         """
-        return self.vulnerabilities.all()
+        return self.vulnerabilities.filter(packagerelatedvulnerability__fix=False)
 
     @property
     def resolved_to(self):
```
pombredanne, line 155:
```suggestion
    # TODO: consider renaming to "fixes" or "fixing" ? (TBD) and updating the docstring
    def resolved_to(self):
```

vulnerabilities/models.py
```diff
@@ -188,13 +196,13 @@ def __str__(self):
...
-    package = models.ForeignKey(
-        Package, on_delete=models.CASCADE, related_name="vulnerable_package"
-    )
+    package = models.ForeignKey(Package, on_delete=models.CASCADE, related_name="package")
     vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
-    patched_package = models.ForeignKey(
-        Package, on_delete=models.CASCADE, null=True, blank=True, related_name="patched_package"
+    source = models.TextField(null=True)
+    confidence = models.PositiveIntegerField(
+        default=1, validators=[MinValueValidator(0), MaxValueValidator(MAX_CONFIDENCE)]
```
pombredanne, line 203: If you have a default, please align this with the "Inference" data class... I also wonder if having a default here makes sense at all? And if a default is needed why 1? why not 0? or max? or 42?
Help text would be nice too.

vulnerabilities/models.py
```diff
@@ -188,13 +196,13 @@ def __str__(self):
...
 class PackageRelatedVulnerability(models.Model):
 
-    package = models.ForeignKey(
-        Package, on_delete=models.CASCADE, related_name="vulnerable_package"
-    )
+    package = models.ForeignKey(Package, on_delete=models.CASCADE, related_name="package")
     vulnerability = models.ForeignKey(Vulnerability, on_delete=models.CASCADE)
-    patched_package = models.ForeignKey(
-        Package, on_delete=models.CASCADE, null=True, blank=True, related_name="patched_package"
+    source = models.TextField(null=True)
```
pombredanne, line 201: Please add help to each field. See below. Also we may want to have a charfield instead?
Please also check the use of blank and null... to be consistent with other Text fields

vulnerabilities/models.py
```diff
@@ -188,13 +196,13 @@ def __str__(self):
...
+    confidence = models.PositiveIntegerField(
+        default=1, validators=[MinValueValidator(0), MaxValueValidator(MAX_CONFIDENCE)]
     )
+    fix = models.BooleanField(default=False)
```
pombredanne, line 205: help text would be more than nice.

vulnerabilities/models.py
```diff
@@ -203,6 +211,36 @@ class Meta:
...
+        try:
+            existing = self.__class__.objects.get(
+                vulnerability=self.vulnerability, package=self.package
+            )
+            if self.confidence > existing.confidence:
+                existing.source = self.source
+                existing.confidence = self.confidence
+                existing.fix = self.fix
+                existing.save()
+            logger.debug(
```
pombredanne, line 228:
```suggestion
            # TODO: later we want these to be part of a log field in the DB
            logger.debug(
```

vulnerabilities/models.py
```diff
@@ -281,3 +317,22 @@ class VulnerabilitySeverity(models.Model):
...
+
+
+class Advisory(models.Model):
+    """
+    An advisory directly obtained from upstream without any modifications.
+    """
+
+    date_published = models.DateField(help_text="UTC Date of publication of the advisory")
+    date_collected = models.DateField(help_text="UTC Date on which the advisory was collected")
+    source = models.CharField(
```
pombredanne, line 329: source may not be the best name. See also above another comments wrt. source naming https://github.com/nexB/vulnerablecode/pull/525/files#diff-7f9f2c92e7163b06d21fa139369d75caed6561d0368b60ea2516cece0220eb5bR201

vulnerabilities/views.py
```diff
@@ -73,8 +74,14 @@ def request_to_queryset(request):
             models.Package.objects.all()
             .filter(name__icontains=package_name, type__icontains=package_type)
             .annotate(
-                vulnerability_count=Count("vulnerabilities"),
-                patched_vulnerability_count=Count("resolved_vulnerabilities"),
+                vulnerability_count=Count(
+                    "vulnerabilities",
+                    filter=Q(vulnerabilities__packagerelatedvulnerability__fix=False),
+                ),
+                patched_vulnerability_count=Count(
```
pombredanne, line 81:
```suggestion
                #TODO: consider renaming to fixed in the future
                patched_vulnerability_count=Count(
```

pombredanne (changes requested):
See some nits for your consideration.

vulnerabilities/import_runner.py
```diff
@@ -107,84 +109,10 @@ def get_vuln_pkg_refs(vulnerability, package):
...
-
-
-def _package_url_to_package(purl: PackageURL) -> models.Package:
-    p = models.Package()
-    p.set_package_url(purl)
-    return p
+def process_advisories(source: str, advisories: Set[Advisory]) -> None:
+    """
+    Insert advisories into the database
+    Advisories are treated as full confidence infererences.
```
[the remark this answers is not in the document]
Hritik14: Averted in new code

vulnerabilities/data_source.py
```diff
@@ -74,12 +76,65 @@ def normalized(self):
...
+            dct["fixed_version"] = str(self.fixed_version)
+        return dct
+
+    @staticmethod
+    def from_dict(aff_pkg: dict):
+        package = PackageURL(*aff_pkg["package"])
+        affected_version_specifier = VersionSpecifier.from_version_spec_string(
+            aff_pkg["affected_version_specifier"]
+        )
+        if "fixed_version" in aff_pkg:
```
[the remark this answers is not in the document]
Hritik14: nexb/univers#9 is now merged

vulnerabilities/importers/nginx.py
```diff
@@ -106,50 +96,58 @@ def to_advisories(self, data):
...
         raw_ranges = version_info.split(",")
-        version_ranges = []
+        versions = []
         for rng in raw_ranges:
-            # Eg. "1.7.3+" gets converted to VersionSpecifier.from_scheme_version_spec_string("semver","^1.7.3")
-            # The advisory in this case uses `+` in the sense that any version
-            # with greater or equal `minor` version satisfies the range.
-            # "1.7.4" satisifes "1.7.3+", but "1.8.4" does not. "1.7.3+" has same
-            # semantics as that of "^1.7.3"
+            # Eg. "1.7.3+" gets converted to SemVersion(1.7.3)
```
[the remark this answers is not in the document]
Hritik14: Now we have version handling properly done in univers

vulnerabilities/importers/nginx.py
```diff
@@ -106,50 +96,58 @@ def to_advisories(self, data):
...
+            # More: https://github.com/nexB/vulnerablecode/issues/553
 
-            version_ranges.append(
-                VersionSpecifier.from_scheme_version_spec_string("semver", "^" + rng[:-1])
-            )
-
-        valid_versions = find_valid_versions(
-            self.version_api.get("nginx/nginx").valid_versions, version_ranges
-        )
+            versions.append(SemverVersion(rng[:-1].strip()))
```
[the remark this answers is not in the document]
Hritik14: This will be handled in univers

vulnerabilities/data_inference.py
```diff
@@ -0,0 +1,70 @@
...
+class Inference:
+    """
+    This data class expresses the contract between data improvers and the improve runner.
+
+    Only inferences with highest confidence for one vulnerability <-> package
+    relationship is to be inserted into the database
+    """
+
+    vulnerability_id: str
+    confidence: int
```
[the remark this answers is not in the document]
Hritik14: Done

vulnerabilities/improvers/default.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Only AffectedPackages with an equality in their VersionSpecifier are
+    considered as exact purls.
+
+    For eg:
+    AffectedPackage with version_specifier as scheme:<=2.0 is treated as
+    version 2.0 but the same with scheme:<2.0 is not considered at all as there
+    is no info about what comes before the supplied version
+
+    Return a list of affected PackageURL and corresponding fixed PackageURL
```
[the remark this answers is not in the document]
Hritik14: Done

vulnerabilities/models.py
Hritik14: Done. It defaults to `MAX_CONFIDENCE` now, also there's a help text for `source`

vulnerabilities/models.py
Hritik14: Added and `blank`ed

vulnerabilities/models.py
Hritik14: Added

vulnerabilities/models.py
Hritik14: renamed to `created_by`

vulnerabilities/improvers/default.py
```diff
@@ -0,0 +1,66 @@
...
+                    fixed_purls=[fixed_purl],
+                    references=advisory_data.references,
+                )
+            )
+        return inferences
+
+
+def exact_purls(aff_pkg: AffectedPackage) -> (List[PackageURL], PackageURL):
+    """
+    Only AffectedPackages with an equality in their VersionSpecifier are
```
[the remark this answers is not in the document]
Hritik14: renamed to `get_exact_purls`

vulnerabilities/models.py
```diff
@@ -327,21 +334,36 @@ class Meta:
...
     """
 
-    date_published = models.DateField(help_text="UTC Date of publication of the advisory")
+    # TODO: Generate vulcoids for null vulnerablity id
+    vulnerability_id = models.CharField(max_length=50, null=True, blank=True)
+    summary = models.TextField(blank=True, null=True)
+    # we use a Text field here to avoid creating a complete relational model for data that
+    # is never queried directly; instead it is only retrieved and processed as a whole by
+    # an improver
+    affected_packages = models.TextField()
```
pombredanne, line 346: Let's use a JSONfield instead

vulnerabilities/data_inference.py
```diff
@@ -0,0 +1,97 @@
...
+@dataclasses.dataclass(order=True)
+class Inference:
+    """
+    This data class expresses the contract between data improvers and the improve runner.
+
+    Only inferences with highest confidence for one vulnerability <-> package
+    relationship is to be inserted into the database
+    """
+
+    vulnerability_id: str = None
```
Hritik14, line 27: Make sure it's VULCOID

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+        return
+
+    logger.info(f"Improving advisory id: {advisory.id}")
+
+    for inference in inferences:
+        vuln = get_or_create_vulnerability_and_aliases(
+            inference.vulnerability_id, inference.aliases, inference.summary
+        )
+        if not vuln:
+            continue
```
Hritik14, line 61: log something

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            existing_aliases.add(alias)
+            if alias.vulnerability:
+                existing_vulns.add(alias.vulnerability)
+
+    # If given set of aliases point to different vulnerabilities in the
+    # database, request is malformed
+    # TODO: It is possible that all those vulnerabilities are actually
+    # the same at data level, figure out a way to merge them
+    if len(existing_vulns) > 1:
```
Hritik14, line 145: Add more information in logging and do not explicitly return None

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+        logger.warn(
+            f"Given aliases {existing_aliases} already exist and do not point "
+            "to a single vulnerability. Cannot improve. Skipped."
+        )
+        return None
+
+    existing_alias_vuln = existing_vulns.pop() if existing_vulns else None
```
Hritik14, line 152: pop(none)

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+            return vuln
+
+        if existing_alias_vuln.vulnerability_id == vulnerability_id:
+            # TODO: What to do with the new summary ?
+            for alias in new_aliases:
+                alias.vulnerability = existing_alias_vuln
+                alias.save()
+            return existing_alias_vuln
+
+        logger.warn(
```
Hritik14, line 173: Bail out early

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+        return None
+
+    existing_alias_vuln = existing_vulns.pop() if existing_vulns else None
+
+    # If we have been supplied with a vulnerability_id and existing aliases do
+    # not have vulnerability_id then create one for all aliases, otherwise use
+    # the vulnerability_id from the existing aliases
+    if vulnerability_id:
+        if not existing_alias_vuln:
+            vuln = models.Vulnerability(summary=summary)
```
Hritik14, line 159: Where's the id ?

vulnerabilities/improve_runner.py
```diff
@@ -0,0 +1,198 @@
...
+    if len(existing_vulns) > 1:
+        logger.warn(
+            f"Given aliases {existing_aliases} already exist and do not point "
+            "to a single vulnerability. Cannot improve. Skipped."
+        )
+        return None
+
+    existing_alias_vuln = existing_vulns.pop() if existing_vulns else None
+
+    # If we have been supplied with a vulnerability_id and existing aliases do
```
Hritik14, line 154:
```py
if existing_alias_vuln and vulnerability_id and existing_alias_vuln.vulnerability_id != vulnerability_id:
    ..... log error and return

if existing_alias_vuln:
    vulnerability = existing_alias_vuln
elif vulnerability_id:
    vulnerability = models.Vulnerability.objects.get(vulnerability_id=vulnerability_id)
    ... if does not exist... there is a problem to log
else:
    .... create a new vulnerability


# here I have vulnerability 
```

Hritik14:
@pombredanne We have `Confidence 90` in NginxBasicImprover, what could be a better confidence ?

vulnerabilities/improvers/default.py
```diff
@@ -0,0 +1,69 @@
...
+    >>> vers = "vers:npm/<1.0.0 | >=2.0.0 | <3.0.0"
+    >>> affected_package = AffectedPackage.from_dict({
+    ...     "package": purl,
+    ...     "affected_version_range": vers,
+    ...     "fixed_version": "5.0.0"
+    ... })
+    >>> get_exact_purls(affected_package)
+    ([PackageURL(type='turtle', namespace=None, name='green', version='2.0.0', qualifiers={}, subpath=None)], PackageURL(type='turtle', namespace=None, name='green', version='5.0.0', qualifiers={}, subpath=None))
+    """
+    affected_purls = set()
```
pombredanne, line 58: IMHO we could refactor this as version resolution procedure using this kind of logic:
```
>>> from univers import *
>>> from univers.version_range import *
>>> VersionRange.from_string("vers:npm/<1.0.0 | >=2.0.0 | <3.0.0")
NpmVersionRange(constraints=(VersionConstraint(comparator='<', version=SemverVersion(string='1.0.0')), VersionConstraint(comparator='>=', version=SemverVersion(string='2.0.0')), VersionConstraint(comparator='<', version=SemverVersion(string='3.0.0'))))
>>> vr= VersionRange.from_string("vers:npm/<1.0.0 | >=2.0.0 | <3.0.0")
>>> range_versions = [c.version for c in vr.constraints]
>>> range_versions
[SemverVersion(string='1.0.0'), SemverVersion(string='2.0.0'), SemverVersion(string='3.0.0')]
>>> resolved_versions = [v for v in range_versions if v in vr]
>>> resolved_versions
[SemverVersion(string='2.0.0')]
>>> 
```

vulnerabilities/management/commands/improve.py
```diff
@@ -0,0 +1,106 @@
...
+
+        sources = options["sources"]
+        if not sources:
+            raise CommandError('Please provide at least one improver to run use "--all".')
+
+        self.improve_data(valid_sources(sources))
+
+    def list_sources(self):
+        improvers = [
+            f"{improver.__module__}.{improver.__qualname__}" for improver in IMPROVER_REGISTRY
```
pombredanne, line 69: We likely have code that already deal with fqn....

vulnerabilities/management/commands/improve.py
```diff
@@ -0,0 +1,106 @@
...
+        ]
+        self.stdout.write("Vulnerability data can be processed by these available improvers:")
```
pombredanne, line 71:
```suggestion
        self.stdout.write("Vulnerability data can be processed by these available improvers:\n")
```

vulnerabilities/views.py
```diff
@@ -102,8 +110,12 @@ def request_to_vulnerabilities(request):
         vuln_id = request.GET["vuln_id"]
         return list(
             models.Vulnerability.objects.filter(vulnerability_id__icontains=vuln_id).annotate(
-                vulnerable_package_count=Count("vulnerable_packages"),
-                patched_package_count=Count("patched_packages"),
+                vulnerable_package_count=Count(
+                    "packages", filter=Q(packagerelatedvulnerability__fix=False)
```
pombredanne, line 114: The query for affected and fixed packages should be abstracted as methods on the models (and/or on a model manager)

vulnerabilities/models.py
```diff
@@ -39,56 +52,41 @@ class Vulnerability(models.Model):
     VulnerabilityReference.
```
pombredanne, line 52: FIXME: the docstring is incorrect now

vulnerabilities/models.py
```diff
@@ -101,69 +99,77 @@ class VulnerabilityReference(models.Model):
...
-    url = models.URLField(max_length=1024, help_text="URL of Vulnerability data", blank=True)
 
     @property
-    def scores(self):
+    def severities(self):
         return VulnerabilitySeverity.objects.filter(reference=self.id)
 
     class Meta:
-        unique_together = ("vulnerability", "reference_id", "url")
+        unique_together = ("vulnerability", "url", "reference_id")
```
pombredanne, line 117:
```suggestion
        unique_together = ("vulnerability", "url", "reference_id",)
```

vulnerabilities/models.py
```diff
@@ -202,11 +221,40 @@ def __str__(self):
...
-class ImportProblem(models.Model):
-
-    conflicting_model = models.JSONField()
+    def update_or_create(self):
+        """
+        Update if supplied record has more confidence than existing record
+        Create if doesn't exist
+        """
+        try:
+            existing = self.__class__.objects.get(
```
pombredanne, line 232: Why the `__class__` here?
```suggestion
            existing = PackageRelatedVulnerability.objects.get(
```

vulnerabilities/models.py
```diff
@@ -202,11 +221,40 @@ def __str__(self):
...
+                existing.fix = self.fix
+                existing.save()
+                # TODO: later we want these to be part of a log field in the DB
+                logger.info(
+                    f"Confidence improved for {self.package} R {self.vulnerability}, "
+                    f"new confidence: {self.confidence}"
+                )
+
+        except self.DoesNotExist:
+            self.__class__.objects.create(
```
pombredanne, line 247:
```suggestion
            PackageRelatedVulnerability.objects.create(
```

pombredanne (approved):
Ready to merge! ... see a few nits for your review.

PR #546, opened 2021-08-29
Fix several problems with the nix setup.

rolfschr:
This PR fixes several issues with the Nix setup. Note that the GitHub actions still doesn't run. I don't know what exactly the problem is but have reported it upstream (https://github.com/DavHau/mach-nix/issues/318).

Signed-off-by: Rolf Schröder <rolf.schr@gmail.com>

### DIFF
--- a/etc/nix/flake.nix
+++ b/etc/nix/flake.nix
@@ -61,6 +61,20 @@
           pypiDataSha256 =
             "0499zl39aia74f0i7fkn5dsy8244dkmcw4vzd5nf4kai605j2jli";
         });
+      # This wrapper allows to setup both the production as well as the
+      # development Python environments in the same way (albeit having
+      # different requirements.txt).
+      getPythonEnv = system: requirements:
+        machnixFor.${system}.mkPython {
+          requirements = ''
+            ${requirements}
+          '';
+          # Fix an issue with an upstream dep of GitPython.
+          # https://github.com/DavHau/mach-nix/issues/287
+          # See https://github.com/DavHau/mach-nix/issues/318
+          _.gitpython.propagatedBuildInputs.mod = pySelf: self: oldVal:
+            oldVal ++ [ pySelf.typing-extensions ];
+        };
 
     in {
 
@@ -68,11 +82,7 @@
       overlay = final: prev:
         with final.pkgs; {
 
-          pythonEnv = machnixFor.${system}.mkPython {
-            requirements = ''
-              ${requirements}
-            '';
-          };
+          pythonEnv = getPythonEnv system requirements;
 
           vulnerablecode = stdenv.mkDerivation {
             inherit version;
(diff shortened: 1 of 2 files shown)


etc/nix/flake.nix
```diff
@@ -78,13 +78,13 @@
...
             name = "vulnerablecode-${version}";
             src = vulnerablecode-src;
-            dontConfigure = true; # do not use ./configure
+            dontBuild = true; # do not use Makefile
             propagatedBuildInputs = [ pythonEnv postgresql gitMinimal ];
 
             postPatch = ''
-              # Make sure the pycodestyle binary in $PATH is used.
-              substituteInPlace vulnerabilities/tests/test_basics.py \
-                --replace 'join(bin_dir, "pycodestyle")' '"pycodestyle"'
```
rolfschr, line 87: `pycodestyle` is not used anymore.

etc/nix/flake.nix
```diff
@@ -78,13 +78,13 @@
             inherit version;
             name = "vulnerablecode-${version}";
             src = vulnerablecode-src;
-            dontConfigure = true; # do not use ./configure
+            dontBuild = true; # do not use Makefile
```
rolfschr, line 81: `configure` doesn't exist anymore, `Makefile` was added.

etc/nix/flake.nix
```diff
@@ -138,6 +138,8 @@
               buildPhase = ''
                 source ${libSh}
                 initPostgres $(pwd)
+                export SECRET_KEY=REALLY_SECRET
```
rolfschr, line 141: ce64bceda7cde0a9 removed the `DJANGO_DEV=1` setting, I would guess without testing it: Setting neither `DJANGO_DEV=1` nor the `SECRET_KEY` results in an error.

Hritik14: Yes, ``DJANGO_DEV`` is not present now. Please use ``make envfile``. A basic make based run is at https://github.com/nexB/vulnerablecode/blob/main/.github/workflows/main.yml

rolfschr: @Hritik14 Thanks, if I understand you correctly, you suggest to run ``make dev envfile`` (https://github.com/nexB/vulnerablecode/blob/main/.github/workflows/main.yml#L24). The ``dev`` target makes no sense since the Nix setup  doesn't run ``pip -r requirements`` on purpose. As for the ``envfile`` target, it creates the file but uses the non-deterministic ``urandom`` source, which is not feasible in Nix. I'm not convinced it is better to use make than to simply set the env var as I proposed here.

sbs2001:
@rolfschr  is this good to merge ?

rolfschr:
Yes, I think so.

sbs2001:
@rolfschr  Could you rebase please ?

sbs2001:
): now tests fail

sbs2001 (approved):

Issue #466, opened 2021-05-30
Collect go vulnerabilities from github api


sbs2001:
Very recently Github's graphql API started providing for vulnerabilities for go packages. We need to ingest this as part of the github importer

There's also very recent https://github.com/golang/vulndb

sify21:
Hope this comes soon. Following code shows how golang fetch module versions through a proxy:
https://github.com/golang/go/blob/master/src/cmd/go/internal/modfetch/proxy.go

sify21:
I made a pull request for this

PR #578, opened 2021-11-17
Collect go vulnerabilities from github api

sify21:
try to fix https://github.com/nexB/vulnerablecode/issues/466

### DIFF
--- a/vulnerabilities/importers/github.py
+++ b/vulnerabilities/importers/github.py
@@ -44,6 +44,7 @@ from vulnerabilities.package_managers import MavenVersionAPI
 from vulnerabilities.package_managers import NugetVersionAPI
 from vulnerabilities.package_managers import ComposerVersionAPI
 from vulnerabilities.package_managers import PypiVersionAPI
+from vulnerabilities.package_managers import GoproxyVersionAPI
 from vulnerabilities.package_managers import RubyVersionAPI
 from vulnerabilities.severity_systems import scoring_systems
 from vulnerabilities.helpers import nearest_patched_package
@@ -206,6 +207,7 @@ class GitHubAPIDataSource(DataSource):
             "COMPOSER": ComposerVersionAPI,
             "PIP": PypiVersionAPI,
             "RUBYGEMS": RubyVersionAPI,
+            "GO": GoproxyVersionAPI,
         }
         versioner = versioners.get(ecosystem)
         if versioner:
@@ -229,7 +231,7 @@ class GitHubAPIDataSource(DataSource):
                 return None
             return vendor, name
 
-        if ecosystem == "NUGET" or ecosystem == "PIP" or ecosystem == "RUBYGEMS":
+        if ecosystem == "NUGET" or ecosystem == "PIP" or ecosystem == "RUBYGEMS" or ecosystem == "GO":
             return None, pkg_name
 
     @staticmethod
@@ -265,6 +267,8 @@ class GitHubAPIDataSource(DataSource):
                     unaffected_purls = []
                     if self.process_name(ecosystem, name):
                         ns, pkg_name = self.process_name(ecosystem, name)
+                        if hasattr(self.version_api, "pkg_mappings"):
+                            pkg_name = self.version_api.pkg_mappings.get(name, pkg_name)
                         aff_range = adv["node"]["vulnerableVersionRange"]
                         aff_vers, unaff_vers = self.categorize_versions(
                             self.version_api.package_type,
--- a/vulnerabilities/package_managers.py
+++ b/vulnerabilities/package_managers.py
@@ -26,9 +26,7 @@ import xml.etree.ElementTree as ET
 from datetime import datetime
 from json import JSONDecodeError
 from subprocess import check_output
-from typing import List
-from typing import Mapping
-from typing import Set
+from typing import Set, List, MutableMapping, Optional
 
 from aiohttp import ClientSession
 import aiohttp
@@ -41,7 +39,7 @@ from dateutil import parser as dateparser
 @dataclasses.dataclass(frozen=True)
 class Version:
     value: str
-    release_date: datetime = None
+    release_date: Optional[datetime] = None
 
 
 @dataclasses.dataclass
@@ -55,10 +53,10 @@ class GraphQLError(Exception):
 
 
 class VersionAPI:
-    def __init__(self, cache: Mapping[str, Set[Version]] = None):
+    def __init__(self, cache: MutableMapping[str, Set[Version]] = None):
         self.cache = cache or {}
 
-    def get(self, package_name, until=None) -> Set[str]:
+    def get(self, package_name, until=None) -> VersionResponse:
         new_versions = set()
         valid_versions = set()
         for version in self.cache.get(package_name, set()):
@@ -67,7 +65,9 @@ class VersionAPI:
                 continue
             valid_versions.add(version.value)
 
-        return VersionResponse(valid_versions=valid_versions, newer_versions=new_versions)
+        return VersionResponse(
+            valid_versions=valid_versions, newer_versions=new_versions
+        )
 
     async def load_api(self, pkg_set):
         async with client_session() as session:
@@ -104,7 +104,7 @@ class LaunchpadVersionAPI(VersionAPI):
                 response = await session.request(method="GET", url=url)
                 resp_json = await response.json()
                 if resp_json["entries"] == []:
-                    self.cache[pkg] = {}
+                    self.cache[pkg] = set()
                     break
                 for release in resp_json["entries"]:
                     all_versions.add(
@@ -118,8 +118,12 @@ class LaunchpadVersionAPI(VersionAPI):
                 else:
                     break
             self.cache[pkg] = all_versions
-        except (ClientResponseError, asyncio.exceptions.TimeoutError, ServerDisconnectedError):
-            self.cache[pkg] = {}
+        except (
+            ClientResponseError,
+            asyncio.exceptions.TimeoutError,
+            ServerDisconnectedError,
+        ):
+            self.cache[pkg] = set()
 
 
 class PypiVersionAPI(VersionAPI):
@@ -242,7 +246,7 @@ class DebianVersionAPI(VersionAPI):
             resp_json = await response.json()
 
             if resp_json.get("error") or not resp_json.get("versions"):
-                self.cache[pkg] = {}
+                self.cache[pkg] = set()
                 return
             for release in resp_json["versions"]:
                 all_versions.add(Version(value=release["version"].replace("0:", "")))
@@ -250,8 +254,12 @@ class DebianVersionAPI(VersionAPI):
             self.cache[pkg] = all_versions
         # TODO : Handle ServerDisconnectedError by using some sort of
         # retry mechanism
-        except (ClientResponseError, asyncio.exceptions.TimeoutError, ServerDisconnectedError):
-            self.cache[pkg] = {}
+        except (
+            ClientResponseError,
+            asyncio.exceptions.TimeoutError,
+            ServerDisconnectedError,
+        ):
+            self.cache[pkg] = set()
 
 
 class MavenVersionAPI(VersionAPI):
@@ -295,10 +303,10 @@ class MavenVersionAPI(VersionAPI):
         return endpoint
 
     @staticmethod
-    def extract_versions(xml_response: ET.ElementTree) -> Set[str]:
+    def extract_versions(xml_response: ET.ElementTree) -> Set[Version]:
         all_versions = set()
         for child in xml_response.getroot().iter():
-            if child.tag == "version":
+            if child.tag == "version" and child.text:
                 all_versions.add(Version(child.text))
 
         return all_versions
@@ -321,7 +329,7 @@ class NugetVersionAPI(VersionAPI):
         return base_url.format(pkg_name)
 
     @staticmethod
-    def extract_versions(resp: dict) -> Set[str]:
+    def extract_versions(resp: dict) -> Set[Version]:
         all_versions = set()
         try:
             for entry_group in resp["items"]:
@@ -329,7 +337,9 @@ class NugetVersionAPI(VersionAPI):
                     all_versions.add(
                         Version(
                             value=entry["catalogEntry"]["version"],
-                            release_date=dateparser.parse(entry["catalogEntry"]["published"]),
+                            release_date=dateparser.parse(
+                                entry["catalogEntry"]["published"]
+                            ),
                         )
                     )
         # FIXME: json response for YamlDotNet.Signed triggers this exception.
@@ -353,7 +363,7 @@ class ComposerVersionAPI(VersionAPI):
             self.cache[pkg] = self.extract_versions(resp, pkg)
 
     @staticmethod
-    def composer_url(pkg_name: str) -> str:
+    def composer_url(pkg_name: str) -> Optional[str]:
         try:
             vendor, name = pkg_name.split("/")
         except ValueError:
@@ -362,7 +372,7 @@ class ComposerVersionAPI(VersionAPI):
         return f"https://repo.packagist.org/p/{vendor}/{name}.json"
 
     @staticmethod
-    def extract_versions(resp: dict, pkg_name: str) -> Set[str]:
+    def extract_versions(resp: dict, pkg_name: str) -> Set[Version]:
         all_versions = set()
         for version in resp["packages"][pkg_name]:
             if "dev" in version:
@@ -374,7 +384,9 @@ class ComposerVersionAPI(VersionAPI):
             all_versions.add(
                 Version(
                     value=version.lstrip("v"),
-                    release_date=dateparser.parse(resp["packages"][pkg_name][version]["time"]),
+                    release_date=dateparser.parse(
+                        resp["packages"][pkg_name][version]["time"]
+                    ),
                 )
             )
         return all_versions
@@ -412,7 +424,7 @@ class GitHubTagsAPI(VersionAPI):
         }
     }"""
 
-    def __init__(self, cache: Mapping[str, Set[Version]] = None):
+    def __init__(self, cache: MutableMapping[str, Set[Version]] = None):
         self.gh_token = os.getenv("GH_TOKEN")
         super().__init__(cache=cache)
 
@@ -427,7 +439,10 @@ class GitHubTagsAPI(VersionAPI):
             session.headers["Authorization"] = "token " + self.gh_token
             endpoint = f"https://api.github.com/graphql"
             owner, name = owner_repo.split("/")
-            query = {"query": self.GQL_QUERY, "variables": {"name": name, "owner": owner}}
+            query = {
+                "query": self.GQL_QUERY,
+                "variables": {"name": name, "owner": owner},
+            }
 
             while True:
                 response = await session.post(endpoint, json=query)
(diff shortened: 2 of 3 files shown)


sify21:
fetching release_date from each version is very slow, commenting out those line will make it a lot faster.

pombredanne:
@sify21 Thank you ++ for this! Let me review this in details.

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+        pkg = pkg.removeprefix("https://pkg.go.dev/")
+        parts = pkg.split("/")
+        if len(parts) >= 2:
+            return "/".join(parts[:-1])
+        else:
+            return None
+
+    @staticmethod
+    def escape_path(path: str) -> str:
+        """escpe uppercase in module/version name"""
```
sbs2001, line 534: nit: typo  "escpe" -> "escape"

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+        escaped_path = ""
+        for c in path:
+            if c >= "A" and c <= "Z":
+                escaped_path += "!" + chr(ord(c) + ord("a") - ord("A"))
+            else:
+                escaped_path += c
+        return escaped_path
+
+    async def fetch(self, pkg: str, session: ClientSession):
+        # escpe uppercase in module path
```
sbs2001, line 544: nit: typo  "escpe" -> "escape"

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+        escaped_pkg = GoproxyVersionAPI.escape_path(pkg)
+        trimmed_pkg = pkg
+        resp_text = None
+        err_pkgs = []
+        while escaped_pkg is not None:
+            url = f"https://proxy.golang.org/{escaped_pkg}/@v/list"
+            try:
+                response = await session.request(method="GET", url=url)
+                resp_text = await response.text()
+            except:
```
sbs2001, line 554: Please be specific in the except statement and specify the type of exception

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+                err_pkgs.append(trimmed_pkg)
```
sbs2001, line 555: Why collect packages which got errored ? What's being done with this list ?

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+                escaped_path += c
+        return escaped_path
+
+    async def fetch(self, pkg: str, session: ClientSession):
+        # escpe uppercase in module path
+        escaped_pkg = GoproxyVersionAPI.escape_path(pkg)
+        trimmed_pkg = pkg
+        resp_text = None
+        err_pkgs = []
+        while escaped_pkg is not None:
```
sbs2001, line 549: I don't understand this, why is there a loop ? Maybe simplify or add a comment  on this ?

sbs2001: My guess is that this code tries to find some package on the proxy.golang thingy if it doesn't find it, some part of package is chopped. This is done till a match is found or everything is chopped.

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+                    escaped_ver = GoproxyVersionAPI.escape_path(value)
+                    try:
+                        response = await session.request(
+                            method="GET",
+                            url=f"https://proxy.golang.org/{escaped_pkg}/@v/{escaped_ver}.info",
+                        )
+                        resp_json = await response.json()
+                        release_date = parse_datetime(resp_json.get("Time", ""))
+                    except:
+                        release_date = None
```
sbs2001, line 581: Maybe refactor this into a new function. Which takes in version_info and spits out `Version` object ?

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+
+    @staticmethod
+    def escape_path(path: str) -> str:
+        """escpe uppercase in module/version name"""
+        escaped_path = ""
+        for c in path:
+            if c >= "A" and c <= "Z":
+                escaped_path += "!" + chr(ord(c) + ord("a") - ord("A"))
+            else:
+                escaped_path += c
```
sbs2001, line 540: This is too crytpic, maybe add test cases + comments.

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
         self.cache[pkg] = versions
+
+
+class GoproxyVersionAPI(VersionAPI):
+
+    package_type = "golang"
+    pkg_mappings = {}
+
+    @staticmethod
+    def trim_package_path(pkg: str) -> Optional[str]:
```
sbs2001, line 520: Maybe rename the method and parameters that the input `pkg` is actually a http url

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
...
+        """github advisories for golang is using package names(e.g. https://github.com/advisories/GHSA-jp4j-47f9-2vc3), yet goproxy works with module names(see https://golang.org/ref/mod#goproxy-protocol).
+        this method removes the last part of a package path, and returns the remaining as the module name.
+        """
+        # some advisories contains this prefix in package name, e.g. https://github.com/advisories/GHSA-7h6j-2268-fhcm
+        pkg = pkg.removeprefix("https://pkg.go.dev/")
```
sbs2001, line 525: I think the code should be explicit that this prefix is "maybe present":

Wrt readablity: 

```python
if pkg.startswith("https://pkg.go.dev/"):
 pkg = pkg.removeprefix("https://pkg.go.dev/")
```
is much better and makes code easy to follow.

vulnerabilities/package_managers.py
```diff
@@ -489,3 +509,75 @@ async def fetch(self, pkg, session):
             pass
 
         self.cache[pkg] = versions
+
+
+class GoproxyVersionAPI(VersionAPI):
+
+    package_type = "golang"
+    pkg_mappings = {}
```
sbs2001, line 517: Why use `pkg_mappings` ? Also `pkg_mappings` as a name is rather unhelpful. It doesn't convey what kind of mapping this is.

 Since this class inherits `VersionAPI` it has the `cache` attribute for this purpose. Maybe use that ? The `get` method also uses the `cache` for this.

sbs2001 (changes requested):
Thanks for the PR @sify21 . 

Most of the stuff looks very good. Thanks for updating the type hints btw. 

As an improvement in clarity and readability, maybe restructure the `GoproxyVersionAPI` such that:

 The `fetch` method only does one thing, it should only fetch and update the cache. It shouldn't 
 - Do low level stuff on modifying the obtained `version_info`.  
 - Construct the api url. 
 
Maybe create separte functions/methods for above two.

Also some unittests would be helpful.

vulnerabilities/package_managers.py
sify21: Oh it's just for error output, it can be removed

vulnerabilities/package_managers.py
sify21: will change to HTTPGone

vulnerabilities/package_managers.py
sify21: yes you're right. github advisory contains package path,  proxy.golang only works with module path and will report error when missing with http status 410.

vulnerabilities/package_managers.py
sify21: This variable maps package name(from github advisory) to module name(variable `trimmed_pkg` in `fetch` method).
I added this mapping because it can't be easilly done in `process_name` method of `github.py`, it  involves trimming url path and making request to proxy.golang to validate the trimmed path.

vulnerabilities/package_managers.py
sify21: Added some comments. It is correct as long as ascii doesn't change

vulnerabilities/package_managers.py
sify21: refactored

vulnerabilities/package_managers.py
sbs2001: Oh I see. Then maybe rename this to something like `module_name_by_package_name` ?

vulnerabilities/package_managers.py
sbs2001: Alright. Please add a comment for this. Otherwise we'll forget what's this doing ;)

vulnerabilities/package_managers.py
```diff
@@ -489,3 +510,86 @@ async def fetch(self, pkg, session):
...
+            try:
+                response = await session.request(method="GET", url=url)
+                resp_text = await response.text()
+            except HTTPGone:
+                escaped_pkg = GoproxyVersionAPI.trim_url_path(escaped_pkg)
+                trimmed_pkg = GoproxyVersionAPI.trim_url_path(trimmed_pkg) or ""
+                continue
+            break
+        if resp_text is None or escaped_pkg is None or trimmed_pkg is None:
+            print("error fetch versions from goproxy: ", pkg)
```
sbs2001, line 585: nit: "error while fetching versions for {pkg} from goproxy" would be better err message ?

vulnerabilities/package_managers.py
sbs2001: Maybe just provide a link to this https://go.dev/ref/mod#resolve-pkg-mod

sbs2001 (changes requested):
Thanks. This would be good to merge after
1. Adding some comments on as  suggested inline. Without prior knowledge it's hard to understand the intent of the code. 
2. Add some test cases or just examples in a doc string `trim_url_path` and  `escape_path` methods.

vulnerabilities/importers/github.py
```diff
@@ -229,7 +231,7 @@ def process_name(ecosystem: str, pkg_name: str) -> Optional[Tuple[Optional[str],
                 return None
             return vendor, name
 
-        if ecosystem == "NUGET" or ecosystem == "PIP" or ecosystem == "RUBYGEMS":
+        if ecosystem == "NUGET" or ecosystem == "PIP" or ecosystem == "RUBYGEMS" or ecosystem == "GO":
```
pombredanne, line 234:
```suggestion
        if ecosystem in ("NUGET", "PIP", "RUBYGEMS", "GO"):
```

vulnerabilities/package_managers.py
```diff
@@ -515,12 +515,12 @@ async def fetch(self, pkg, session):
...
 
     package_type = "golang"
-    pkg_mappings = {}
+    module_name_by_package_name = {}
 
     @staticmethod
     def trim_url_path(url_path: str) -> Optional[str]:
         """github advisories for golang is using package names(e.g. https://github.com/advisories/GHSA-jp4j-47f9-2vc3), yet goproxy works with module names(see https://golang.org/ref/mod#goproxy-protocol).
-        this method removes the last part of a package path, and returns the remaining as the module name.
+        this method removes the last part of a package path, and returns the remaining as the module name. For example: trim_url_path("https://github.com/xx/a/b") returns "https://github.com/xx/a"
```
pombredanne, line 523: Do you mind to wrap this whole docstring to a shorter length, like 80 characters?

vulnerabilities/package_managers.py
```diff
@@ -533,7 +533,7 @@ def trim_url_path(url_path: str) -> Optional[str]:
 
     @staticmethod
     def escape_path(path: str) -> str:
-        """escape uppercase in module/version name"""
+        """escape uppercase in module/version name. For example: escape_path("github.com/FerretDB/FerretDB") returns "github.com/!ferret!d!b/!ferret!d!b" """
```
pombredanne, line 536: Do you mind to wrap this whole docstring to a shorter length, like 80 characters?

pombredanne (changes requested):
Thanks!
Do you think you could add a few unit tests before we merge?

sify21:
Hi @pombredanne, I added some tests. I don't know how to resolve conflicts on github, should I pull the latest changes and resolve the conflicts on my fork?

pombredanne:
> I added some tests. I don't know how to resolve conflicts on github, should I pull the latest changes and resolve the conflicts on my fork?

yes, you need to resolve these git conflicts locally, eventually rebasing and amending as needed and then force push to your branch once this is done

sify21:
@pombredanne merged

pombredanne:
Thanks! There is a merge commit that's making the DCO bot unhappy... I will ignore this

pombredanne (changes requested):
Thank you for the update. Do you mind to run black to reformat your code?
`make black` should be enough

sify21:
@pombredanne reformatted

vulnerabilities/tests/test_package_managers.py
```diff
@@ -448,6 +453,52 @@ def test_fetch(self):
         assert self.version_api.get("org.apache:kafka") == VersionResponse(valid_versions=expected)
 
 
+class TestGoproxyVersionAPI(TestCase):
+    @classmethod
```
pombredanne, line 457: I am not a big fan of using setups as they make much harder to read and understand each of the tests .... What do you gain from this when compared to in-lining the setup as needed in each test method?
IMHO self-contained tests are easier to deal with in almost all cases.

vulnerabilities/package_managers.py
```diff
@@ -489,3 +500,93 @@ async def fetch(self, pkg, session):
...
+    def trim_url_path(url_path: str) -> Optional[str]:
+        """
+        github advisories for golang is using package names(e.g.
+        https://github.com/advisories/GHSA-jp4j-47f9-2vc3),
+        yet goproxy works with module names(see
+        https://golang.org/ref/mod#goproxy-protocol).
+        this method removes the last part of a package path, and returns the
+        remaining as the module name. For example: trim_url_path(
+        "https://github.com/xx/a/b") returns "https://github.com/xx/a"
+        """
```
pombredanne, line 520: What about turning this example in a doctest?

pombredanne: Also what about this style of docstring instead?

```
        Return a trimmed Go  ``url_path` removing trailing
        package references and keeping only the module
        references.

        Github advisories for Go are using package names 
        such as "https://github.com/nats-io/nats-server/v2/server" 
        (e.g., https://github.com/advisories/GHSA-jp4j-47f9-2vc3 ),
        yet goproxy works with module names instead such as
        "https://github.com/nats-io/nats-server" (see for details 
        https://golang.org/ref/mod#goproxy-protocol ).
        This functions trims the trailing part(s) of a package URL
        and returns the remaining the module name.
        For example: 
        >>> module = "https://github.com/xx/a"
        >>> assert trim_url_path("https://github.com/xx/a/b") == module
```

vulnerabilities/package_managers.py
```diff
@@ -489,3 +500,93 @@ async def fetch(self, pkg, session):
...
+        parts = url_path.split("/")
+        if len(parts) >= 2:
+            return "/".join(parts[:-1])
+        else:
+            return None
+
+    @staticmethod
+    def escape_path(path: str) -> str:
+        """
+        escape uppercase in module/version name.
```
pombredanne, line 533:
```suggestion
        Return an escaped uppercase module path.
```

pombredanne: Can you also explain the rationale for this in the docstring or a comment?

vulnerabilities/package_managers.py
```diff
@@ -489,3 +500,93 @@ async def fetch(self, pkg, session):
...
+            else:
+                escaped_path += c
+        return escaped_path
+
+    @staticmethod
+    async def parse_version_info(
+        version_info: str, escaped_pkg: str, session: ClientSession
+    ) -> Optional[Version]:
+        v = version_info.split()
+        if len(v) > 0:
```
pombredanne, line 549: Typically you want to bail out early on condition that are not met. What about using this instead:
```suggestion
        if not v:
            return
```
then you can dedent the block below. This becomes IMHO easier to read

vulnerabilities/package_managers.py
```diff
@@ -489,3 +500,93 @@ async def fetch(self, pkg, session):
...
+            value = v[0]
+            if len(v) > 1:
```
pombredanne, line 551: Are you sure you did not mean `if len(v) == 1` ?

vulnerabilities/package_managers.py
```diff
@@ -489,3 +500,93 @@ async def fetch(self, pkg, session):
...
+                escaped_ver = GoproxyVersionAPI.escape_path(value)
+                try:
+                    response = await session.request(
+                        method="GET",
+                        url=f"https://proxy.golang.org/{escaped_pkg}/@v/{escaped_ver}.info",
+                    )
+                    resp_json = await response.json()
+                    release_date = parse_datetime(resp_json.get("Time", ""))
+                except:
+                    release_date = None
```
pombredanne, line 563: what if there was a 500 or a 404? ... catching all exceptions feels a bit brutal to me.

pombredanne (changes requested):
@sify21 I have added a few extra comments for your consideration.

vulnerabilities/package_managers.py
pombredanne: You did add tests so I think we are good... but You should paste these URLS as comments so we can understand why you do this this way.

vulnerabilities/tests/test_package_managers.py
sify21: Well, I just imitated other tests, I thought it was a tradition. I agree with you, setups are only useful when they are shared by multiple tests.

pombredanne: Good catch! IMHO they are rarely useful... so let's enter an issue to fix this and avoid using them everywhere unless absolutely necessary.
I entered https://github.com/nexB/vulnerablecode/issues/619

vulnerabilities/package_managers.py
sify21: This logic is from `golang/go` source: https://github.com/golang/go/blob/be0d049a42ee4b07bfb71acb5e8f7c3d2735049a/src/cmd/go/internal/modfetch/proxy.go#L308, I couldn't find references in official docs. It seems that sometimes the second part contains the version's release time, and it can avoids sending another query to `$module/@v/$version.info`. Though I didn't handle the case where it infers the timestamp from pseudo-version.

sify21: I added comments under this line.

vulnerabilities/package_managers.py
sify21: I thought release date was optional, so I just set it to None when having problems querying it.

pombredanne: @sify21 wrt. to errors, we need to have a centralized way to make network requests and deal with errors and retries so what you can do for now is create an issue as a reminder to revisit this later to properly handle errors.
But in anycase, do not silently pass all exceptions there. Instead do some logging at the minimum for now.

sify21: I added some code to log error messages, and created the issue here https://github.com/nexB/vulnerablecode/issues/619

Hritik14:
@sify21 As the development in this branch is going on and we wanted to move ahead with #476, rebasing/merging this with/from main will likely cause merge conflicts. Please accept the incoming changes for the import statements.
Alternatively, give [kdiff3](http://kdiff3.sourceforge.net/) a try for solving merge conflicts, it's really smart.

sify21:
@Hritik14 merged

pombredanne (approved):
isort fails... I am running isort and merging now!
Thank you ++

PR #592, opened 2022-01-24
Fix deprecated Django API

### LINKED ISSUE #587: Upgrade to Django 4.x

yilmi:
Issue:
The application can be started in it's current state because ugettext_lazy was deprecated. See https://forum.djangoproject.com/t/importerror-cannot-import-name-ugettext-lazy-from-django-utils-translation/10943

Fix:
Replaced deprecated API with supported one

This Fixes #587 (actually, only the use of an API deprecated in Django 4.x)

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -25,7 +25,7 @@ from datetime import datetime
 
 from django.db import models
 from django.core.exceptions import ValidationError
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 from packageurl.contrib.django.models import PackageURLMixin
 from packageurl import PackageURL
 


pombredanne (approved):
Thank you ++
LGTM!

pombredanne:
@yilmi we are just missing a DCO signoff

yilmi:
> @yilmi we are just missing a DCO signoff

Thanks, I missed it, should be fixed now :+1:

yilmi:
Just in case, it seems that you need to approve the workflow again :sweat_smile:

pombredanne:
@yilmi I am not sure  that `<9211413+yilmi@users.noreply.github.com>` is a proper email for signoff... is it? :) the idea is to use a real email

pombredanne:
Now this is pretty minor change, so I will not bug you more with this!

pombredanne (approved):

Issue #501, opened 2021-07-15
refactor importer_yielder.py


Hritik14:
Do we really need this file ? Could the importer classes help enclosing these information ?
@pombredanne can you shed a little more light here ?

via https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#meeting-on-thursday-2021-07-15-at-1200-utc

Hritik14:
Dumping importer_yielder
- Simply store data source as concrete lists
- Don't store configuration or last run, maybe we can have better Log Importer Runs later. Ticket: https://github.com/nexB/vulnerablecode/issues/526

Consider: https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/__init__.py#L41

via https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#meeting-on-tuesday-2021-08-10-at-1400-utc

Issue #597, opened 2022-01-26
Migrate all existing importers to importer improver structure.


Hritik14:
Old structure:
- Overview: https://github.com/nexB/vulnerablecode/blob/main/docs/source/how-to-guides/add_new_importer.rst
- Codebase: https://github.com/nexB/vulnerablecode/tree/ed2131656b1b3030c2f9eb28e25c10dbbedc8e1d

New structure:
- Overview: https://github.com/nexB/vulnerablecode/pull/525
- Codebase: https://github.com/nexB/vulnerablecode

After migrating each importer make sure to
- Modify tests under [vulnerabilities/tests](https://github.com/nexB/vulnerablecode/tree/main/vulnerabilities/tests)
- Enable the modified test by removing it from [collect_ignore](https://github.com/nexB/vulnerablecode/blob/main/vulnerabilities/tests/conftest.py#L43)
- Enable doctests for the importer by removing it from [ignore list](https://github.com/nexB/vulnerablecode/blob/main/pytest.ini#L7) 

Following need migration to the new framework:

- [x] nginx https://github.com/nexB/vulnerablecode/pull/525
- [x] alpine https://github.com/nexB/vulnerablecode/issues/620
- [x] nginx tests https://github.com/nexB/vulnerablecode/issues/643
- [x] github https://github.com/nexB/vulnerablecode/issues/604
- [x] openssl https://github.com/nexB/vulnerablecode/issues/615
- [x] #666
- [x] Pysec https://github.com/nexB/vulnerablecode/issues/607
- [x] gitlab https://github.com/nexB/vulnerablecode/issues/438
- [x] #718
- [x] #716
- [x] archlinux
- [x] #717
- [x] #741
- [x] #969
- [x] npm https://github.com/nexB/vulnerablecode/issues/605
- [x] #606
- [x] #971
- [x] https://github.com/nexB/vulnerablecode/issues/78
- [x] #1055
- [x] #1059
- [x] #972 ( PR under progress )
- [x] #970 ( PR under progress )
- [x] https://github.com/nexB/vulnerablecode/issues/103 ( PR under progress )
- [x] #1052 ( PR under progress )
- [x] #1051 ( PR under progress )
- [x] #1040 ( PR under progress )
- [x] #1060 ( PR under progress )
- [x] osv https://github.com/nexB/vulnerablecode/issues/608 ( PR under progress ) -  it's not an old importer that needs to be migrated we are writing a new importer.
[...]

Issue #590, opened 2022-01-24
Improve logging


pombredanne:
We need to have multiple levels of logging, preferably stored in the DB:

1. at the importer level, we want to know when we ran an import with a summary of what was imported, a date
2. this could also have a list of errors in a separate log (table?)
3. at the improver level, we should update a per-record log that briefly tells when which improver ran that modified data

sify21:
Is there a way to set global logging level? When I run ubuntu importer, I see lots of debug messages from univers.

pombredanne:
@sify21 re:
> Is there a way to set global logging level?

Not at the moment. The story on logging is in flux for now. Eventually we want to have logging going to the DB as described in this issue and I reckon that logging is way too verbose at this stage.
If you feel like lending a hand here that would be great!

PR #600, opened 2022-01-26
Dump importer_yielder in favor of IMPORTER_REGISTRY and drop Etags

Hritik14:
- Fixes: https://github.com/nexB/vulnerablecode/issues/501
- Fixes: https://github.com/nexB/vulnerablecode/issues/321
- Fixes: https://github.com/nexB/vulnerablecode/issues/601

The name `DataSource` and `Importer` are majorly used interchangeably. This needs to be fixed in a different PR.


vulnerabilities/data_source.py
```diff
@@ -214,104 +212,42 @@ def __post_init__(self):
...
-        self.validate_configuration()
-
-    def __enter__(self):
-        pass
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        pass
+    def __init__(self):
+        if not self.license:
+            warnings.warn(f"Running importer {self!r} without a license", NoLicenseWarning)
```
pombredanne, line 230: What about rather raising an exception and make this mandatory? I am not a big fan of warnings.
```suggestion
            raise Exception(f"Cannot run importer {self!r} without a license")
```

vulnerabilities/data_source.py
```diff
@@ -214,104 +212,42 @@ def __post_init__(self):
...
         raise NotImplementedError
 
-    def error(self, msg: str) -> None:
-        """
-        Helper method for raising InvalidConfigurationError with the class name in the message.
-        """
-        raise InvalidConfigurationError(f"{type(self).__name__}: {msg}")
+
+# TODO: Adopt the same design as that for DataSource
+class DataSourceConfiguration:
```
pombredanne, line 249: Can you explain what this is?
IMHO we do not want another class for configuration, but rather actual direct class-level attributes/constants on each of the Importers which would be simpler, direct and more explicit. You can have a base class/mixin to document these fields too.

vulnerabilities/import_runner.py
```diff
@@ -32,65 +32,34 @@
...
-
-    Efficiency:
-        - Bulk inserts should be used whenever possible.
-        - Checking whether a record already exists should be kept to a minimum
-        (the data source should know this instead).
-        - All update and select operations must use indexed columns.
     """
 
-    def __init__(self, importer: models.Importer):
+    def __init__(self, importer: DataSource):
```
pombredanne, line 51: This should be an Importer, not a data source IMHO

vulnerabilities/import_runner.py
```diff
@@ -32,65 +32,34 @@
...
-
-
-def get_vuln_pkg_refs(vulnerability, package):
-    return models.PackageRelatedVulnerability.objects.filter(
-        vulnerability=vulnerability,
-        package=package,
-    )
+        logger.info(f"Starting import for {self.importer.qualified_name}")
+        advisory_datas = self.importer().advisory_data()
+        importer_name = self.importer.qualified_name
```
pombredanne, line 60: Move this up and do not call it twice like above and below.

vulnerabilities/import_runner.py
```diff
@@ -32,65 +32,34 @@
...
+        process_advisories(advisory_datas=advisory_datas, importer_name=importer_name)
+        logger.info(f"Finished import for {self.importer.qualified_name}.")
```
pombredanne, line 62: Can you provide a better message? some number of things processed?

vulnerabilities/import_runner.py
```diff
@@ -32,65 +32,34 @@
...
-
-def get_vuln_pkg_refs(vulnerability, package):
-    return models.PackageRelatedVulnerability.objects.filter(
-        vulnerability=vulnerability,
-        package=package,
-    )
+        logger.info(f"Starting import for {self.importer.qualified_name}")
+        advisory_datas = self.importer().advisory_data()
+        importer_name = self.importer.qualified_name
+        process_advisories(advisory_datas=advisory_datas, importer_name=importer_name)
```
pombredanne, line 61: This would be best returning something IMHO, possibly yielding what the key for inserted records?

vulnerabilities/importers/__init__.py
```diff
@@ -19,34 +19,8 @@
...
-# from vulnerabilities.importers.ruby import RubyDataSource
-# from vulnerabilities.importers.rust import RustDataSource
-# from vulnerabilities.importers.safety_db import SafetyDbDataSource
-# from vulnerabilities.importers.suse_scores import SUSESeverityScoreDataSource
-# from vulnerabilities.importers.ubuntu import UbuntuDataSource
-# from vulnerabilities.importers.ubuntu_usn import UbuntuUSNDataSource
-# from vulnerabilities.importers.istio import IstioDataSource
-
-# from vulnerabilities.importers.suse_backports import SUSEBackportsDataSource
+importer_mapping = {x.qualified_name: x for x in IMPORTER_REGISTRY}
```
pombredanne, line 26: This is a constant, name it `IMPORTER_REGISTRY` and the ,keys() are the fqn and .values() gets you importers.

vulnerabilities/management/commands/import.py
```diff
@@ -21,16 +21,14 @@
...
-from datetime import datetime
 import traceback
 
 from django.core.management.base import BaseCommand
 from django.core.management.base import CommandError
 
-from vulnerabilities.models import Importer
-from vulnerabilities.importer_yielder import IMPORTER_REGISTRY
+from vulnerabilities.importers import IMPORTER_REGISTRY
+from vulnerabilities.importers import importer_mapping
```
pombredanne, line 30: There should be only one IMPORTER_REGISTRY

vulnerabilities/importers/__init__.py
pombredanne, line 26:
```suggestion
IMPORTER_REGISTRY = {x.qualified_name: x for x in IMPORTER_REGISTRY}
```

vulnerabilities/improvers/__init__.py
```diff
@@ -3,4 +3,4 @@
 
 IMPROVER_REGISTRY = [default.DefaultImprover, importers.nginx.NginxBasicImprover]
 
-improver_mapping = {x.qualified_name(): x for x in IMPROVER_REGISTRY}
+improver_mapping = {x.qualified_name: x for x in IMPROVER_REGISTRY}
```
pombredanne, line 6: Use the same design as for importers:
```suggestion
IMPROVER_REGISTRY = {x.qualified_name: x for x in IMPROVER_REGISTRY}
```

vulnerabilities/management/commands/import.py
pombredanne, line 30:
```suggestion
```

vulnerabilities/management/commands/import.py
```diff
@@ -47,22 +45,16 @@ def add_arguments(self, parser):
...
     def handle(self, *args, **options):
-        # load_importers() seeds the DB with Importers
-        load_importers()
         if options["list"]:
             self.list_sources()
             return
 
         if options["all"]:
-            self._import_data(Importer.objects.all(), options["cutoff_date"])
+            self.import_data(IMPORTER_REGISTRY)
```
pombredanne, line 57:
```suggestion
            self.import_data(IMPORTER_REGISTRY.values())
```

vulnerabilities/management/commands/import.py
```diff
@@ -71,44 +63,47 @@ def handle(self, *args, **options):
                 'Please provide at least one data source to import from or use "--all".'
             )
 
-        self.import_data(sources, options["cutoff_date"])
+        self.import_data(valid_sources(sources))
 
     def list_sources(self):
-        importers = IMPORTER_REGISTRY
+        importers = [importer.qualified_name for importer in IMPORTER_REGISTRY]
```
pombredanne, line 69: this should just be:
```suggestion
        importers = list(IMPORTER_REGISTRY)
```

vulnerabilities/management/commands/import.py
```diff
@@ -71,44 +63,47 @@ def handle(self, *args, **options):
...
-                unknown_importers.add(name)
-
-        if unknown_importers:
-            unknown_importers = ", ".join(unknown_importers)
-            raise CommandError(f"Unknown data sources: {unknown_importers}")
-
-        self._import_data(importers, cutoff_date)
-
-    def _import_data(self, importers, cutoff_date):
+    def import_data(self, importers):
```
pombredanne, line 73: What you call importers are importer classes?
Please add a docstring

vulnerabilities/management/commands/import.py
```diff
@@ -71,44 +63,47 @@ def handle(self, *args, **options):
...
+                    self.style.ERROR(
+                        f"Failed to run importer {importer.qualified_name}. Continuing..."
+                    )
                 )
+
         if failed_importers:
             raise CommandError(f"{len(failed_importers)} failed!: {','.join(failed_importers)}")
+
+
+def valid_sources(sources):
```
pombredanne, line 98: May be:

```suggestion
def validate_importers(sources):
```

Also please add a docstring

vulnerabilities/management/commands/improve.py
```diff
@@ -65,25 +62,29 @@ def handle(self, *args, **options):
         self.improve_data(valid_sources(sources))
 
     def list_sources(self):
-        improvers = [improver.qualified_name() for improver in IMPROVER_REGISTRY]
+        improvers = [improver.qualified_name for improver in IMPROVER_REGISTRY]
```
pombredanne, line 65:
```suggestion
        improvers = list(IMPROVER_REGISTRY)
```

vulnerabilities/models.py
```diff
@@ -261,53 +261,6 @@ def update_or_create(self):
             )
 
 
-class Importer(models.Model):
```
pombredanne, line 264: Why not adding an ``ImporterRun`` model right away to log each run there?

pombredanne (changes requested):
Thanks! I have a few suggestions and ... Could you add tests?

vulnerabilities/data_source.py
Hritik14: @pombredanne I initially made this an exception but couldn't find a license for nginx advisories, that way we'd have no working importers.

vulnerabilities/data_source.py
Hritik14: This class exists only to make sure all existing (but inactive) importers don't cause errors while django bootup. These will be removed

vulnerabilities/import_runner.py
Hritik14: Correct. I intend to rename all DataSource to Importer but that's for next PR, out of context for this PR.

vulnerabilities/import_runner.py
Hritik14: Can you explain a bit more ?

vulnerabilities/importers/__init__.py
Hritik14: We already have an `IMPORTER_REGISTRY`. 

 importer_mapping = {x.qualified_name: x for x in IMPORTER_REGISTRY} 
--------------------------------------------------------^^^^^^^^^^^

Hritik14: IMPORTER_REGISTRY cannot iterate over itself to generate itself ...

vulnerabilities/improvers/__init__.py
Hritik14: Same as above

vulnerabilities/management/commands/import.py
Hritik14: Do ignore all imports. We have #476 for those.

vulnerabilities/models.py
Hritik14: Logging will be a different module as we discussed earlier. Other than printing to stderr, there's no logging here (so that nothing interferes with our logging module)

vulnerabilities/import_runner.py
Hritik14: Done

vulnerabilities/import_runner.py
Hritik14: This sounds a little off to me as `process_advisories` is a function that is supposed to process all the advisories and be done with it. It is not required to run as you consume the output (in case of generators).

I've implemented it anyway.

vulnerabilities/importers/__init__.py
Hritik14: Using this 
```py
IMPORTERS = [nginx.NginxDataSource]

IMPORTER_REGISTRY = {x.qualified_name: x for x in IMPORTERS}
```

vulnerabilities/import_runner.py
Hritik14: https://github.com/nexB/vulnerablecode/issues/601

vulnerabilities/management/commands/import.py
Hritik14: Fixed

vulnerabilities/management/commands/import.py
Hritik14: Done

vulnerabilities/management/commands/import.py
Hritik14: Done

vulnerabilities/management/commands/import.py
Hritik14: Changed

vulnerabilities/management/commands/improve.py
Hritik14: Done

vulnerabilities/import_runner.py
Hritik14: For now, I've made it to return the total count of imported advisories

vulnerabilities/data_source.py
Hritik14: we'll use `license_expression`

vulnerabilities/data_source.py
Hritik14: Remove it from all data sources

vulnerabilities/import_runner.py
Hritik14:
```py
importer_class = self.importer
importer_name = self.importer.qual...
```

vulnerabilities/importers/__init__.py
Hritik14: `IMPORTERS_REGISTRY` and one variable

vulnerabilities/models.py
Hritik14: https://github.com/nexB/vulnerablecode/issues/590

vulnerabilities/import_runner.py
```diff
@@ -25,79 +25,50 @@
...
-        vulnerability=vulnerability,
-        package=package,
-    )
-
-
-def process_advisories(advisory_datas: Iterable[AdvisoryData], importer_name: str) -> None:
+def process_advisories(advisory_datas: Iterable[AdvisoryData], importer_name: str) -> List:
     """
     Insert advisories into the database
+    Return list of ids of inserted advisories
```
pombredanne, line 69:
```suggestion
    Return the number of inserted advisories.
```

vulnerabilities/importer.py
```diff
@@ -28,15 +28,13 @@
...
-from typing import Any
-from typing import ContextManager
-from typing import Iterable
 from typing import List
 from typing import Mapping
 from typing import Optional
 from typing import Set
 from typing import Iterable
 from typing import Tuple
+import warnings
```
pombredanne, line 37:
```suggestion
```

vulnerabilities/importer.py
```diff
@@ -214,118 +213,38 @@ def __post_init__(self):
...
+        if not self.spdx_license_expression:
+            raise Exception(f"Cannot run importer {self!r} without a license")
 
-            setattr(self, "_cutoff_timestamp", max(last_run, cutoff))
-
-        return self._cutoff_timestamp
-
-    def validate_configuration(self) -> None:
+    @classproperty
+    def qualified_name(self):
```
pombredanne, line 233:
```suggestion
    def qualified_name(cls):
```

vulnerabilities/importer.py
```diff
@@ -214,118 +213,38 @@ def __post_init__(self):
...
         """
-        Subclasses can perform more complex validation than what is handled by data classes and
-        their type annotations.
-
-        This method is called in the constructor. It should raise InvalidConfigurationError with a
-        human-readable message.
+        Fully qualified name prefixed with the module name of the improver used in logging.
         """
+        return f"{self.__module__}.{self.__qualname__}"
```
pombredanne, line 237:
```suggestion
        return f"{cls.__module__}.{cls.__qualname__}"
```

vulnerabilities/importer.py
```diff
@@ -214,118 +213,38 @@ def __post_init__(self):
...
-    branch: Optional[str] = None
-    create_working_directory: bool = True
-    remove_working_directory: bool = True
-    working_directory: Optional[str] = None
-
-
-class GitDataSource(DataSource):
-    CONFIG_CLASS = GitDataSourceConfiguration
 
+# TODO: Needs rewrite
```
pombredanne, line 246: May be an issue to track may work better?

vulnerabilities/improver.py
```diff
@@ -72,27 +74,28 @@ def from_advisory_data(cls, advisory_data, confidence, affected_purls, fixed_pur
...
-    Improvers are responsible to improve the already imported data by a datasource.
-    Inferences regarding the data could be generated based on multiple factors.
+    Improvers are responsible to improve already imported data by an importer.  An improver is
+    required to override the ``interesting_advisories`` property method to return a QuerySet of
+    ``Advisory`` objects. These advisories are then passed to ``get_inferences`` method which is
+    responsible for returning an iterable of ``Inferences`` for that particular ``Advisory``
     """
 
+    @classproperty
+    def qualified_name(self):
```
pombredanne, line 84:
```suggestion
    def qualified_name(cls):
```

vulnerabilities/improver.py
```diff
@@ -72,27 +74,28 @@ def from_advisory_data(cls, advisory_data, confidence, affected_purls, fixed_pur
...
+        """
+        Fully qualified name prefixed with the module name of the improver used in logging.
+        """
+        return f"{self.__module__}.{self.__qualname__}"
```
pombredanne, line 88:
```suggestion
        return f"{cls.__module__}.{cls.__qualname__}"
```

vulnerabilities/tests/test_import_runner.py
```diff
@@ -20,260 +20,89 @@
...
-    assert vuln_refs.count() == 1
-    assert vuln_refs[0].url == "https://example.com/with/more/info/CVE-2020-13371337"
+def test_process_advisories_with_advisories(db):
+    process_advisories(ADVISORY_DATAS, "test_importer")
+    advisories = models.Advisory.objects.all()
+    advisory_datas = [x.to_advisory_data() for x in advisories]
+    assert advisory_datas == ADVISORY_DATAS
 
 
-def test_ImportRunner_existing_package_and_new_vulnerability(db):
```
pombredanne, line 128: Could this test be kept somehow? to valid that we never will allow twice the same set of imported data?

vulnerabilities/tests/test_import_runner.py
```diff
@@ -20,260 +20,89 @@
...
-
-    vuln = prv.vulnerability
-    assert vuln.vulnerability_id == "CVE-2020-13371337"
-
-    vuln_refs = models.VulnerabilityReference.objects.filter(vulnerability=vuln)
-    assert vuln_refs.count() == 1
-    assert vuln_refs[0].url == "https://example.com/with/more/info/CVE-2020-13371337"
-
-
-def test_ImportRunner_new_package_version_affected_by_existing_vulnerability(db):
```
pombredanne, line 159: IMHO this may be need to be kept

vulnerabilities/tests/test_import_runner.py
```diff
@@ -20,260 +20,89 @@
...
-    assert len(qs) == 1
-    added_package = qs[0]
-
-    qs = models.PackageRelatedVulnerability.objects.filter(package=added_package)
-    assert len(qs) == 1
-    impacted_package = qs[0]
-    assert impacted_package.vulnerability.vulnerability_id == "CVE-2020-13371337"
-
-
-# def test_ImportRunner_fixed_package_version_is_added(db):
```
pombredanne, line 207: Could this test have been of value? and need to be kept?

vulnerabilities/tests/test_import_runner.py
```diff
@@ -20,260 +20,89 @@
...
-
-#     qs = models.PackageRelatedVulnerability.objects.filter(
-#         package=added_package, is_vulnerable=False
-#     )
-#     assert len(qs) == 1
-#     resolved_package = qs[0]
-#     assert resolved_package.vulnerability.vulnerability_id == "CVE-2020-13371337"
-
-
-def test_ImportRunner_updated_vulnerability(db):
```
pombredanne, line 247: IMHO this may be need to be kept, moved to improvers?

pombredanne (changes requested):
Thanks... see a few final nits for your consideration.

vulnerabilities/importer.py
Hritik14: This will be a part of https://github.com/nexB/vulnerablecode/issues/597

vulnerabilities/tests/test_import_runner.py
Hritik14: This is done here
https://github.com/nexB/vulnerablecode/blob/c62f81f84c272686be972acebe15e2d2d6246e85/vulnerabilities/tests/test_import_runner.py#L80-L86

pombredanne (approved):
LGTM! :+1:

Issue #607, opened 2022-02-02
Add PYSEC importer


pombredanne:
This would be a first OSV-based importer

pombredanne:
As a follow up, we need to check the schema version (and eventually fail if this is not what is expected)
Also we should contemplate using the https://github.com/pypa/advisory-database/ separately and directly as this and the OSV two may not be exactly in sync

pombredanne:
See https://github.com/nexB/vulnerablecode/issues/774 for followup. This is all imported and merged otehrwise.
@ziadhany Thank you ++ :bow:

PR #632, opened 2022-02-20
Add PyPI OSV

ziadhany:
#607 

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -113,3 +113,4 @@ wcwidth==0.2.5
 websocket-client==0.59.0
 yarl==1.7.2
 zipp==3.8.0
+dateparser==1.1.1
--- a/vulnerabilities/helpers.py
+++ b/vulnerabilities/helpers.py
@@ -267,3 +267,16 @@ def _get_gh_response(gh_token, graphql_query):
     endpoint = "https://api.github.com/graphql"
     headers = {"Authorization": f"bearer {gh_token}"}
     return requests.post(endpoint, headers=headers, json=graphql_query).json()
+
+
+def dedupe(original: List) -> List:
+    """
+    Remove all duplicate items and return a new list preserving ordering
+    >>> dedupe(["z","i","a","a","d","d"])
+    ['z', 'i', 'a', 'd']
+    """
+    new_list = []
+    for i in original:
+        if i not in new_list:
+            new_list.append(i)
+    return new_list
--- a/vulnerabilities/importers/pysec.py
+++ b/vulnerabilities/importers/pysec.py
@@ -65,7 +65,7 @@ class PyPIImporter(Importer):
                     logger.error(f"NotImplementedError PyPI package file_name: {file_name}")
                 else:
                     with zip_file.open(file_name) as f:
-                        vul_info = json.loads(f.read())
+                        vul_info = json.load(f)
                         yield parse_advisory_data(vul_info)
 
 
--- a/vulnerabilities/importers/pysec.py
+++ b/vulnerabilities/importers/pysec.py
@@ -1,0 +1,239 @@
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+import json
+import logging
+from io import BytesIO
+from typing import Iterable
+from typing import Optional
+from zipfile import ZipFile
+
+import dateparser
+import requests
+from packageurl import PackageURL
+from univers.version_range import InvalidVersionRange
+from univers.version_range import PypiVersionRange
+from univers.versions import InvalidVersion
+from univers.versions import PypiVersion
+from univers.versions import SemverVersion
+
+from vulnerabilities.helpers import dedupe
+from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
+from vulnerabilities.importer import Importer
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+
+logger = logging.getLogger(__name__)
+
+
+class PyPIImporter(Importer):
+    license_url = "https://github.com/pypa/advisory-database/blob/main/LICENSE"
+    spdx_license_expression = "CC-BY-4.0"
+
+    def advisory_data(self) -> Iterable[AdvisoryData]:
+        """
+        1. Fetch the data from osv api
+        2. unzip the file
+        3. open the file one by one
+        4. yield the json file to parse_advisory_data
+        """
+        url = "https://osv-vulnerabilities.storage.googleapis.com/PyPI/all.zip"
+        response = requests.get(url).content
+        with ZipFile(BytesIO(response)) as zip_file:
+            for file_name in zip_file.namelist():
+                if not file_name.startswith("PYSEC-"):
+                    logger.error(f"NotImplementedError PyPI package file_name: {file_name}")
+                else:
+                    with zip_file.open(file_name) as f:
+                        vul_info = json.loads(f.read())
+                        yield parse_advisory_data(vul_info)
+
+
+def parse_advisory_data(raw_data: dict) -> Optional[AdvisoryData]:
+    raw_id = raw_data["id"]
+    summary = raw_data.get("summary") or ""
+    aliases = get_aliases(raw_data)
+    date_published = get_published_date(raw_data)
+    severity = list(get_severities(raw_data))
+    references = get_references(raw_data, severity)
+
+    affected_packages = []
+    if "affected" not in raw_data:
+        logger.error(f"affected_packages not found - {raw_id !r}")
+        return
+
+    for affected_pkg in raw_data.get("affected") or []:
+        purl = get_affected_purl(affected_pkg, raw_id)
+        if purl.type != "pypi":
+            logger.error(f"Non PyPI package found in PYSEC advisories: {purl} - from: {raw_id !r}")
+            continue
+
+        affected_version_range = get_affected_version_range(affected_pkg, raw_id)
+        for fixed_range in affected_pkg.get("ranges", []):
+            fixed_version = get_fixed_version(fixed_range, raw_id)
+
+            for version in fixed_version:
+                affected_packages.append(
+                    AffectedPackage(
+                        package=purl,
+                        affected_version_range=affected_version_range,
+                        fixed_version=version,
+                    )
+                )
+
+    return AdvisoryData(
+        aliases=aliases,
+        summary=summary,
+        affected_packages=affected_packages,
+        references=references,
+        date_published=date_published,
+    )
+
+
+def fixed_filter(fixed_range) -> []:
+    """
+    Return a list of fixed version strings given a ``fixed_range`` mapping of OSV data.
+    >>> list(fixed_filter({"type": "SEMVER", "events": [{"introduced": "0"}, {"fixed": "1.6.0"}]}))
+    ['1.6.0']
+    >>> list(fixed_filter({"type": "ECOSYSTEM","events":[{"introduced": "0"},{"fixed": "1.0.0"},{"fixed": "9.0.0"}]}))
+    ['1.0.0', '9.0.0']
+    """
+    for event in fixed_range.get("events") or []:
+        fixed = event.get("fixed")
+        if fixed:
+            yield fixed
+
+
+def get_aliases(raw_data) -> []:
+    """
+    aliases field is optional , id is required and these are all aliases from our perspective
+    converting list of two fields to a dict then , convert it to a list to make sure a list is unique
+    >>> get_aliases({"id": "GHSA-j3f7-7rmc-6wqj"})
+    ['GHSA-j3f7-7rmc-6wqj']
+    >>> get_aliases({"aliases": ["CVE-2021-40831"]})
+    ['CVE-2021-40831']
+    >>> get_aliases({"aliases": ["CVE-2022-22817", "GHSA-8vj2-vxx3-667w"], "id": "GHSA-j3f7-7rmc-6wqj"})
+    ['CVE-2022-22817', 'GHSA-8vj2-vxx3-667w', 'GHSA-j3f7-7rmc-6wqj']
+    """
+    vulnerability_id = raw_data.get("id")
+    vulnerability_aliases = raw_data.get("aliases") or []
+    if vulnerability_id:
+        vulnerability_aliases.append(vulnerability_id)
+    return vulnerability_aliases
+
+
+def get_published_date(raw_data):
+    published = raw_data.get("published")
+    return published and dateparser.parse(published)
+
+
+def get_severities(raw_data) -> []:
+    for sever_list in raw_data.get("severity") or []:
+        if sever_list.get("type") == "CVSS_V3":
+            yield VulnerabilitySeverity(
+                system=SCORING_SYSTEMS["cvssv3.1_vector"], value=sever_list["score"]
+            )
+        else:
+            logger.error(f"NotImplementedError severity type- {raw_data['id']!r}")
+
+    ecosys = raw_data.get("ecosystem_specific") or {}
+    sever = ecosys.get("severity")
+    if sever:
+        yield VulnerabilitySeverity(
+            system=SCORING_SYSTEMS["generic_textual"],
+            value=sever,
+        )
+
+    database_specific = raw_data.get("database_specific") or {}
+    sever = database_specific.get("severity")
+    if sever:
+        yield VulnerabilitySeverity(
+            system=SCORING_SYSTEMS["generic_textual"],
+            value=sever,
+        )
+
+
+def get_references(raw_data, severities) -> []:
+    references = raw_data.get("references") or []
+    return [Reference(url=ref["url"], severities=severities) for ref in references if ref]
+
+
+def get_affected_purl(affected_pkg, raw_id):
+    package = affected_pkg.get("package") or {}
+    purl = package.get("purl")
+    if purl:
+        try:
+            return PackageURL.from_string(purl)
+        except ValueError:
+            logger.error(f"PackageURL ValueError - {raw_id !r} - purl: {purl !r}")
+
+    ecosys = package.get("ecosystem")
+    name = package.get("name")
+    if ecosys and name:
+        return PackageURL(type=ecosys, name=name)
+    else:
+        logger.error(f"purl affected_pkg not found - {raw_id !r}")
+
+
+def get_affected_version_range(affected_pkg, raw_id):
+    affected_versions = affected_pkg.get("versions")
+    if affected_versions:
+        try:
+            return PypiVersionRange(affected_versions)
+        except InvalidVersionRange:
+            logger.error(f"InvalidVersionRange affected_pkg_version_range Error - {raw_id !r} ")
+    else:
+        logger.error(f"affected_pkg_version_range not found - {raw_id !r} ")
+
+
+def get_fixed_version(fixed_range, raw_id) -> []:
+    """
+    Return a list of fixed versions, using fixed_filter we get the list of fixed version strings,
+    then we pass every element to their univers.versions , then we dedupe the result
+    >>> get_fixed_version({}, "GHSA-j3f7-7rmc-6wqj")
+    []
+    >>> get_fixed_version({"type": "ECOSYSTEM", "events": [{"fixed": "1.7.0"}]}, "GHSA-j3f7-7rmc-6wqj")
+    [PypiVersion(string='1.7.0')]
+    """
+    fixed_version = []
+    if "type" not in fixed_range:
+        logger.error(f"Invalid type - {raw_id!r}")
+    else:
+        list_fixed = fixed_filter(fixed_range)
+        fixed_range_type = fixed_range["type"]
+        for i in list_fixed:
+            if fixed_range_type == "ECOSYSTEM":
+                try:
+                    fixed_version.append(PypiVersion(i))
+                except InvalidVersion:
+                    logger.error(f"Invalid Version - PypiVersion - {raw_id !r} - {i !r}")
+            if fixed_range_type == "SEMVER":
+                try:
+                    fixed_version.append(SemverVersion(i))
+                except InvalidVersion:
+                    logger.error(f"Invalid Version - SemverVersion - {raw_id !r} - {i !r}")
+            if fixed_range_type == "GIT":
+                # TODO add GitHubVersion univers fix_version
+                logger.error(f"NotImplementedError GIT Version - {raw_id !r} - {i !r}")
+
+    return dedupe(fixed_version)
--- a/vulnerabilities/tests/test_pysec.py
+++ b/vulnerabilities/tests/test_pysec.py
@@ -1,0 +1,513 @@
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/vulnerablecode/
+# The VulnerableCode software is licensed under the Apache License version 2.0.
+# Data generated with VulnerableCode require an acknowledgment.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with VulnerableCode or any VulnerableCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with VulnerableCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  VulnerableCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  VulnerableCode is a free software tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+import datetime
+import json
+import os
+from unittest import TestCase
+
+from packageurl import PackageURL
+from univers.version_range import PypiVersionRange
+from univers.versions import PypiVersion
+
+from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.importers.pysec import fixed_filter
+from vulnerabilities.importers.pysec import get_affected_purl
+from vulnerabilities.importers.pysec import get_affected_version_range
+from vulnerabilities.importers.pysec import get_aliases
+from vulnerabilities.importers.pysec import get_fixed_version
+from vulnerabilities.importers.pysec import get_published_date
+from vulnerabilities.importers.pysec import get_references
+from vulnerabilities.importers.pysec import get_severities
+from vulnerabilities.importers.pysec import parse_advisory_data
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+from vulnerabilities.severity_systems import ScoringSystem
+
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+TEST_DATA = os.path.join(BASE_DIR, "test_data/pysec", "pysec_test.json")
+
+
+class TestPyPIImporter(TestCase):
+    def test_to_advisories(self):
+        first_aff_range = [
+            "0.2.4",
+            "0.2.9",
+            "0.3.0",
+            "1.0.2",
+            "1.0.3",
+            "1.0.5",
+            "1.0.6",
+            "1.1.0",
+            "1.2.0",
+            "1.2.1",
+            "1.3.0",
+            "1.3.1",
+            "1.3.2",
+            "1.4.0",
+            "1.5.0",
+            "1.5.1",
+            "1.5.10",
+            "1.5.11",
+            "1.5.12",
+            "1.5.13",
+            "1.5.14",
+            "1.5.15",
+            "1.5.16",
+            "1.5.17",
+            "1.5.18",
+            "1.5.2",
+            "1.5.3",
+            "1.5.4",
+            "1.5.5",
+            "1.5.6",
+            "1.5.7",
+            "1.5.8",
+            "1.6.0",
+            "1.6.1",
+            "1.6.2",
+        ]
+        second_aff_range = [
+            "1.0",
+            "1.1",
+            "1.2",
+            "1.3",
+            "1.4",
+            "1.5",
+            "1.6",
+            "1.7.0",
+            "1.7.1",
+            "1.7.2",
+            "1.7.3",
+            "1.7.4",
+            "1.7.5",
+            "1.7.6",
+            "1.7.7",
+            "1.7.8",
+            "2.0.0",
+            "2.1.0",
+            "2.2.0",
+            "2.2.1",
+            "2.2.2",
+            "2.3.0",
+            "2.3.1",
+            "2.3.2",
+            "2.4.0",
+            "2.5.0",
+            "2.5.1",
+            "2.5.2",
+            "2.5.3",
+            "2.6.0",
+            "2.6.1",
+            "2.6.2",
+            "2.7.0",
+            "2.8.0",
+            "2.8.1",
+            "2.8.2",
+            "2.9.0",
+            "3.0.0",
+            "3.1.0",
+            "3.1.0.rc1",
+            "3.1.0rc1",
+            "3.1.1",
+            "3.1.2",
+            "3.2.0",
+            "3.3.0",
+            "3.3.1",
+            "3.3.2",
+            "3.3.3",
+            "3.4.0",
+            "3.4.1",
+            "3.4.2",
+            "4.0.0",
+            "4.1.0",
+            "4.1.1",
+            "4.2.0",
+            "4.2.1",
+            "4.3.0",
+            "5.0.0",
+            "5.1.0",
+            "5.2.0",
+            "5.3.0",
+            "5.4.0",
+            "5.4.0.dev0",
+            "5.4.1",
+            "6.0.0",
+            "6.1.0",
+            "6.2.0",
+            "6.2.1",
+            "6.2.2",
+            "7.0.0",
+            "7.1.0",
+            "7.1.1",
+            "7.1.2",
+            "7.2.0",
+            "8.0.0",
+            "8.0.1",
+            "8.1.0",
+            "8.1.1",
+            "8.1.2",
+            "8.2.0",
+            "8.3.0",
+            "8.3.1",
+            "8.3.2",
+            "8.4.0",
+        ]
+        with open(TEST_DATA) as f:
+            mock_response = json.load(f)
+
+        expected_advisories = [
+            AdvisoryData(
+                aliases=["CVE-2021-40831", "GHSA-j3f7-7rmc-6wqj"],
+                summary="Improper certificate management in AWS IoT Device SDK v2",
+                affected_packages=[
+                    AffectedPackage(
+                        package=PackageURL.from_string("pkg:pypi/awsiotsdk"),
+                        affected_version_range=PypiVersionRange(first_aff_range),
+                        fixed_version=PypiVersion("1.7.0"),
+                    ),
+                ],
+                references=[
+                    Reference(
+                        url="https://nvd.nist.gov/vuln/detail/CVE-2021-40831",
+                        severities=[
+                            VulnerabilitySeverity(
+                                system=SCORING_SYSTEMS["cvssv3.1_vector"],
+                                value="CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:N",
+                            )
+                        ],
+                    ),
+                    Reference(
+                        url="https://github.com/aws/aws-iot-device-sdk-cpp-v2",
+                        severities=[
+                            VulnerabilitySeverity(
+                                system=SCORING_SYSTEMS["cvssv3.1_vector"],
+                                value="CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:N",
+                            )
+                        ],
+                    ),
+                ],
+                date_published=datetime.datetime(2021, 11, 24, 20, 35, 3, 0).replace(
+                    tzinfo=datetime.timezone.utc
+                ),
+            ),
+            AdvisoryData(
+                aliases=["CVE-2022-22817", "GHSA-8vj2-vxx3-667w", "PYSEC-2022-10"],
+                summary="",
+                affected_packages=[
+                    AffectedPackage(
+                        package=PackageURL.from_string("pkg:pypi/pillow"),
+                        affected_version_range=PypiVersionRange(second_aff_range),
+                        fixed_version=PypiVersion("9.0.0"),
+                    )
+                ],
+                references=[
+                    Reference(
+                        url="https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#restrict-builtins"
+                        "-available-to-imagemath-eval",
+                        severities=[
+                            VulnerabilitySeverity(
+                                system=ScoringSystem(
+                                    identifier="generic_textual",
+                                    name="Generic textual severity rating",
+                                    url="",
+                                    notes="Severity for generic scoring systems. Contains generic textual values like High, Low etc",
+                                ),
+                                value="HIGH",
+                            )
+                        ],
+                    ),
+                    Reference(
+                        url="https://lists.debian.org/debian-lts-announce/2022/01/msg00018.html",
+                        severities=[
+                            VulnerabilitySeverity(
+                                system=ScoringSystem(
+                                    identifier="generic_textual",
+                                    name="Generic textual severity rating",
+                                    url="",
+                                    notes="Severity for generic scoring systems. Contains generic textual values like High, Low etc",
+                                ),
+                                value="HIGH",
+                            )
+                        ],
+                    ),
+                    Reference(
+                        url="https://github.com/advisories/GHSA-8vj2-vxx3-667w",
+                        severities=[
+                            VulnerabilitySeverity(
+                                system=ScoringSystem(
+                                    identifier="generic_textual",
+                                    name="Generic textual severity rating",
+                                    url="",
+                                    notes="Severity for generic scoring systems. Contains generic textual values like High, Low etc",
+                                ),
+                                value="HIGH",
+                            )
+                        ],
+                    ),
+                ],
+                date_published=datetime.datetime(2022, 1, 10, 14, 12, 0, 853348).replace(
+                    tzinfo=datetime.timezone.utc
+                ),
+            ),
+        ]
+        found_data = []
+        for response in mock_response:
+            found_data.append(parse_advisory_data(response))
+
+        assert expected_advisories == found_data
+
+    def test_fixed_filter(self):
+        assert list(
+            fixed_filter({"type": "SEMVER", "events": [{"introduced": "0"}, {"fixed": "1.6.0"}]})
+        ) == ["1.6.0"]
+        assert list(
+            fixed_filter(
+                {
+                    "type": "ECOSYSTEM",
+                    "events": [
+                        {"introduced": "0"},
+                        {"fixed": "1.0.0"},
+                        {"introduced": "0"},
+                        {"fixed": "9.0.0"},
+                    ],
+                }
+            )
+        ) == ["1.0.0", "9.0.0"]
+        assert list(
+            fixed_filter(
+                {
+                    "type": "ECOSYSTEM",
+                    "events": [
+                        {"introduced": "1.5.0"},
+                        {"fixed": "1.5.0"},
+                        {"introduced": "4.01"},
+                        {"fixed": "9.0g0"},
+                        {"introduced": "8.0.4"},
+                        {"fixed": "10.8"},
+                    ],
+                }
+            )
+        ) == ["1.5.0", "9.0g0", "10.8"]
+
+    def test_get_aliases(self):
+        assert get_aliases({"id": "GHSA-j3f7-7rmc-6wqj"}) == ["GHSA-j3f7-7rmc-6wqj"]
+        assert get_aliases({"aliases": ["CVE-2021-40831"]}) == ["CVE-2021-40831"]
+        self.assertCountEqual(
+            get_aliases({"aliases": ["CVE-2021-40831"], "id": "GHSA-j3f7-7rmc-6wqj"}),
+            [
+                "CVE-2021-40831",
+                "GHSA-j3f7-7rmc-6wqj",
+            ],
+        )
+        self.assertCountEqual(
+            get_aliases(
+                {"aliases": ["CVE-2022-22817", "GHSA-8vj2-vxx3-667w"], "id": "GHSA-j3f7-7rmc-6wqj"}
+            ),
+            ["CVE-2022-22817", "GHSA-8vj2-vxx3-667w", "GHSA-j3f7-7rmc-6wqj"],
+        )
+
+    def test_get_published_date(self):
+        assert get_published_date(
+            {"id": "GHSA-j3f7-7rmc-6wqj", "published": "2022-01-10T14:12:00Z"}
+        ) == datetime.datetime(2022, 1, 10, 14, 12, 0, 0).replace(tzinfo=datetime.timezone.utc)
+        assert get_published_date(
+            {"id": "GHSA-j3f7-7rmc-6wqj", "published": "2022-01-10T14:12:00.44444Z"}
+        ) == datetime.datetime(2022, 1, 10, 14, 12, 0, 444440).replace(tzinfo=datetime.timezone.utc)
+        assert get_published_date({"id": "GHSA-j3f7-7rmc-6wqj"}) is None
+
+    def test_get_severities(self):
+        assert list(
+            get_severities(
+                {"id": "GHSA-j3f7-7rmc-6wqj", "ecosystem_specific": {"severity": "HIGH"}}
+            )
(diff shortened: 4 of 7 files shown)


pombredanne (changes requested):
@ziadhany Thank you ++ do you mind to add a few tests?
Also may be rename this module to pysec.py?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,127 @@
...
+                system=SCORING_SYSTEMS["generic_textual"],
+                value=raw_data["ecosystem_specific"]["severity"],
+            )
+
+        references = []
+        for ref in raw_data["references"]:
+            references.append(Reference(url=ref["url"], severities=severity))
+
+        affected_package = []
+        # fixed_version = TODO
```
ziadhany, line 83: I don't know what is the best type for fixed version

pombredanne: @ziadhany The current design is to use an AffectedPackage: with the fixed version set.
See https://github.com/nexB/vulnerablecode/blob/4390b6aca872cbde378516199f143fa11e64bc5a/vulnerabilities/importer.py#L116

pombredanne (changes requested):
BTW you may want to amend your commit message in https://github.com/nexB/vulnerablecode/pull/632/commits/2fb225d15ed8783e0ec6d9c86f0b8caf5d804c59
"Add PyPL OSV" could be something like:

```
Add PyPI OSV importer

Reference: https://github.com/nexB/vulnerablecode/issues/607
```

and you could squash your commits too

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,132 @@
...
+                for file_name in zip_file.namelist():
+                    with zip_file.open(file_name) as f:
+                        vul_info = json.loads(f.read())
+                        yield parse_advisory_data(vul_info)
+        except requests.exceptions.RequestException:
+            logger.error(f"Failed to fetch osv-vulnerabilities PyPI: HTTP ")
+
+
+def parse_advisory_data(raw_data: dict) -> AdvisoryData:
+    summary = raw_data["summary"] if "summary" in raw_data else raw_data["details"][slice(50)]
```
TG1999, line 63: what if "details" is not in `raw_data` ?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,132 @@
...
+    aliases = raw_data["aliases"] if "aliases" in raw_data else raw_data["id"]
```
TG1999, line 64: what if "id" is not in `raw_data` ?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,132 @@
...
+
+    severity = []
+    if "ecosystem_specific" in raw_data and "severity" in raw_data["ecosystem_specific"]:
+        severity = VulnerabilitySeverity(
+            system=SCORING_SYSTEMS["generic_textual"],
+            value=raw_data["ecosystem_specific"]["severity"],
+        )
+
+    references = []
+    for ref in raw_data["references"]:
```
TG1999, line 78: what if "references" is not in raw_data ?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,132 @@
...
+        references.append(Reference(url=ref["url"], severities=severity))
+
+    affected_package = []
+    # HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
+    for affected_pkg in raw_data["affected"]:
```
TG1999, line 83: what if "affected" is not in `raw_data` ?

TG1999 (changes requested):
Thank You for your contribution, please check for items before accessing them, if they are not there log it, please check alpine importer for reference. I have added some comments for your consideration.

TG1999:
@ziadhany Please rebase your branch as well :)

vulnerabilities/importers/pysec.py
ziadhany: The id and modified fields are required. All other fields are optional . According to schema
https://ossf.github.io/osv-schema/#examples

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,177 @@
...
+                        fixed_version={PypiVersion("1.7.0")},
+                    ),
+                    AffectedPackage(
+                        package=PackageURL(type="npm", name="aws-iot-device-sdk-v2"),
+                        affected_version_range=None,
+                        fixed_version={SemverVersion("1.6.0")},
+                    ),
+                    AffectedPackage(
+                        package=PackageURL(
+                            type="Maven",
```
Hritik14, line 103: Please ensure type is small case always.
For eg:
```
GHSA-743r-5g92-5vgf.json:        "purl": "pkg:maven/software.amazon.awssdk.iotdevicesdk/aws-iot-device-sdk"
```

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,177 @@
...
+                        url="https://github.com/aws/aws-iot-device-sdk-cpp-v2", severities=[]
+                    ),
+                ],
+                date_published=datetime.datetime(2021, 11, 24, 20, 35, 3, 0).replace(
+                    tzinfo=datetime.timezone.utc
+                ),
+            )
+        ]
+        found_data = []
+        for i in range(len(self.mock_response)):
```
Hritik14, line 174:
```suggestion
        for response in self.mock_response:
```

Hritik14 (changes requested):
Thank you for researching on OSV! There are a few points I'd like you to consider

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,180 @@
...
+    else:
+        summary = ""
+        logger.error(f"summary not found- {raw_data['id'] !r}")
+
+    aliases = raw_data["aliases"] if "aliases" in raw_data else raw_data["id"]
+
+    if raw_data["published"] is not None:
+        date_published = datetime.strptime(
+            raw_data["published"][0:19], "%Y-%m-%dT%H:%M:%S"
+        ).replace(tzinfo=timezone.utc)
```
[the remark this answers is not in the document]
ziadhany: RFC3339 format could have a float number in the end like this '2021-02-27T00:00:29.109256Z'

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,180 @@
...
+                            )
+                            list_fixed = [i["fixed"] for i in filter_fixed]
+                            for i in list_fixed:
+                                try:
+                                    fixed_version.add(PypiVersion(i))
+                                except InvalidVersion:
+                                    logger.error(
+                                        f"fixed_version InvalidVersion - PypiVersion - {raw_data['id'] !r}"
+                                    )
+
```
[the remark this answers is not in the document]
ziadhany: 1. Check range type 
2. filter by fixed keys  {{"fixed": "2.12.21"},{"fixed": "2.13.11"}}
3. get the values of fixed keys {"2.12.21, "2.13.11"}
4. for loop to fixed_version add Version  

 ex:  PYSEC-2014-74   
```
 "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            {
              "introduced": "0"
            },
            {
              "fixed": "2.12.21"
            },
            {
              "introduced": "2.13"
            },
            {
              "fixed": "2.13.11"
            }
          ]
        }
      ]
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+                purl = PackageURL(type=package["ecosystem"], name=package["name"])
+            else:
+                purl = ""
+                logger.error(f"purl affected_pkg not found - {raw_data['id'] !r}")
+
+            affected_version_range = None
+            if "versions" in affected_pkg:
+                affected_version_range = (
+                    # HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
+                    PypiVersionRange(affected_pkg["versions"][slice(200)])
```
ziadhany, line 126: I use `slice()` to get ride of this error , Any Suggestion ?  
Screenshot from 2022-03-08 18-30-52

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,180 @@
...
+            affected_version_range = (
+                PypiVersionRange(affected_pkg["versions"][slice(50)])
+                if "versions" in affected_pkg
+                else None
+            )
+
+            if "ranges" in affected_pkg:
+                for fix_range in affected_pkg["ranges"]:
+                    fixed_version = set()
+                    if "type" in fix_range:
```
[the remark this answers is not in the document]
pombredanne: agreed... but careful as either way means that you silently skip fixes. We cannot skip things silently... especially fxies. I would rather have a big exception thrown in such case

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+                    # HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
+                    PypiVersionRange(affected_pkg["versions"][slice(200)])
+                )
+
+            if "ranges" in affected_pkg:
+                for fix_range in affected_pkg["ranges"]:
+                    fixed_version = set()
+                    if "type" in fix_range:
+                        if fix_range["type"] == "ECOSYSTEM":
+                            filter_fixed = list(
```
pombredanne, line 134: This code is too complicated for me to understand... can you find a way to decompose this in something simpler?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+                affected_version_range = (
+                    # HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
+                    PypiVersionRange(affected_pkg["versions"][slice(200)])
+                )
+
+            if "ranges" in affected_pkg:
+                for fix_range in affected_pkg["ranges"]:
+                    fixed_version = set()
+                    if "type" in fix_range:
+                        if fix_range["type"] == "ECOSYSTEM":
```
pombredanne, line 133: Extracting ` fix_range["type"]` in a variable will help making the code more readable.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+                            filter_fixed = list(
+                                filter(lambda x: x.keys() == {"fixed"}, fix_range["events"])
+                            )
+                            list_fixed = [i["fixed"] for i in filter_fixed]
+                            for i in list_fixed:
+                                try:
+                                    fixed_version.add(PypiVersion(i))
+                                except InvalidVersion:
+                                    logger.error(
+                                        f"fixed_version InvalidVersion - PypiVersion - {raw_data['id'] !r}"
```
pombredanne, line 143: There is not enough detail in this message to act on this:
What what the version? what was the InvalidVersion error message?

vulnerabilities/importers/pysec.py
pombredanne: Also, extract `raw_data["id"]` in a variable named "vulnerability_id" or similar and use this afterwards everywhere below.
In general try to create named variables early to improve readability and avoid having to make gets (and possibly checks) over and over again.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+                                    )
+
+                        if fix_range["type"] == "SEMVER":
```
pombredanne, line 146: How would it be possible for a PyPI range to use semver?
Also this code is too complicated for me to understand as is the same code above... can you find a way to decompose this in something simpler? Also it is duplicated almost exactly with the code above... so creating a function would make sense to avoid duplicated code.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+
+                            for i in list_fixed:
+                                try:
+                                    fixed_version.add(SemverVersion(i))
+                                except InvalidVersion:
+                                    logger.error(
+                                        f"fixed_version InvalidVersion - SemverVersion - {raw_data['id'] !r}"
+                                    )
+
+                        # TODO add GitHubVersion univers fix_version
```
pombredanne, line 160: Not sure what this TODO means... but it could be best to create an issue and link the issue instead?

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,292 @@
...
+    def test_to_advisories(self):
+        expected_advisories = [
+            AdvisoryData(
+                aliases=["CVE-2021-40831"],
+                summary="Improper certificate management in AWS IoT Device SDK v2",
+                affected_packages=[
+                    AffectedPackage(
+                        package=PackageURL(type="PyPI", name="awsiotsdk"),
+                        affected_version_range=PypiVersionRange(
+                            [
```
pombredanne, line 56: For readability, please create a variable with this long list outside before using it here.

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,292 @@
...
+from vulnerabilities.importer import Reference
+from vulnerabilities.importers.pysec import parse_advisory_data
+
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+TEST_DATA = os.path.join(BASE_DIR, "test_data/pysec", "pysec_test.json")
+
+
+class TestPyPLImporter(TestCase):
+    @classmethod
+    def setUpClass(cls):
```
pombredanne, line 43: Avoid setup entirely if possible. Fold this code in the test method.

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,292 @@
...
+                        url="https://github.com/advisories/GHSA-8vj2-vxx3-667w", severities=[]
+                    ),
+                ],
+                date_published=datetime.datetime(2022, 1, 10, 14, 12, 0, 0).replace(
+                    tzinfo=datetime.timezone.utc
+                ),
+            ),
+        ]
+        found_data = []
+        for response in self.mock_response:
```
pombredanne, line 289: Why do you use a mock response?
This may be OK for integration testing, but we need unit tests.
You could have a much simpler testing by testing simply one item at a time in parse_advisory_data() and have multiple test functions, one for each specific case that needs to be checked

pombredanne (changes requested):
Thank you ++ for the updates!
I added some comments for your kind consideration.

vulnerabilities/importers/pysec.py
ziadhany: django.db.utils.OperationalError: index row size 2720 exceeds btree version 4 maximum 2704 for index "vulnerabilities_advisory_aliases_summary_affected_d8f3d9e1_uniq"
DETAIL:  Index row references tuple (11064,3) in relation "vulnerabilities_advisory".
HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
Consider a function index of an MD5 hash of the value, or use full text indexing.

vulnerabilities/importers/pysec.py
ziadhany: ex: `GHSA-2ghc-6v89-pw9j.json ` 
https://osv.dev/vulnerability/GHSA-2ghc-6v89-pw9j

vulnerabilities/importers/pysec.py
pombredanne: Thanks!... so we have a model problem then... 

Which record does trigger this issue? which upstream advisory URL has this?

The solution suggested by PostgreSQL may be OK:

`Consider a function index of an MD5 hash of the value, or use full text indexing.`

My guess or my recall is that there is a very long list of enumerated affected versions...If this makes the DB choke as we have an index on this, there multiple solutions alright. All would involved changing the unique together and not use the full field details as a text blob.

1. add a new field for a checksum (or some DB index/constraint function without a field?) that would be used for the unique together and would always be small. This needs good design as it requires recomputing when the value is updated... unless we would never update this imported data field. 
2. change the design to create a new model to store these in a table. This may still require 1. but the data would never exceed any size limit of a text field. This may be the right time 
3. treat this as a one time exception: my guess is that this is a long enumeration of ALL the versions of a package up to a point in time. This could be instead treated as a range with a one-off hardcoded hack.
4. do nothing just now: just log the problem  and enter an issue to revisit this  

Not sure which one is best. In anycase, we never want to store truncated data.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,178 @@
...
+from vulnerabilities.importer import Importer
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+
+logger = logging.getLogger(__name__)
+
+
+class PyPIImporter(Importer):
+    spdx_license_expression = "Apache-2.0"
```
pombredanne, line 48:
```suggestion
    spdx_license_expression = "CC-BY-4.0"
```

vulnerabilities/importers/pysec.py
pombredanne: @ziadhany Thanks ... so in https://osv.dev/vulnerability/GHSA-2ghc-6v89-pw9j I see  three different affected packages:

- `npm/body-parser-xml` which would become the purl `pkg:npm/body-parser-xml` and is reported as SEMVER versioning which would become the `npm` vers scheme on our side. This is found at https://www.npmjs.com/package/body-parser-xml

- `NuGet/body-parser-xml-nuget` which would become the purl `pkg:nuget/body-parser-xml-nuget` and is reported as ECOSYSTEM versioning and would become the `nuget` vers scheme on our side... BUT this does NOT exist as a nuget package at all. https://www.nuget.org/packages?q=body-parser-xml ... not even close

- `PyPI/body-parser-xml-pip` which would become the purl `pkg:nuget/body-parser-xml-pip` and is reported as ECOSYSTEM versioning and would become the `pypi` vers scheme on our side. BUT this does NOT exist as a pypi package at all: https://pypi.org/search/?q=body-parser-xml


So what to make of this? OSV and GitHub data are likely not really reviewed after all. That's not the first time I see junk ghost records like this that are supposedly reviewed.

In anycase this are not a Pypi package... but it is going to be hard to figure this out ... at least not at import time.
I note that this is not seen at all in https://github.com/pypa/advisory-database 
So practically:
- I entered https://github.com/nexB/vulnerablecode/issues/644 to add improvers to validate purls
- short term we could flag these with lower confidence but I am not sure we have enough at import time
- so we can import these ... each should use their own specific vers scheme though and PyPI is NOT semver alright

There is not much harm done to import non-existing purl BTW because these would never be found in the wild.

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
pombredanne: This would mean multiple  affected packages in our world IMHO: `< 2.12.21` and other range `>=2.13 to <2.13.11`

vulnerabilities/importers/pysec.py
ziadhany: some affected package have as fixed_version as GIT . Does [univers](https://github.com/nexB/univers) have GitVersion ?
  ```
 "ranges": [
            {
                "type": "GIT",
                "repo": "https://github.com/unicode-org/icu.git",
                "events": [
                  { "introduced": "6e5755a2a833bc64852eae12967d0a54d7adf629" },
                  { "fixed": "c43455749b914feef56b178b256f29b3016146eb" }
                ]
            }
        ]
```

Comment (pombredanne) [reply vulnerabilities/importers/pysec.py:160]: Nope ... good catch!... we need to add this. This is going to be weird versioning scheme as you cannot determine if a commit is within a range, or before or after another commit using only the commit information.... You need a full git log!
As a first step, any such comparison can return a NotImplemented exception.
Later we can devise how to best handle these

File: vulnerabilities/importers/pysec.py [hunk 21]
```diff
@@ -0,0 +1,206 @@
...
+    return AdvisoryData(
+        aliases=aliases,
+        summary=summary,
+        affected_packages=affected_package,
+        references=references,
+        date_published=date_published,
+    )
+
+
+def fixed_filter(fixed_range) -> []:
```
TG1999, line 106: do add some doctest

TG1999 (commented):
Thanks Ziad, for your changes, I have added some comments for your consideration

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,206 @@
...
+    list_fixed = [i["fixed"] for i in filter_fixed]
+    return list_fixed
+
+
+def get_aliases(raw_data) -> []:
+    vulnerability_id = raw_data.get("id")
+    vulnerability_aliases = raw_data.get("aliases") or []
+    if vulnerability_id:
+        vulnerability_aliases.append(vulnerability_id)
+    return list(dict.fromkeys(vulnerability_aliases))
```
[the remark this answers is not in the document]
ziadhany: because aliases field is optional ,  id is require and these are all aliases from our perspective

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,206 @@
...
+
+def parse_advisory_data(raw_data: dict) -> AdvisoryData:
+    summary = raw_data.get("summary") or ""
+    aliases = get_aliases(raw_data)
+    date_published = get_published_date(raw_data)
+    severity = get_severity(raw_data)
+    references = get_references(raw_data, severity)
+
+    affected_package = []
+    if "affected" in raw_data:
```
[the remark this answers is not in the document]
ziadhany: Line 94:
```
   else:
        logger.error(f"affected_package not found - {raw_data['id'] !r}") 
```

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
TG1999: I got your point, but why are we converting it to a dict first and then converting it to a list, why can't we simply return it here ?

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
TG1999: @ziadhany sorry for confusing you, I meant you can check inverse of that expression, this will save us from having a big code of block indented.
```
if not "affected" in raw_data:
    logger.error(f"affected_package not found - {raw_data['id'] !r}")
    # return the advisory with affected_package as an empty list 
---process affected and return advisory with affected_package from the data you got after processing ----
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,206 @@
...
+        return PackageURL(type=package["ecosystem"], name=package["name"])
+    else:
+        logger.error(f"purl affected_pkg not found - {raw_id !r}")
+
+
+def get_aff_version_range(affected_pkg, raw_id):
+    # HINT: Values larger than 1/3 of a buffer page cannot be indexed.
+    if "versions" in affected_pkg and len(affected_pkg["versions"]) < 200:
+        try:
+            return PypiVersionRange(affected_pkg["versions"])
```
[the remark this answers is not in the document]
pombredanne: @TG1999 good point... @ziadhany can you provide an example of what is the value of  `affected_pkg["versions"]` ?

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
ziadhany: to make sure every aliases in the list is unique and maintaining order 
 ex :      
```
assert get_aliases(
            {"aliases": ["CVE-2022-22817", "GHSA-8vj2-vxx3-667w"], "id": "GHSA-8vj2-vxx3-667w"}
        ) == ["CVE-2022-22817", "GHSA-8vj2-vxx3-667w"]
```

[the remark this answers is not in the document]
TG1999:
```suggestion
    return list(set(vulnerability_aliases))
```
Also add one comment above this return statement stating why are we doing this

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
ziadhany: > @ziadhany there is a field called "database_specific" in test data, do see if we can derive some severities from there

the problem is that “database_specific”,contains a JSON object with unspecified fields. 

npm:
```
 "database_specific": {
        "CWE": "CWE-327",
    }
```
but Python vulnerability :  
```
"database_specific": {
        "source": "https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2018/11/GHSA-2jc8-4r6g-282j/GHSA-2jc8-4r6g-282j.json"
      }
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,206 @@
...
+    if "severity" in raw_data:
+        for sever_list in raw_data["severity"]:
+            if "type" in sever_list and sever_list["type"] == "CVSS_V3":
+                severity.append(
+                    VulnerabilitySeverity(
+                        system=SCORING_SYSTEMS["cvssv3_vector"],
+                        value=sever_list["score"],
+                    )
+                )
+    if "ecosystem_specific" in raw_data and "severity" in raw_data["ecosystem_specific"]:
```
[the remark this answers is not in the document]
ziadhany: > @ziadhany there is a field called "database_specific" in test data, do see if we can derive some severities from there


the problem is that “database_specific”,contains a JSON object with unspecified fields. 

npm:
```
 "database_specific": {
        "CWE": "CWE-327",
    }
```
but Python vulnerability :  
```
"database_specific": {
        "source": "https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2018/11/GHSA-2jc8-4r6g-282j/GHSA-2jc8-4r6g-282j.json"
      }
```

ziadhany:
> @ziadhany there is a field called "database_specific" in test data, do see if we can derive some severities from there

Do we support CWE Scoring System ?

vulnerabilities/importers/pysec.py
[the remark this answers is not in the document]
TG1999: @ziadhany , you can use your code, we can have an improver later on to verify if these VersionConstraints are same or not

requirements.txt
```diff
@@ -20,3 +20,6 @@ defusedxml==0.7.1
 license-expression>=21.6.14
 
 Markdown==3.3.4
+
+
+dateparser~=1.1.1
```
pombredanne, line 25: Please either pin requirements exactly or use a minimum version here... but no tilde ranges. They are evil as they imply an upper bound.  
Also please remove all empty lines from there

pombredanne:
> > @ziadhany there is a field called "database_specific" in test data, do see if we can derive some severities from there
> 
> @TG1999 Do we support CWE Scoring System ?

We do not support CWE and this would not be a Scoring System but something separate, more like a categorization since there is no score attached to it. I created https://github.com/nexB/vulnerablecode/issues/651 to track this separately!

pombredanne (commented):
Hey! Thank for your update and pushing this through!
There are a few feedback that still need some input like the ones about doctests?

TG1999:
@ziadhany ,please rebase your branch, the posgresql index issue is solved now

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,219 @@
...
+
+def get_aff_version_range(affected_pkg, raw_id, purl):
+    if "versions" in affected_pkg:
+        if purl.type == "pypi":
+            try:
+                return PypiVersionRange(affected_pkg["versions"])
+            except Exception as e:
+                logger.error(f"affected_pkg_version_range Error - {raw_id !r} - purl: {e !r}")
+        else:
+            logger.error(f"NotImplementedError Ecosystem fixed range - {raw_id !r} ")
```
Hritik14, line 191: As this is PyPI only importer, how about just doing 
```
                 return PypiVersionRange(affected_pkg["versions"])
```
and not checking explicitly for pypi

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def get_fixed_version(fixed_range, raw_id, purl) -> {}:
+    fixed_version = set()
```
Hritik14, line 195: Is this set only for deduplication ?
If so, how about defining this in ``helpers.py``
```py
def dedupe(original: List) -> List:
    newlist = []
    for i in original:
      if i not in newlist:
         newlist.append(i)
```

This will preserve order and only dedupe (as it should)

Hritik14 (changes requested):
PR is already looking great. I've added my suggestions, please have a look.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,180 @@
...
+        summary = raw_data["summary"]
+    elif "details" in raw_data:
+        summary = raw_data["details"][slice(50)]
+    else:
+        summary = ""
+        logger.error(f"summary not found- {raw_data['id'] !r}")
+
+    aliases = raw_data["aliases"] if "aliases" in raw_data else raw_data["id"]
+
+    if raw_data["published"] is not None:
```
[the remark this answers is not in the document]
pombredanne: I think this is good to fail for now if this is not there. We can adjust as we go

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,219 @@
...
+    ['1.6.0']
+    >>> fixed_filter({"type": "ECOSYSTEM","events":[{"introduced": "0"},{"fixed": "1.0.0"},{"fixed": "9.0.0"}]})
+    ['1.0.0', '9.0.0']
+    """
+    filter_fixed = list(filter(lambda x: x.keys() == {"fixed"}, fixed_range["events"]))
+    list_fixed = [i["fixed"] for i in filter_fixed]
+    return list_fixed
+
+
+def get_aliases(raw_data) -> []:
```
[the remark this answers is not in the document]
pombredanne: doctest or test ... any test ;) will be welcomed

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    aliases field is optional , id is required and these are all aliases from our perspective
+    converting list of two fields to a dict then , convert it to a list to make sure a list is unique
+    """
+    vulnerability_id = raw_data.get("id")
+    vulnerability_aliases = raw_data.get("aliases") or []
```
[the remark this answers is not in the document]
pombredanne: @Hritik14 FWIW, I kinda like `raw_data.get("aliases") or []` as this will gte you a list for sure.
`raw_data.get("aliases", [])` can get you a `None` if `aliases` is `None`, or an empty string or an empty dict or any emptyish value that is set.

Hritik14:
@ziadhany Feel free to ``Resolve Conversation`` after you commit that in your codebase, it'd help us to reduce clutter.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,219 @@
...
+    if vulnerability_id:
+        vulnerability_aliases.append(vulnerability_id)
+    return list(set(vulnerability_aliases))
```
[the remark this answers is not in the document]
pombredanne: > I don't think so but I just added it for extra validation

Using sets can be a source many small issues... If anything the ordering is not stable. So IMHO:
1. do not de-duplicate unless you see dupes in practice
2. OR if you see dupes at least return a sorted result using rather `return sorted(set(vulnerability_aliases))`
3. BUT in many cases when you de-duplicate it is best to preserve the original ordering which makes using set not practical.

IMHO for now either use 1. or 2.

pombredanne:
I merged some other code... so you  have some merge conflicts to resolve now

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,489 @@
...
+    def test_get_published_date(self):
+        assert get_published_date(
+            {"id": "GHSA-j3f7-7rmc-6wqj", "published": "2022-01-10T14:12:00Z"}
+        ) == datetime.datetime(2022, 1, 10, 14, 12, 0, 0).replace(tzinfo=datetime.timezone.utc)
+        assert get_published_date(
+            {"id": "GHSA-j3f7-7rmc-6wqj", "published": "2022-01-10T14:12:00.44444Z"}
+        ) == datetime.datetime(2022, 1, 10, 14, 12, 0, 444440).replace(tzinfo=datetime.timezone.utc)
+        assert get_published_date({"id": "GHSA-j3f7-7rmc-6wqj"}) is None
+
+    def test_get_severity(self):
```
pombredanne, line 335: That's not essential here, but creating multiple tests is never a bad idea

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,489 @@
...
+        assert get_severity(
+            {
+                "id": "PYSEC-2022-10",
+                "severity": [
+                    {"type": "CVSS_V3", "score": "CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:N"}
+                ],
+            }
+        ) == [
+            VulnerabilitySeverity(
+                system=SCORING_SYSTEMS["cvssv3_vector"],
```
pombredanne, line 350: Are you sure this may not be CVSS:3.1 and `cvssv3.1_vector`  instead? https://github.com/nexB/vulnerablecode/blob/47f6ae62d919fff5094b960c1f56bc0c6b8b4be3/vulnerabilities/severity_systems.py#L83

vulnerabilities/tests/test_pysec.py
```diff
@@ -0,0 +1,489 @@
...
+            ),
+        ]
+
+    def test_get_references(self):
+        assert get_references(
+            {
+                "id": "PYSEC-2022-10",
+                "references": [
+                    {
+                        "type": "FIX",
```
pombredanne, line 380: Should we also track this "type" as a new model field? If you think so, please create a follow up issue to solve this later.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,220 @@
...
+def get_references(raw_data, severity) -> []:
+    if "references" in raw_data:
+        return [
+            Reference(url=ref["url"], severities=severity) for ref in raw_data["references"] if ref
+        ]
+    else:
+        return []
+
+
+def get_aff_purl(affected_pkg, raw_id):
```
pombredanne, line 175: What about this? Spelling out whole words is fine

```suggestion
def get_affected_purl(affected_pkg, raw_id):
    """
    Return an affected PackageURL given an ``affected_pkg`` mapping and a ``raw_id``
    """
```


BTW, why do you need a raw_id there?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,220 @@
...
+        except ValueError:
+            logger.error(f"PackageURL ValueError - {raw_id !r} - purl: {package['purl'] !r}")
+
+    if "ecosystem" in package and "name" in package:
+        return PackageURL(type=package["ecosystem"], name=package["name"])
+    else:
+        logger.error(f"purl affected_pkg not found - {raw_id !r}")
+
+
+def get_aff_version_range(affected_pkg, raw_id):
```
pombredanne, line 189: Same comment as above...
```suggestion
def get_affected_version_range(affected_pkg, raw_id):
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,220 @@
...
+    if "versions" in affected_pkg:
```
pombredanne, line 190: What about something like this instead?
```
    affected_versions = affected_pkg.get("versions")
    if affected_versions:
        try:
            return PypiVersionRange(affected_versions)

File: vulnerabilities/importers/pysec.py [hunk 39]
```diff
@@ -0,0 +1,220 @@
...
+        try:
+            return PypiVersionRange(affected_pkg["versions"])
+        except Exception as e:
```
pombredanne, line 193: What could be the exceptions raised here?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,220 @@
...
+    else:
+        logger.error(f"affected_pkg_version_range not found - {raw_id !r} ")
+
+
+def get_fixed_version(fixed_range, raw_id) -> []:
+    fixed_version = []
+    if "type" in fixed_range:
+        list_fixed = fixed_filter(fixed_range)
+        for i in list_fixed:
+            if fixed_range["type"] == "ECOSYSTEM":
```
pombredanne, line 204: Always prefer to create a named variable in such case: 
```suggestion
            fixed_range_type = fixed_range["type"]
            if fixed_range_type == "ECOSYSTEM":
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,220 @@
...
+    if "versions" in affected_pkg:
+        try:
+            return PypiVersionRange(affected_pkg["versions"])
+        except Exception as e:
+            logger.error(f"affected_pkg_version_range Error - {raw_id !r} - purl: {e !r}")
+    else:
+        logger.error(f"affected_pkg_version_range not found - {raw_id !r} ")
+
+
+def get_fixed_version(fixed_range, raw_id) -> []:
```
pombredanne, line 199:
```suggestion
def get_fixed_version(fixed_range, raw_id) -> []:
    """"
    Return a list of ....
    
```

pombredanne (changes requested):
Looking great!
I am nitpicking on a few extra things because I am picky!, but the code overall looks nice as it is.

Some general remarks:
- create variables early rather than looking up a dict over and over
- a small docstring to explain what a function does is not mandatory but useful for no-trivial, non-obvious functions.

vulnerabilities/tests/test_pysec.py
ziadhany: No, I'm not sure.
but have a look at osv-schema:
https://ossf.github.io/osv-schema/#severitytype-field
Severity Type: CVSS_V3
Score Description : **A CVSS vector string** ...

ex GHSA-xc3p-ff3m-f46v.json:
"score": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:N"

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+    references = get_references(raw_data, severity)
+
+    affected_packages = []
+    if not "affected" in raw_data:
+        logger.error(f"affected_packages not found - {raw_data['id'] !r}")
+
+    if "affected" in raw_data:
+        for affected_pkg in raw_data["affected"]:
+            purl = get_affected_purl(affected_pkg, raw_data["id"])
+            if purl.type == "pypi":
```
pombredanne, line 82:
```suggestion
            if purl.type != "pypi":
                logger.error(f"Non PyPI package found in PYSEC advisories: {purl} - from: {raw_data!r}")
            else:
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+                affected_version_range = get_affected_version_range(affected_pkg, raw_data["id"])
+                for fixed_range in affected_pkg.get("ranges", []):
+                    fixed_version = get_fixed_version(fixed_range, raw_data["id"])
```
pombredanne, line 85:
```suggestion
                    fixed_version = get_fixed_version(fixed_range, vulnid)
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+        return parse(raw_data["published"])
+    else:
+        logger.warning(f"date_published not found {raw_data['id'] !r}")
+
+
+def get_severity(raw_data) -> []:
+    severity = []
+    if "severity" in raw_data:
+        for sever_list in raw_data["severity"]:
+            if "type" in sever_list and sever_list["type"] == "CVSS_V3":
```
pombredanne, line 146: This may be a bit more compact?
```suggestion
            if sever_list.get("type") == "CVSS_V3":
```

Also what if there is another severity that we do not know how to handle? this should be logged.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+                severity.append(
+                    VulnerabilitySeverity(
+                        system=SCORING_SYSTEMS["cvssv3.1_vector"],
+                        value=sever_list["score"],
+                    )
+                )
+    if "ecosystem_specific" in raw_data and "severity" in raw_data["ecosystem_specific"]:
```
pombredanne, line 153: Same comment as above: get a value in a variable early and use that afterwards rather than re-doing lookups in the mappings multiple times. This help with readability and makes the code more robust.

```suggestion
    ecosys = raw_data.get("ecosystem_specific") or {}
    sever = ecosys.get("severity")
    if sever:
        severity.append(
            VulnerabilitySeverity(
                system=SCORING_SYSTEMS["generic_textual"],
                value=value,
            )
        )    
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+            )
+        )
+    else:
+        logger.warning(f"severity not found- {raw_data['id']!r}")
+
+    return severity
+
+
+def get_references(raw_data, severity) -> []:
+    if "references" in raw_data:
```
pombredanne, line 167: You replace the whole function body by:
```suggestion
    references = raw_data.get("references") or []
    return [Reference(url=ref["url"], severities=severity) for ref in references if ref]
```
Also what is `severity` here? a list since it afterward assigned to `severities`? 
how does it apply to the whole list of references and not just to a single one?

Do you have an example of data?

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+
+def get_affected_purl(affected_pkg, raw_id):
+    package = affected_pkg["package"]
+    if "purl" in package:
+        try:
+            return PackageURL.from_string(package["purl"])
+        except ValueError:
+            logger.error(f"PackageURL ValueError - {raw_id !r} - purl: {package['purl'] !r}")
+
+    if "ecosystem" in package and "name" in package:
```
pombredanne, line 183: Same nit as above: you are doing two lookups in package: here and below on line 184.
You should  collect ecosystem and name in variables once instead

vulnerabilities/importers/pysec.py
pombredanne: I am suggested to add a nit of docstring that explains what this function does too ...

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+    []
+    >>> get_fixed_version({"type": "ECOSYSTEM", "events": [{"fixed": "1.7.0"}]}, "GHSA-j3f7-7rmc-6wqj")
+    [PypiVersion(string='1.7.0')]
+    """
+
+    fixed_version = []
+    if "type" in fixed_range:
+        list_fixed = fixed_filter(fixed_range)
+        for i in list_fixed:
+            fixed_range_type = fixed_range["type"]
```
pombredanne, line 212: Should this be outside of the loop rather?

pombredanne (changes requested):
See a few final nit pickings for your kind consideration.
As a general note, think about creating a variable early and then using this rather than doing multiple lookups in a mapping: this is more readable and more robust.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,229 @@
...
+        aliases=aliases,
+        summary=summary,
+        affected_packages=affected_packages,
+        references=references,
+        date_published=date_published,
+    )
+
+
+def fixed_filter(fixed_range) -> []:
+    """
```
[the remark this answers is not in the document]
ziadhany: yes , this function is just return a list of fixed version strings

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+    Return a list of fixed version strings given a ``fixed_range`` mapping of OSV data.
+    >>> fixed_filter({"type": "SEMVER", "events": [{"introduced": "0"}, {"fixed": "1.6.0"}]})
+    ['1.6.0']
+    >>> fixed_filter({"type": "ECOSYSTEM","events":[{"introduced": "0"},{"fixed": "1.0.0"},{"fixed": "9.0.0"}]})
+    ['1.0.0', '9.0.0']
+    """
+    filter_fixed = list(filter(lambda x: x.keys() == {"fixed"}, fixed_range["events"]))
```
pombredanne, line 118: What about this instead? It may be easier to read than a filter with a lambda?
```
    for event in fixed_range.get("events", []):
        fixed = event.get("fixed")
        if fixed:
            yield fixed
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+#  VulnerableCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/vulnerablecode/ for support and download.
+import json
+import logging
+from io import BytesIO
+from typing import Iterable
+from zipfile import ZipFile
+
+import requests
+from dateparser import parse
```
pombredanne, line 29: Since parse is a rather generic name, it may be best to import the module? 
```suggestion
import dateparser
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+    vulnerability_id = raw_data.get("id")
+    vulnerability_aliases = raw_data.get("aliases") or []
+    if vulnerability_id:
+        vulnerability_aliases.append(vulnerability_id)
+    return vulnerability_aliases
+
+
+def get_published_date(raw_data):
+    if "published" in raw_data:
+        return parse(raw_data["published"])
```
pombredanne, line 143: Rather then test and get, I prefer the simpler:
```
published = raw_data.get("published")
return published and parse(published)
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+    else:
+        logger.warning(f"date_published not found {raw_data['id'] !r}")
```
pombredanne, line 145: I am not sure the date is important enough to warrant logging

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def get_severity(raw_data) -> []:
+    severities = []
+    if "severity" in raw_data:
+        for sever_list in raw_data["severity"]:
```
pombredanne, line 151:
```suggestion
        for sever_list in raw_data.get("severity") or []:
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+            if sever_list.get("type") == "CVSS_V3":
+                severities.append(
+                    VulnerabilitySeverity(
+                        system=SCORING_SYSTEMS["cvssv3.1_vector"],
+                        value=sever_list["score"],
+                    )
+                )
+            else:
+                logger.error(f"NotImplementedError severity type- {raw_data['id']!r}")
```
pombredanne, line 160: That's OK not to log this. It is not an error not to have one IMHO

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def get_published_date(raw_data):
+    if "published" in raw_data:
+        return parse(raw_data["published"])
+    else:
+        logger.warning(f"date_published not found {raw_data['id'] !r}")
+
+
+def get_severity(raw_data) -> []:
```
pombredanne, line 148: get_severity --> get_severities since there are more than one

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,252 @@
...
+    severities = []
```
pombredanne, line 149: IMHO you should rather yield as you go instead of accumulating in a list. The code will be leaner then

pombredanne (changes requested):

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+    return vulnerability_aliases
+
+
+def get_published_date(raw_data):
+    published = raw_data.get("published")
+    return published and dateparser.parse(published)
+
+
+def get_severities(raw_data) -> []:
+    if "severity" in raw_data:
```
pombredanne, line 151: This is no longer needed since the code below (`for sever_list in raw_data.get("severity") or []:` handles all the possible cases:
```suggestion
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+def get_affected_purl(affected_pkg, raw_id):
+    package = affected_pkg.get("package") or {}
+    purl = package.get("purl")
+    if purl:
+        try:
+            return PackageURL.from_string(purl)
+        except ValueError:
+            logger.error(f"PackageURL ValueError - {raw_id !r} - purl: {purl !r}")
+
+    ecosys = package.get("ecosystem")
```
pombredanne, line 191: I think there is a need to translate the "ecosystem" to the purl type as the values used by OSV are not exactly that of purl types.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+                        yield parse_advisory_data(vul_info)
+
+
+def parse_advisory_data(raw_data: dict) -> AdvisoryData:
+    summary = raw_data.get("summary") or ""
+    aliases = get_aliases(raw_data)
+    date_published = get_published_date(raw_data)
+    severity = list(get_severities(raw_data))
+    references = get_references(raw_data, severity)
+
```
pombredanne, line 77: IMHO you should extract `raw_data['id']` in a named variable and used it throughout below afterwards.

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+    affected_packages = []
+    if "affected" not in raw_data:
+        logger.error(f"affected_packages not found - {raw_data['id'] !r}")
```
pombredanne, line 80: It is always best to return early if there is nothing left to do.
This way you can de-indent the code that comes after
```suggestion
        logger.error(f"affected_packages not found - {raw_data['id'] !r}")
        return
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+
+    if "affected" in raw_data:
```
pombredanne, line 82: Following from above, this becomes not needed: 
```suggestion
    if "affected" in raw_data:
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+        for affected_pkg in raw_data["affected"]:
```
pombredanne, line 83: affected may not be a list, so to be safe, I would use this: 
```suggestion
        for affected_pkg in raw_data.get("affected") or []:
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+            purl = get_affected_purl(affected_pkg, raw_data["id"])
```
pombredanne, line 84: Use a raw_id variable defined above

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+            if purl.type != "pypi":
+                logger.error(
+                    f"Non PyPI package found in PYSEC advisories: {purl} - from: {raw_data['id']!r}"
+                )
+            else:
```
pombredanne, line 89: Return early and de-indent:
```suggestion
                return
```

vulnerabilities/importers/pysec.py
```diff
@@ -0,0 +1,240 @@
...
+                vulnid = raw_data["id"]
```
pombredanne, line 90: Use raw_id variable

pombredanne (changes requested):
Thank you for the updates. A very last round few nitpickings for your consideration!

vulnerabilities/importers/pysec.py
ziadhany: yes , we translated the "ecosystem" if the osv purl not found but what is wrong with osv purl "pkg:pypi/tensorflow-gpu"
Can you explain it  more?
```
>>> PackageURL.from_string("pkg:pypi/tensorflow-gpu") == PackageURL(type="pypi", name="tensorflow-gpu")
True 
```

pombredanne (approved):
LGTM. Merging! Thank you ++ for your peseverance!r

vulnerabilities/helpers.py
```diff
@@ -267,3 +267,16 @@ def _get_gh_response(gh_token, graphql_query):
     endpoint = "https://api.github.com/graphql"
     headers = {"Authorization": f"bearer {gh_token}"}
     return requests.post(endpoint, headers=headers, json=graphql_query).json()
+
+
+def dedupe(original: List) -> List:
```
armijnhemel, line 272: You could also have implemented this using a `OrderedDict` from `collections`.

```
>>> import collections
>>> collections.OrderedDict.fromkeys(["z","i","a","a","d","d"]).keys()
odict_keys(['z', 'i', 'a', 'd'])
```

pombredanne: good point! actually on Python 3.6+, this could be even shorter: `dict.fromkeys(["z","i","a","a","d","d"]).keys()`

Issue #617, opened 2022-02-08
Fix failing nix tests


Hritik14:
Nix tests are failing. Probably some kind of dependency issue.
See more: https://github.com/nexB/vulnerablecode/runs/5094793358?check_suite_focus=true

Hritik14:
@rolfschr, As you're our nix savior, can you please look into it ?

rolfschr:
Hi @Hritik14, I can have a look into this. I see that the nix-related documentation has been removed from the readme. I guess that's ok. Still, could you please dig it up again an place it into `/etc/nix`? Thanks.

Hritik14:
Thanks @rolfschr for responding. The nix docs have just been moved to RTD at https://vulnerablecode.readthedocs.io/en/latest/installation.html#using-nix as I couldn't do a quick and neat tl;dr to place on README.
We'll be glad to accept a tl;dr PR on readme if you'd like to.

rolfschr:
@Hritik14 See #635.

Hritik14:
@rolfschr Thank you so much! We have the green ticks now.

PR #635, opened 2022-02-22
Fix nix setup

rolfschr:
This PR addresses https://github.com/nexB/vulnerablecode/issues/617. I incorporated several changes that should make the Nix setup more compatible with the rest of the code base. Mainly, the `Makefile` is now used wherever possible. Fortunately, several env vars can be used to tweak the Makefile behavior (or better: the invoked commands).


Hritik14 (approved):
thank you !

PR #642, opened 2022-03-09
Migrate github importer to importer improver model

TG1999:


vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
     "-class.-jw.util.version.Version-",
 }
 
+ECOSYSTEM_PKG_TYPE_MAPPING = {
```
pombredanne, line 125: What about this more explicit name?
```suggestion
PACKAGE_TYPE_BY_GITHUB_ECOSYSTEM = {
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    "MAVEN": "maven",
+    "NUGET": "nuget",
+    "COMPOSER": "composer",
+    "PIP": "pypi",
+    "RUBYGEMS": "gem",
+    "GO": "golang",
+}
+
+ECOSYSTEM_VERSION_RANGE_MAPPING = {
```
pombredanne, line 134: and this?

```suggestion
VERSION_RANGE_CLASS_BY_GITHUB_ECOSYSTEM = {
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    "MAVEN": MavenVersionRange,
+    "NUGET": NugetVersionRange,
+    "COMPOSER": ComposerVersionRange,
+    "PIP": PypiVersionRange,
+    "RUBYGEMS": GemVersionRange,
+    "GO": GolangVersionRange,
+}
+
+PKG_TYPE_ECOSYSTEM_MAPPING = {value: key for (key, value) in ECOSYSTEM_PKG_TYPE_MAPPING.items()}
```
pombredanne, line 143: and this (that you may want to to move up):
```suggestion
GITHUB_ECOSYSTEM_BY_PACKAGE_TYPE = {value: key for (key, value) in ECOSYSTEM_PKG_TYPE_MAPPING.items()}
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+
+QUERY = """
+        query{
```
pombredanne, line 146: Can you dedent this 8 chars?  we would not lose any redability.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+
+
 class GitHubAPIImporter(Importer):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    spdx_license_expression = "CC-BY-4.0"
+    token = os.environ["GH_TOKEN"]
+    endpoint = "https://api.github.com/graphql"
+
+    ecosystems = [
```
pombredanne, line 202: What about just using the `PACKAGE_TYPE_BY_GITHUB_ECOSYSTEM` above?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        "GO",
+    ]
+
+    def advisory_data(self) -> Iterable[AdvisoryData]:
+        headers = {"Authorization": f"token {self.token}"}
+        end_cursor_exp = ""
+        advisories = []
+        for ecosystem in self.ecosystems:
+            while True:
+                query_json = {"query": QUERY % (ecosystem, end_cursor_exp)}
```
pombredanne, line 217: May be this is enough of a name?
```suggestion
                query = {"query": QUERY % (ecosystem, end_cursor_exp)}
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+                if not get_item(page_info, "hasNextPage"):
+                    break
+                end_cursor = get_item(page_info, "endCursor")
+                if end_cursor:
+                    end_cursor = f'"{end_cursor}"'
+                    end_cursor_exp = f"after: {end_cursor}"
+        return advisories
+
+
+def get_ecosystem_from_type(type: str):
```
pombredanne, line 232: Do we need a function for this, as opposed to make a direct get call?
Same for def get_type_from_ecosystem ... it feels too much

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    return PKG_TYPE_ECOSYSTEM_MAPPING.get(type)
+
+
+def get_type_from_ecosystem(ecosystem: str):
+    return ECOSYSTEM_PKG_TYPE_MAPPING.get(ecosystem)
+
+
+def extract_references(reference_data: List):
+    """
+    Yield `reference` by itearting over `reference_data`
```
pombredanne, line 242:
```suggestion
    Yield `reference` by iterating over `reference_data`
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+            reference = Reference(url=url)
+        yield reference
+
+
+def process_name(ecosystem: str, pkg_name: str) -> Optional[Tuple[Optional[str], str]]:
+    """
+    Return `namespace` and `name` by processing `ecosystem` and `pkg_name`
+    """
+    if ecosystem == "MAVEN":
+        artifact_comps = pkg_name.split(":")
```
pombredanne, line 260: What are comps?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        if len(artifact_comps) != 2:
+            return
```
pombredanne, line 262: Does this mean we silently skip this?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        ns, name = artifact_comps
+        return ns, name
+
+    if ecosystem == "COMPOSER":
         try:
-            self.gh_token = os.environ["GH_TOKEN"]
-        except KeyError:
-            raise GitHubTokenError("Environment variable GH_TOKEN is missing")
+            vendor, name = pkg_name.split("/")
```
pombredanne, line 268: You could also use `pkg_name.partition("/")` that will never fail and always return three values.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        except ValueError:
+            # TODO log this
+            return None
+        return vendor, name
 
-    def __enter__(self):
-        self.advisories = self.fetch()
+    if ecosystem in ("NUGET", "PIP", "RUBYGEMS", "GO"):
+        return None, pkg_name
```
pombredanne, line 275: What does this None means here?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
-    def fetch(self) -> Mapping[str, List[Mapping]]:
-        headers = {"Authorization": "token " + self.gh_token}
-        api_data = {}
-        for ecosystem in self.config.ecosystems:
 
-            api_data[ecosystem] = []
-            end_cursor_exp = ""
+def process_name_namespace(ecosystem: str, name: str, namespace: str) -> str:
+    """
+    Return `namespace` and `name` by processing `ecosystem` and `pkg_name`
```
pombredanne, line 284: What's the use for this function? is this to recreate the naming convention that GitHub uses?

vulnerabilities/importers/github.py
```diff
@@ -21,69 +21,45 @@
...
 from vulnerabilities.package_managers import MavenVersionAPI
 from vulnerabilities.package_managers import NugetVersionAPI
 from vulnerabilities.package_managers import PypiVersionAPI
 from vulnerabilities.package_managers import RubyVersionAPI
-from vulnerabilities.severity_systems import scoring_systems
+from vulnerabilities.package_managers import Version
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
 
-# set of all possible values of first '%s' = {'MAVEN','COMPOSER', 'NUGET', 'RUBYGEMS', 'PYPI'}
-# second '%s' is interesting, it will have the value '' for the first request,
```
pombredanne, line 52: This would be valuable comments to keep IMHO.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+}
+
+PKG_TYPE_ECOSYSTEM_MAPPING = {value: key for (key, value) in ECOSYSTEM_PKG_TYPE_MAPPING.items()}
+
+QUERY = """
+        query{
+        securityVulnerabilities(first: 100, ecosystem: %s, %s) {
+            edges {
+            node {
+                advisory {
```
pombredanne, line 150: Are there more fields than these? can you double check and at least put in a comment the ones we do not fetch for now?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    "NUGET": NugetVersionRange,
+    "COMPOSER": ComposerVersionRange,
+    "PIP": PypiVersionRange,
+    "RUBYGEMS": GemVersionRange,
+    "GO": GolangVersionRange,
+}
+
+PKG_TYPE_ECOSYSTEM_MAPPING = {value: key for (key, value) in ECOSYSTEM_PKG_TYPE_MAPPING.items()}
+
+QUERY = """
```
pombredanne, line 145: What about this more verbose but self-explanatory?
```suggestion
GRAPHQL_VULNERBILITY_QUERY_TEMPLATE = """
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    Return `item` by going through all the `attributes` present in the `json_object`
+    """
+    if not json_object:
+        LOGGER.error(f"json_object is empty: {json_object}")
+        return None
+    item = json_object
+    for attribute in attributes:
+        if not attribute in item:
+            LOGGER.error(f"Missing attribute {attribute} in {item}")
+            return None
```
pombredanne, line 192:
```suggestion
            return
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+def get_item(json_object: dict, *attributes):
+    """
+    Return `item` by going through all the `attributes` present in the `json_object`
+    """
+    if not json_object:
+        LOGGER.error(f"json_object is empty: {json_object}")
+        return None
+    item = json_object
+    for attribute in attributes:
+        if not attribute in item:
```
pombredanne, line 190:
```suggestion
        if attribute not in item:
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        }
+        }
+        """
+
 
 class GitHubTokenError(Exception):
     pass
 
 
+def get_item(json_object: dict, *attributes):
```
pombredanne, line 181: This is not specifically JSON, correct? just some plain data dictionary that was loaded from JSON... but may be we do not need to have JSON in its name at this stage?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+            return None
+        item = item[attribute]
+    return item
+
+
 class GitHubAPIImporter(Importer):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    spdx_license_expression = "CC-BY-4.0"
+    token = os.environ["GH_TOKEN"]
```
pombredanne, line 199: Could this be made more robust? this will crash on import if GH_TOKEN is not defined.
IMHO put this in a method instead and raise a proper exception if not present only when needed.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
-        if ecosystem == "COMPOSER":
+
+def process_response(resp: dict, ecosystem: str) -> Iterable[AdvisoryData]:
+    """
+    Yield `AdvisoryData` by taking
+    `resp` and `ecosystem` as input
+    """
+    edges = get_item(resp, "data", "securityVulnerabilities", "edges") or []
+    if not edges:
+        LOGGER.error(f"No edges found in response {resp!r}")
```
pombredanne, line 303: what does no edges mean? that there are no results? "edges" feels very specific as a error message and hard to remember in the flow of logging messages. Also what if the response is super long?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    for adv in edges:
+        affected_packages = []
+        cve_ids = set()
+        node = get_item(adv, "node")
+        if not node:
+            LOGGER.error(f"No node found in {adv!r}")
```
pombredanne, line 309: Node and edges feel super generic.... what could be a more explicit message that can make sense without knowing the internals of the GitHub GraphQL API?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+            continue
+        name = get_item(node, "package", "name")
+        if not name:
+            LOGGER.error(f"No name found in {node!r}")
+            continue
+        if process_name(ecosystem, name):
+            ns, pkg_name = process_name(ecosystem, name)
+            aff_range = get_item(node, "vulnerableVersionRange")
```
pombredanne, line 317: Can we get only a fix without an affected range? In this case why skipping? this sounds like a good enough data to me.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+            LOGGER.error(f"No advisory found in {node!r}")
+            continue
+        references = get_item(advisory, "references") or []
+        if references:
+            references: List[Reference] = list(extract_references(references))
+        vuln_desc = get_item(advisory, "summary")
+        identifiers = get_item(advisory, "identifiers") or []
+        for identifier in identifiers:
+            # collect CVEs
+            if identifier["type"] == "CVEjson_object":
```
pombredanne, line 347: I think you fat fingered this:
```suggestion
            if identifier["type"] == "CVE":
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
-    @staticmethod
-    def categorize_versions(
-        package_type: str, version_range: str, all_versions: Set[str]
-    ) -> Tuple[List[str], List[str]]:
-        version_class = version_class_by_package_type[package_type]
-        version_scheme = version_class.scheme
-        version_range = VersionSpecifier.from_scheme_version_spec_string(
-            version_scheme, version_range
+                        ]
+                        # Each Node has only one GHSA, hence exit after attaching
```
pombredanne, line 360: Are we sure we will not skip data this way?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
-                        adv_list.append(
-                            Advisory(
-                                vulnerability_id=cve_id,
-                                summary=vuln_desc,
-                                affected_packages=nearest_patched_package(
-                                    affected_purls, unaffected_purls
-                                ),
-                                references=references,
+            # attach the GHSA with severity score
+            if identifier["type"] == "GHSA":
```
pombredanne, line 351: IMHO use elif here:
```suggestion
            elif identifier["type"] == "GHSA":
```

then add an else: afterwards... if the id is neither GHSA nor CVE and we do not handle it yet it must be logged as an error.

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+                for ref in references:
+                    if ref.reference_id == identifier["value"]:
```
pombredanne, line 353: What if this is not equal?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    def get_inferences(self, advisory_data: AdvisoryData) -> Iterable[Inference]:
+        """
+        Generate and return Inferences for the given advisory data
+        """
+        try:
+            purl, affected_version_ranges, _ = AffectedPackage.merge(
+                advisory_data.affected_packages
+            )
+        except KeyError:
+            return iter([])
```
pombredanne, line 389: Is there a possibility that we have skipped data silently? Should this be logged? when can this happen?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        pkg_name = purl.name
+        pkg_namespace = purl.namespace
+        pkg_type = purl.type
+        ecosystem = get_ecosystem_from_type(pkg_type)
+        self.set_api(ecosystem, name)
+        name = process_name_namespace(ecosystem, pkg_name, pkg_namespace)
+        if hasattr(self.version_api, "module_name_by_package_name"):
+            pkg_name = self.version_api.module_name_by_package_name.get(name, pkg_name)
```
pombredanne, line 397: What if this differs from the purl.name ?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+        for affected_version_range in affected_version_ranges:
+            aff_vers, unaff_vers = categorize_versions(
+                ecosystem,
+                affected_version_range,
+                self.version_api.get(name, until=advisory_data.date_published).valid_versions,
+            )
+            affected_purls = [
+                PackageURL(name=pkg_name, namespace=pkg_namespace, version=version, type=pkg_type)
```
pombredanne, line 405: Why use pkg_namespace here and purl.namespace below?

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+    unaffected_versions = []
+    for version in all_versions:
+        if version in IGNORE_VERSIONS:
+            continue
+
+        if version in version_class.from_string(version_range.to_string()):
+            affected_versions.append(version)
+        else:
+            unaffected_versions.append(version)
+    return (affected_versions, unaffected_versions)
```
pombredanne, line 448: nit: no parens needed
```suggestion
    return affected_versions, unaffected_versions
```

vulnerabilities/importers/github.py
```diff
@@ -146,189 +122,339 @@
...
+
+
+def get_version_api(ecosystem: str) -> None:
```
pombredanne, line 451: Why not a simple constant as in  `VERSIONAPI_CLASS_BY_GITHUB_ECOSYSTEM`? The function feels like too much indirection here.

pombredanne (changes requested):
Thanks! I have a few nits for your consideration

vulnerabilities/importers/github.py
TG1999: Adding log for this

vulnerabilities/importers/github.py
TG1999: Yes!

vulnerabilities/helpers.py
```diff
@@ -167,11 +173,11 @@ def __lt__(self, other):
         patched_package_index = bisect.bisect_right(resolved_packages, vulnerable_package)
         patched_package = None
         if patched_package_index < resolved_package_count:
-            patched_package = resolved_packages[patched_package_index].package
+            patched_package = resolved_packages[patched_package_index].purl
 
         affected_package_with_patched_package_objects.append(
             AffectedPackage(
-                vulnerable_package=vulnerable_package.package, patched_package=patched_package
+                vulnerable_package=vulnerable_package.purl, patched_package=patched_package
```
pombredanne, line 180:
```suggestion
                vulnerable_package=vulnerable_package.purl, patched_package=patched_package.purl
```

vulnerabilities/importers/github.py
```diff
@@ -20,315 +20,521 @@
...
+            aliases.add(value)
+            # attach the GHSA with severity score
+            if identifier_type == "GHSA":
+                # Each Node has only one GHSA, hence exit after attaching
+                # score to this GHSA
+                for ref in references:
+                    if ref.reference_id == value:
+                        severity = get_item(advisory, "severity")
+                        if not severity:
+                            LOGGER.error(f"No severity found in {advisory!r}")
```
pombredanne, line 395: This is not an error and IMHO we should not log this

vulnerabilities/importers/github.py
```diff
@@ -20,315 +20,521 @@
...
+            version_api: VersionAPI = VERSION_API_CLASSES_BY_PACKAGE_TYPE[package_url.type]
+            self.version_api_by_purl_type[package_url.type] = version_api()
+        api_object = self.version_api_by_purl_type[package_url.type]
+        api_object.load_api([api_name])
+        self.version_api_by_purl_type[package_url.type] = api_object
+        return api_object.get(package_name=api_name, until=until).valid_versions
+
+    def get_inferences(self, advisory_data: AdvisoryData) -> Iterable[Inference]:
+        """
+        Generate and return Inferences for the given advisory data
```
pombredanne, line 451:
```suggestion
        Yield Inferences for the given advisory data
```

vulnerabilities/importers/github.py
```diff
@@ -20,315 +20,521 @@
...
+                    package.vulnerable_package
+                )
+
+            for (
+                fixed_package,
+                affected_packages,
+            ) in unique_patched_packages_with_affected_packages.items():
+                yield Inference.from_advisory_data(
+                    advisory_data,
+                    confidence=90,
```
pombredanne, line 511: Add a comment to explain why we use confidence of 90

vulnerabilities/importers/nginx.py
```diff
@@ -199,7 +200,7 @@ def get_inferences(self, advisory_data: AdvisoryData) -> Iterable[Inference]:
             purl, affected_version_ranges, fixed_versions = AffectedPackage.merge(
                 advisory_data.affected_packages
             )
-        except KeyError:
+        except UnMergeablePackageError:
```
pombredanne, line 203: Let's log this too

vulnerabilities/package_managers_2.py
```diff
@@ -0,0 +1,377 @@
...
+from typing import List
+from typing import MutableMapping
+from typing import Optional
+from typing import Set
+
+import requests
+from dateutil import parser as dateparser
+from django.utils.dateparse import parse_datetime
+
+from vulnerabilities.package_managers import Version as LegacyVersion
```
pombredanne, line 16: Let's inline and copy here the code you need from package_managers instead.

vulnerabilities/package_managers_2.py
```diff
@@ -0,0 +1,377 @@
...
+        try:
+            for entry_group in resp["items"]:
+                for entry in entry_group["items"]:
+                    all_versions.add(
+                        LegacyVersion(
+                            value=entry["catalogEntry"]["version"],
+                            release_date=dateparser.parse(entry["catalogEntry"]["published"]),
+                        )
+                    )
+        # FIXME: json response for YamlDotNet.Signed triggers this exception.
```
pombredanne, line 226: Can you have a test for this?

vulnerabilities/package_managers_2.py
```diff
@@ -0,0 +1,377 @@
...
+        # Some packages with many versions give a response of a list of endpoints.
+        # In such cases rather, we should collect data from those endpoints.
+        except KeyError:
+            pass
```
pombredanne, line 230: Log an error

vulnerabilities/package_managers_2.py
```diff
@@ -0,0 +1,377 @@
...
+        return base_url.format(pkg_name)
+
+    @staticmethod
+    def extract_versions(resp: dict) -> Set[LegacyVersion]:
+        all_versions = set()
+        try:
+            for entry_group in resp["items"]:
+                for entry in entry_group["items"]:
+                    all_versions.add(
+                        LegacyVersion(
```
pombredanne, line 221: I reckon this is old code carried over, but since we are touching it, what about extracting the two duplicated calls below in a variable?
```suggestion
                        catalog_entry = entry["catalogEntry"]
                        LegacyVersion(
```

vulnerabilities/package_managers_2.py
```diff
@@ -0,0 +1,377 @@
...
+        https://golang.org/ref/mod#goproxy-protocol ).
+        This functions trims the trailing part(s) of a package URL
+        and returns the remaining the module name.
+        For example:
+        >>> module = "https://github.com/xx/a"
+        >>> assert GoproxyVersionAPI.trim_url_path("https://github.com/xx/a/b") == module
+        """
+        # some advisories contains this prefix in package name, e.g. https://github.com/advisories/GHSA-7h6j-2268-fhcm
+        if url_path.startswith("https://pkg.go.dev/"):
+            url_path = url_path.removeprefix("https://pkg.go.dev/")
```
pombredanne, line 264: This works only on Python 3.9 ... this is a bug at multiple levels:
1. we do not want to support exclusively the 3.9 syntax now, since our minimum Python version is 3.8
2. this is clear that this is not tested at all... we need a test for this
3. we may not have a CI test running on Python 3.8

There are some possible bugs:
```
>>> trim_url_path("https://pkg.go.dev/github.com/containous/traefik/v2")
'github.com/containous/traefik'
>>> trim_url_path("https://pkg.go.dev/github.com/containous")
'github.com'
>>> trim_url_path("https://pkg.go.dev/github.com")
```

I reckon this is a carry over from the old package_managers, but this whole thing is both undertested and may not make sense. We need tests :)

pombredanne (approved):
Here are some nits for your review... Please fix the ones you think are essential and create issues for the ones that can wait and merge at you convenience.

vulnerabilities/importers/github.py
```diff
@@ -20,315 +20,528 @@
...
 from typing import Tuple
 
 import requests
 from dateutil import parser as dateparser
+from django.db.models.query import QuerySet
 from packageurl import PackageURL
-from univers.version_specifier import VersionSpecifier
-from univers.versions import version_class_by_package_type
+from univers.version_range import VersionRange
+from univers.version_range import build_range_from_github_advisory_constraint
```
pombredanne, line 37: Should `build_range_from_github_advisory_constraint` be  a class factory method of https://github.com/nexB/univers/blob/bc8e33fd923c5ec1dbe16528d3e207015f0460e2/src/univers/version_range.py#L714 

I would move https://github.com/nexB/univers/blob/bc8e33fd923c5ec1dbe16528d3e207015f0460e2/src/univers/version_range.py#L902 and https://github.com/nexB/univers/blob/bc8e33fd923c5ec1dbe16528d3e207015f0460e2/src/univers/version_range.py#L912 to the GitHubVersionRange class.

pombredanne: Note that this can be done separately and is not in the way of merging this.

pombredanne (commented):
Here are a few nits for your consisderation

vulnerabilities/importers/github.py
```diff
@@ -20,315 +20,528 @@
...
+    Yield `reference` by iterating over `reference_data`
+    >>> list(extract_references([{'url': "https://github.com/advisories/GHSA-c9hw-wf7x-jp9j"}]))
+    [Reference(url="https://github.com/advisories/GHSA-c9hw-wf7x-jp9j"), reference_id = "GHSA-c9hw-wf7x-jp9j" ]
+    >>> list(extract_references([{'url': "https://github.com/advisories/c9hw-wf7x-jp9j"}]))
+    [Reference(url="https://github.com/advisories/c9hw-wf7x-jp9j")]
+    """
+    for ref in reference_data:
+        url = ref["url"]
+        if not isinstance(url, str):
+            continue
```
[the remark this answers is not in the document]
TG1999: We are using url.upper(), integer or dict don't have "upper" attribute, so just assuring that we are dealing with a string while calling "upper" attribute, will add a log for this

Issue #650, opened 2022-03-21
Values larger than 1/3 of a buffer page cannot be indexed.


TG1999:
https://github.com/nexB/vulnerablecode/pull/632#discussion_r823700889

TG1999:
A possible solution that I got https://stackoverflow.com/questions/41971217/values-larger-than-1-3-of-a-buffer-page-cannot-be-indexed-django-error, although this is Postgres specific not Django specific

pombredanne:
#653 has been merged. Thanks!

PR #653, opened 2022-03-25
Work around PosgreSQL index issue

TG1999:
Update model to use an index on a checksum of the large summary text field.
Otherwise there can be times when this exceeds PostgreSQL limits.
See https://github.com/nexB/vulnerablecode/pull/632#discussion_r823700889
The failing test before this patch https://github.com/nexB/vulnerablecode/pull/659
Reference:  https://github.com/nexB/vulnerablecode/issues/650 


pombredanne (changes requested):
This should work, but IMHO we need first a minimal test case that reproduces the error.

pombredanne (commented):
Looking great!
You may be missing one migration though and this should be a data migration that computes the md5 for the actual data?

TG1999:
When I run `./manage.py makemigrations` command again, it does nothing, md5 hash computation part is not getting stored as a migration file.

ziadhany:
Why do we have three fields in the model (summary_md5, package affected_md5, references_md5) separately. IMHO we can have one field and hash all the fields inside it. This will help us not to store useless data in the database
instead using unique_together , use unique=True .
ex: 
```
md5_field = models.CharField(
                    max_length = 32,  
                    unique = True
                    )

def save(self, *args, **kwargs):
        self.md5_field = hashlib.md5(self.summary.encode("utf-8") + self.affected_packages.encode("utf-8") + ... 
        super(Advisory, self).save(*args, **kwargs)
```

vulnerabilities/models.py
```diff
@@ -326,6 +327,26 @@ class Advisory(models.Model):
...
+        # therefore hash( hash(a) + hash(b) ) is better than hash(a + b)
+        for json_field in json_fields:
+            if json_field:
+                hash = hashlib.md5(
+                    (hash + hashlib.md5(json.dumps(json_field).encode("utf-8")).hexdigest()).encode(
+                        "utf-8"
+                    )
+                ).hexdigest()
+        self.unique_content_id = hash
+        super(Advisory, self).save(*args, **kwargs)
```
pombredanne, line 347: This is likely enough?
```suggestion
        super().save(*args, **kwargs)
```

vulnerabilities/migrations/0005_auto_20220328_1929.py
```diff
@@ -0,0 +1,37 @@
...
+import json
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    def md5hash(apps, schema_editor):
+        Advisory = apps.get_model("vulnerabilities", "Advisory")
+        for advisory in Advisory.objects.all():
+            hash = ""
```
pombredanne, line 13: Since you can also ``json.dumps()`` a plain string or even a null, what about this simpler loop?
```
checksum = hashlib.md5()
for field in (advisory.summary, advisory.affected_packages, advisory.references):
    value = json.dumps(field, separators=(",", ":")).encode("utf-8")
    checksum.update(value)
advisory.unique_content_id = checksum.hexdigest()
advisory.save()
```

You likely want to use the same below too.

vulnerabilities/models.py
```diff
@@ -326,6 +327,26 @@ class Advisory(models.Model):
...
+
+    unique_content_id = models.CharField(max_length=32, blank=True)
```
pombredanne, line 349: Since this is computed, should a null=True be allowed too?

vulnerabilities/tests/test_postgres_workaround.py
```diff
@@ -0,0 +1,432 @@
...
+from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.models import Advisory
+from vulnerabilities.severity_systems import ScoringSystem
+
+data = AdvisoryData(
+    aliases=["CVE-2020-8908", "GHSA-5mg8-w23w-74h3"],
+    summary="lore ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua ut enim ad minim veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur excepteur sint occaecat cupidatat non proident sunt in culpa qui officia deserunt mollit anim id est laborum",
```
pombredanne, line 20: Please either wrap this long line ...
```suggestion
    summary="""lore ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et
             dolore magna aliqua ut enim ad minim veniam quis nostrud exercitation ullamco laboris nisi ut aliquip ex 
             ea commodo consequat duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu 
             fugiat nulla pariatur excepteur sint occaecat cupidatat non proident sunt in culpa qui officia deserunt 
             mollit anim id est laborum""",
```

or make it very long this way:

```suggestion
    summary="lore ipsum dolor sit amet consectetur adipiscing elit sed do" * 10000,
```


the point is that we want something long enough that used to fail before... Did you check these value are failing without the patch?

pombredanne (changes requested):
Thank you!
See a few comments for you consideration!

vulnerabilities/tests/test_postgres_workaround.py
TG1999: Added the failing test PR(before this patch)

PR #662, opened 2022-03-31
Add gitlab importer 

TG1999:


pombredanne (changes requested):
This looks good overall... we are just missing tests?

TG1999:
@pombredanne some work related to Nuget VersionRange is pending at univers side, once that is done I will add tests and the improver for this importer as well.

vulnerabilities/helpers.py
```diff
@@ -267,3 +267,81 @@ def _get_gh_response(gh_token, graphql_query):
...
+
+
+def get_reference_id(url: str):
+    """
+    Return the reference id from a URL
+    For example:
+    >>> get_reference_id("https://github.com/advisories/GHSA-c9hw-wf7x-jp9j")
+    'GHSA-c9hw-wf7x-jp9j'
+    """
+    url_parts = url.split("/")
```
pombredanne, line 279: Using a partition is always more robust IMHO
```suggestion
    _url, _, ref_id = url.strip("/").rpartition("/")
```

vulnerabilities/helpers.py
```diff
@@ -267,3 +267,81 @@ def _get_gh_response(gh_token, graphql_query):
...
+    last_url_part = url_parts[-1]
+    return last_url_part
+
+
+def extract_references(urls: List[str]):
+
+    # To avoid circular import
```
pombredanne, line 286: IMHO this means this is likely best as a factory class method of Reference....
may be more of less:
```Python
class Reference:
    @classmethod
    def from_url(cls, url):
        reference_id = get_reference_id(url)
        return cls(.....)
```

vulnerabilities/helpers.py
```diff
@@ -267,3 +267,81 @@ def _get_gh_response(gh_token, graphql_query):
...
+        return [], []
+    affected_versions = []
+    unaffected_versions = []
+    for package_version in package_versions or []:
+        if package_version in ignorable_versions:
+            continue
+        # Remove whitespace
+        package_version = package_version.replace(" ", "")
+        # Remove leading 'v'
+        package_version = package_version.lstrip("vV")
```
pombredanne, line 331: Should not these two removal be in some `version_class.from_string()` method of sorts?

vulnerabilities/importers/github.py
```diff
@@ -204,36 +199,12 @@ def advisory_data(self) -> Iterable[AdvisoryData]:
...
-        url = ref["url"]
-        if not isinstance(url, str):
-            logger.error(f"extract_references: url is not of type `str`: {url}")
-            continue
-        if "GHSA-" in url.upper():
-            reference = Reference(url=url, reference_id=get_reference_id(url))
-        else:
-            reference = Reference(url=url)
-        yield reference
+    for ref in references:
```
pombredanne, line 206: Doe this need a function? `(ref["url"] for ref in references)` does the same and may be best inlined?

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,329 @@
...
+    gitlab_url = "git+https://gitlab.com/gitlab-org/advisories-community/"
+
+    def advisory_data(self) -> Iterable[AdvisoryData]:
+        """
+        Return a list of AdvisoryData objects
+        """
+        try:
+            fork_directory = fork_and_get_dir(self.gitlab_url)
+        except Exception as e:
+            logger.error(f"Can't fetch url {self.gitlab_url}")
```
pombredanne, line 95: It is a fetch or a clone?

pombredanne (changes requested):
Thanks! Some partial and early feedback for your review.

vulnerabilities/helpers.py
TG1999: This is a shared helper function, so in case if a version_class does not strip 'v', this will help us there

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,329 @@
...
+    for fixed_version in fixed_versions:
+        yield AffectedPackage(
+            package=purl,
+            fixed_version=fixed_version,
+            affected_version_range=affected_version_range,
+        )
+
+
+def parse_yaml_file(file):
+    with open(file, "r") as f:
```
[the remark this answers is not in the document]
TG1999: We are already leveraging the subdirectory in `get_purl` function where we use package slug, package slug essentially gives detail about the package name and package type. 
https://gitlab.com/gitlab-org/advisories-community/-/blob/main/ci/schema/schema.json#L56

And the file name is stored in the identifiers array with other vuln IDs

vulnerabilities/helpers.py
```diff
@@ -292,3 +292,54 @@ def get_affected_packages_by_patched_package(
...
     return affected_packages_by_patched_package
+
+
+def get_reference_id(url: str):
+    """
+    Return the reference id from a URL
+    For example:
+    >>> get_reference_id("https://github.com/advisories/GHSA-c9hw-wf7x-jp9j")
+    'GHSA-c9hw-wf7x-jp9j'
+    """
```
pombredanne, line 303: Please add an issue to improve this to support and validate many URL styles correctly (including testing the prefix is correct like in GHSA, or CVE, etc.)

vulnerabilities/importers/github.py
```diff
@@ -451,6 +385,7 @@ def get_inferences(self, advisory_data: AdvisoryData) -> Iterable[Inference]:
             aff_vers, unaff_vers = resolve_version_range(
                 affected_version_range=affected_version_range,
                 package_versions=valid_versions,
+                ignorable_versions=WEIRD_IGNORABLE_VERSIONS,
```
pombredanne, line 388: The more I see these ignorable_versions and in particular the list of "WEIRD_IGNORABLE_VERSIONS,"  the more I think this is smelling bad or weird. IMHO we should rather have a new  approach outlined in https://github.com/nexB/vulnerablecode/issues/737

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    purl: PackageURL = get_purl(gitlab_advisory.get("package_slug"))
+    if not purl:
+        logger.error(f"parse_yaml_file: purl is not valid: {file!r}")
+        return AdvisoryData(
+            aliases=aliases,
+            summary=summary,
+            references=references,
+            date_published=date_published,
+        )
+    vrc: VersionRange = RANGE_CLASS_BY_SCHEMES[purl.type]
```
pombredanne, line 205:
```suggestion
    if affected_range: 
        vrc: VersionRange = RANGE_CLASS_BY_SCHEMES[purl.type]
        ......
```

e.g., indent all the block

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    version_class = vrc.version_class
+    gitlab_native_schemes = ["pypi", "gem", "npm", "go", "packagist"]
+    gitlab_scheme = GITLAB_SCHEME_BY_PURL_TYPE[purl.type]
+    try:
+        if gitlab_scheme in gitlab_native_schemes:
+            if affected_range:
```
pombredanne, line 211: IMHO this check should in the outer block... do not even try to compute a affected_version_range if you do not have an affected_range

```suggestion
```

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+                affected_version_range = from_gitlab_native(
+                    gitlab_scheme=gitlab_scheme, string=affected_range
+                )
+        else:
+            affected_version_range = vrc.from_native(affected_range) if affected_range else None
```
pombredanne, line 216:
```suggestion
            affected_version_range = vrc.from_native(affected_range)
```

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    except Exception as e:
+        logger.error(
+            f"parse_yaml_file: affected_range is not parsable: {affected_range!r} type:{purl.type} {e} {traceback.format_exc()}"
```
pombredanne, line 219:
```suggestion
            f"parse_yaml_file: affected_range is not parsable: {affected_range!r} type:{purl.type!r} error: {e!r}\n {traceback.format_exc()}"
```

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+
+    parsed_fixed_versions = []
+    for fixed_version in fixed_versions or []:
+        try:
+            fixed_version = version_class(fixed_version)
+            parsed_fixed_versions.append(fixed_version)
+        except Exception as e:
+            logger.error(f"parse_yaml_file: fixed_version is not parsable`: {fixed_version!r}")
```
pombredanne, line 228: Add the stacktrace to the log... otherwise this is not easy to diagnose.
```suggestion
            logger.error(f"parse_yaml_file: fixed_version is not parsable`: {fixed_version!r} error: {e!r}\n {traceback.format_exc()}")
```

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+
+    if parsed_fixed_versions:
+        affected_packages = list(
+            extract_affected_packages(affected_version_range, parsed_fixed_versions, purl)
```
pombredanne, line 232: Prefer named over positional arguments when possible

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    return AdvisoryData(
+        aliases=aliases,
+        summary=summary,
+        references=references,
+        date_published=date_published,
+        affected_packages=affected_packages,
+    )
+
+
+class GitLabBasicImprover(Improver):
```
pombredanne, line 253: Please add a docstring explaining why this is needed and what this does at a high level.

vulnerabilities/package_managers.py
```diff
@@ -678,3 +682,41 @@ def fetch(self, pkg: str) -> Iterable[PackageVersion]:
...
+    DebianVersionAPI,
+    GitHubTagsAPI,
+}
+
+VERSION_API_CLASSES_BY_PACKAGE_TYPE = {cls.package_type: cls for cls in VERSION_API_CLASSES}
+
+
+def get_api_package_name(purl: PackageURL) -> str:
+    """
+    Return the package name expected by the GitHub API given a PackageURL
```
pombredanne, line 707: Is this GitHub API specific? or for Gitlab?

pombredanne (changes requested):
Thanks!
I have a few review comments for your consideration.

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    Return a PackageURL object from a package slug
+    """
+    parts = package_slug.split("/")
+    parts = list(filter(not_empty, parts))
+    gitlab_scheme = parts[0]
+    purl_type = PURL_TYPE_BY_GITLAB_SCHEME[gitlab_scheme]
+    if gitlab_scheme == "go":
+        name = "/".join(parts[1:])
+        return PackageURL(type=purl_type, namespace=None, name=name)
```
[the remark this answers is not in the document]
TG1999: I am following the same convention that is used in github importer to store package name in go package type (We don't store namespace in github and only depends on name to fetch data from API)

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    return value is not None and value != ""
+
+
+def get_purl(package_slug):
+    """
+    Return a PackageURL object from a package slug
+    """
+    parts = package_slug.split("/")
+    parts = list(filter(not_empty, parts))
+    gitlab_scheme = parts[0]
```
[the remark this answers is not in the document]
TG1999: This may get wrong in go type packages

vulnerabilities/importers/gitlab.py
```diff
@@ -0,0 +1,347 @@
...
+    if gitlab_scheme == "go":
+        name = "/".join(parts[1:])
+        return PackageURL(type=purl_type, namespace=None, name=name)
+    # if package slug is of the form:
+    # "nuget/NuGet.Core"
+    if len(parts) == 2:
+        name = parts[1]
+        return PackageURL(type=purl_type, name=name)
+    # if package slug is of the form:
+    # "nuget/github/user/abc/NuGet.Core"
```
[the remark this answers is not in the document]
TG1999: `nuget/github.com/beego/beego/v2/nuget`  https://gitlab.com/gitlab-org/advisories-community/-/blob/main/nuget/github.com/beego/beego/v2/nuget/CVE-2021-39391.yml

vulnerabilities/package_managers.py
TG1999: This is not specific to any importer, It is shared between both

vulnerabilities/importer.py
```diff
@@ -105,6 +107,15 @@ def from_dict(cls, ref: dict):
             ],
         )
 
+    @classmethod
+    def from_url(cls, url):
+        reference_id = get_reference_id(url)
```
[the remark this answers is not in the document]
TG1999: #738

vulnerabilities/importers/gitlab.py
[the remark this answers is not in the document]
pombredanne: As an example I think that this slug `go/gitlab.com/gitlab-org/gitlab-shell` would become this purl: `pkg:golang/gitlab.com/gitlab-org/gitlab-shell` and if you parse it you get this:
```Python
>>> PackageURL.from_string("pkg:golang/gitlab.com/gitlab-org/gitlab-shell")
PackageURL(type='golang', namespace='gitlab.com/gitlab-org', name='gitlab-shell', version=None, qualifiers={}, subpath=None)
```
But this is incorrect:
```Python
>>> PackageURL(type='golang', namespace=None, name='gitlab.com/gitlab-org/gitlab-shell')
PackageURL(type='golang', namespace=None, name='gitlab.com/gitlab-org/gitlab-shell', version=None, qualifiers={}, subpath=None)
```
.... even if this may look correct on the surface:
```Python
>>> PackageURL(type='golang', namespace=None, name='gitlab.com/gitlab-org/gitlab-shell').to_string()
'pkg:golang/gitlab.com/gitlab-org/gitlab-shell'
```

IMHO this would be much safer and clean
```Python
>>> slug = "go/gitlab.com/gitlab-org/gitlab-shell"
>>> scheme, _, ns_name = slug.partition("/")
>>> assert scheme == "go"
>>> purl = f"pkg:golang/{ns_name}"
>>> purl
'pkg:golang/gitlab.com/gitlab-org/gitlab-shell'
>>> PackageURL.from_string(purl)
PackageURL(type='golang', namespace='gitlab.com/gitlab-org', name='gitlab-shell', version=None, qualifiers={}, subpath=None)

```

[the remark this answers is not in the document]
TG1999: refer #742 for followup on this

vulnerabilities/importers/gitlab.py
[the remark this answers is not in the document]
TG1999: refer https://github.com/nexB/vulnerablecode/issues/742 for followup on this

vulnerabilities/importers/gitlab.py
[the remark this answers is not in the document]
TG1999: This doesn't exist on nuget https://www.nuget.org/packages?q=beego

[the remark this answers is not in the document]
TG1999: https://github.com/nexB/vulnerablecode/issues/743

vulnerabilities/package_managers.py
```diff
@@ -678,3 +682,41 @@ def fetch(self, pkg: str) -> Iterable[PackageVersion]:
...
+    'foo/bar'
+    """
+    if purl.type == "maven":
+        return f"{purl.namespace}:{purl.name}"
+
+    if purl.type == "composer":
+        return f"{purl.namespace}/{purl.name}"
+
+    if purl.type in ("nuget", "pypi", "gem", "golang", "npm"):
+        return purl.name
```
pombredanne, line 720: This will unlikely be correct for goland and npm as both can have namespaces.

pombredanne (commented):
Just a tiny nit left wrt. purl for npm and golang.
You should consider the namespace there IMHO when doing API lookups. Both are using `/` as a separator. Then please go ahead and merge!
:bow:

pombredanne (approved):

vulnerabilities/package_managers.py
TG1999: @pombredanne good catch, adding to your point, some packages may not have a namespace so in that case we can directly return name for example this https://registry.npmjs.org/hapi and those who may have namespace can be clubbed using '/' like this https://registry.npmjs.org/@curveball/a12n-server

TG1999:
While I was at it I also solved a minor wart in github importer in this commit https://github.com/nexB/vulnerablecode/pull/662/commits/b7ae8dcd2d129e748fc26dd579ee33a919046a6d, solving #673

PR #667, opened 2022-04-06
Lookup Vulnerabilities by CPE

TG1999:


vulnerabilities/helpers.py
```diff
@@ -38,6 +38,8 @@
 from packageurl import PackageURL
 from univers.version_range import RANGE_CLASS_BY_SCHEMES
 
+import vulnerabilities.models as models
```
pombredanne, line 41:
```suggestion
from vulnerabilities import models
```

pombredanne (changes requested):
Can you add a `Reference: https... ` URL to the issue this about?
Also is this really a helper? or would this be rather a method on advisories where a CPE get you vulnerabilities back first? and a second method could build on this and get CVEs from a vulnerabilities query set?

pombredanne (approved):
LGTM!

PR #671, opened 2022-04-07
Add tests for checking the API

TG1999:


pombredanne (approved):
Thanks!
I am merging this for now. We will need more tests though including one that would check the whole API response format for each endpoint.

PR #672, opened 2022-04-07
Add example importer and improver

Hritik14:
This is a dummy importer and improver that runs without network
connection only to test vulnerablecode framework


vulnerabilities/importers/__init__.py
```diff
@@ -20,9 +20,17 @@
...
 from vulnerabilities.importers import nginx
+from vulnerabilities.importers import npm
 
-IMPORTERS_REGISTRY = [nginx.NginxImporter, alpine_linux.AlpineImporter, github.GitHubAPIImporter]
+IMPORTERS_REGISTRY = [
+    example.ExampleImporter,
+    github.GitHubAPIImporter,
+    nginx.NginxImporter,
+    alpine_linux.AlpineImporter,
+    npm.NpmImporter,
```
TG1999, line 33: NPM importer is not migrated yet, so I think we should not have it here

TG1999:
@Hritik14 Should we have some tests here for testing the example importer and improver ?

pombredanne (changes requested):
IMHO this should not be registered as this creates junk data.

vulnerabilities/importers/__init__.py
Hritik14: oops! My mistake

Hritik14:
@TG1999 Good idea! Though, instead of testing this example importer / improver, I suppose I can test the entire framework _using_ the example.

@pombredanne 
Makes sense. The only way to run this example will now be using tests.

vulnerabilities/importers/example.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime
```
pombredanne, line 1: It makes sense to have it as a test... but since this is a test, why not move `
vulnerabilities/importers/example.py` to the tests directory instead? This would make more sense.
Especially I would like to move the tests out of the main code directories at some point in time in the future

Hritik14: The initial idea behind putting it in ``vulnerabilities/importers/`` was to offer a starting point to contributors about how a real importer/improver might look like but that is covered in detail in the documentation as well so we might as well move these to ``tests``

pombredanne (approved):
LGTM!
Thanks

Issue #468, opened 2021-06-02
Create and deploy stable/permanent host for VulnerableCode


pombredanne:
We need a host and home

pombredanne:
See https://public.vulnerablecode.io/ the puppy is free now !

PR #677, opened 2022-04-08
Enable deployment

pombredanne:
This PR makes multiple adjustments to enable deployment:

- update settings, Makefile and configuration
- clean, streamline and align Docker configuration
- improve README and documentation

A lot of this was done to align and borrow ScanCode.io config

Reference: https://github.com/nexB/vulnerablecode/issues/468


docs/source/installation.rst
```diff
@@ -39,9 +41,7 @@ VulnerableCode is distributed with ``Dockerfile`` and ``docker-compose.yml`` fil
 required for the creation of the Docker image.
 
 Clone the git `VulnerableCode repo <https://github.com/nexB/vulnerablecode>`_,
-create an environment file, and build the Docker image:
-
-.. code-block:: bash
+create an environment file, and build the Docker image::
```
Hritik14, line 45: Is there a reason to prefer ``::`` over ``code-block`` ?

vulnerablecode/settings.py
```diff
@@ -7,95 +7,102 @@
 
 # Environment
 
-env_file = str(ROOT_DIR.joinpath(".env"))
-if Path(env_file).exists():
-    environ.Env.read_env(env_file)
+ENV_FILE = "/etc/vulnerablecode/.env"
+if not Path(ENV_FILE).exists():
+    ENV_FILE = ROOT_DIR / ".env"
```
Hritik14, line 12: Shouldn't the local config be preferred over the one in ``/etc`` ?

Imagine a user with a bad ``/etc`` config deleting and recloning vulnerablecode a thousand times and expecting ``make env`` to fix their issue but it won't.

Hritik14 (approved):
This is really neat! Here are a few suggestions from my end.

Dockerfile
```diff
@@ -1,11 +1,24 @@
...
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# VulnerableCode is a trademark of nexB Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# See https://github.com/nexB/vulnerablecode for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects
 
-# Force unbuffered stdout and stderr (i.e. they are flushed to terminal immediately)
+FROM python:3.9
```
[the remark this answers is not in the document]
pombredanne: @Hritik14 I am not sure this is what you think this is :D
See this proposed global approach https://github.com/nexB/aboutcode/pull/111/files#diff-7c33a51b412e7a778049cdc4990b84bec18b0e1ed48fa5b4c65fb2f5d7fe0ba6R43 based on this suggested new way: https://github.com/nexB/aboutcode/issues/110

Makefile
```diff
@@ -103,26 +109,31 @@ sqlite:
 	@$(MAKE) migrate
 
 run:
-	${ACTIVATE} ./manage.py runserver
+	${MANAGE} runserver 8001 --noreload --insecure
```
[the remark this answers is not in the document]
pombredanne: I just want this aligned with scancode.io. 8000 is the port used by the docker gunicorn. So it could not be used by a plain run. `--noreload` avoids several quirk when doing auto-reloads

docs/source/installation.rst
pombredanne: This makes this more readable when looking at the plain text ... it has the same effect. Less is more maybe?

vulnerablecode/settings.py
pombredanne: > Shouldn't the local config be preferred over the one in /etc ?

not when doing a deployment. So in general the global is what matters on a server. The local is only for development use.

> Imagine a user with a bad /etc config deleting and recloning vulnerablecode a thousand times and expecting make env to fix their issue but it won't.

That's a good point... I should add some doc there. I also want to keep this in tight sync with scancode.io and avoid variations between the two unless there is a reason.  We may want to evolve a skeleton for this to keep them in sync

CHANGELOG.rst
```diff
@@ -1,14 +1,21 @@
...
+Version v30.0.0
+----------------
 
 - Add support for nix
+
 - Improve documentation
 
+- Refactor core processing with Importers that import data and Improvers that
+  get the imported data and convert that in Vulnerabilities and Packages and can
+  also improve abd refine imported and existing data. The migration to this new
```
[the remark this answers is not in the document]
pombredanne: This has been fixed in the latest push

vulnerablecode/settings.py
pombredanne: I pushed some doc at https://github.com/nexB/vulnerablecode/pull/677/commits/bf4864bb88e54d13ee88e8e4322ec82eecbeedd3

Dockerfile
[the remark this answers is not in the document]
Hritik14: Ouch !

PR #687, opened 2022-04-09
Improve test run

pombredanne:
This is to streamline CI runs:
- run import tests once a day on schedule
- run tests on all supported Pythons


pombredanne (commented):
I am merging this now as otherwise the import tests failure are a PITA

PR #699, opened 2022-04-13
Remove null from string based fields

TG1999:
Remove null from string based fields


pombredanne (changes requested):
LGTM, but are we missing a migration may be?

pombredanne (approved):
LGTM: please just add the missing migration and merge.

PR #702, opened 2022-04-19
Amend to_dict function for inferences

TG1999:
Also return an empty list instead of returning None when there is no advisory data passed in get_inferences


pombredanne (approved):
LGTM!

PR #719, opened 2022-05-03
Migrate redhat importer

TG1999:

### DIFF
--- a/vulnerabilities/importers/redhat.py
+++ b/vulnerabilities/importers/redhat.py
@@ -20,81 +20,99 @@
 #  VulnerableCode is a free software code from nexB Inc. and others.
 #  Visit https://github.com/nexB/vulnerablecode/ for support and download.
 
+import logging
+from typing import Dict
+from typing import Iterable
+from typing import List
+
 import requests
 from packageurl import PackageURL
+from univers.version_range import RpmVersionRange
 
 from vulnerabilities import severity_systems
-from vulnerabilities.helpers import nearest_patched_package
+from vulnerabilities.helpers import get_item
 from vulnerabilities.helpers import requests_with_5xx_retry
 from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
 from vulnerabilities.importer import Importer
 from vulnerabilities.importer import Reference
 from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.rpm_utils import rpm_to_purl
 
-
-class RedhatImporter(Importer):
-    def __enter__(self):
-
-        self.redhat_cves = fetch()
-
-    def updated_advisories(self):
-        processed_advisories = list(map(to_advisory, self.redhat_cves))
-        return self.batch_advisories(processed_advisories)
-
+logger = logging.getLogger(__name__)
 
 requests_session = requests_with_5xx_retry(max_retries=5, backoff_factor=1)
 
 
-def fetch():
-    """
-    Return a list of CVE data mappings fetched from the RedHat API.
-    See:
-        https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/index
-    """
-    cves = []
+def fetch_list_of_cves() -> Iterable[List[Dict]]:
     page_no = 1
-    url_template = "https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={}"  # nopep8
-
     cve_data = None
     while True:
-        current_url = url_template.format(page_no)
+        current_url = f"https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={page_no}"  # nopep8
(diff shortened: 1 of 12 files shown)


vulnerabilities/importers/redhat.py
```diff
@@ -20,73 +20,83 @@
...
         if not cve_data:
             break
-        cves.extend(cve_data)
         page_no += 1
+        yield cve_data
+
+
+class RedhatImporter(Importer):
+
+    spdx_license_expression = "CC-BY-4.0"
```
pombredanne, line 71: Is there a license URL too?

vulnerabilities/importers/redhat.py
```diff
@@ -20,73 +20,83 @@
...
-    affected_purls = []
-    if advisory_data.get("affected_packages"):
-        for rpm in advisory_data["affected_packages"]:
-            purl = rpm_to_purl(rpm)
-            if purl:
-                affected_purls.append(purl)
+    affected_packages: List[AffectedPackage] = []
+    for rpm in advisory_data.get("affected_packages") or []:
+        purl = rpm_to_purl(rpm)
+        if purl:
```
pombredanne, line 82: What if there is no purl? are we skipping data silently then?

vulnerabilities/importers/redhat.py
```diff
@@ -166,25 +179,98 @@ def to_advisory(advisory_data):
 
     references.append(Reference(severities=redhat_scores, url=advisory_data["resource_url"]))
     return AdvisoryData(
-        vulnerability_id=advisory_data["CVE"],
-        summary=advisory_data["bugzilla_description"],
-        affected_packages=nearest_patched_package(affected_purls, []),
+        aliases=advisory_data.get("CVE") or "",
```
pombredanne, line 182: I think aliases is a list not a string for not a `or ""``

vulnerabilities/importers/redhat.py
```diff
@@ -166,25 +179,98 @@ def to_advisory(advisory_data):
...
+    n, v, r, a = m.groups()
+    if ":" not in v:
+        return None, n, v, r, a
+    e, v = v.split(":", 1)
+    if e.isdigit():
+        e = int(e)
+    return (e, n, v, r, a)
+
+
 def rpm_to_purl(rpm_string):
```
pombredanne, line 247: IMHO move this to an rpm_utils.py module too, and accept a new argument which would be the "namespace" (here redhat)

pombredanne (changes requested):
Thanks! I posted a few review comments,
The key thing is to make sure we do not skip over data silently without logging.

vulnerabilities/importers/redhat.py
TG1999: We are not skipping the data silently, whenever we return None from rpm_to_purl function we log the reason why we can't return a purl, and since there could by many cases for returning a Purl none, we can't infer it at the calling side

vulnerabilities/importers/redhat.py
```diff
@@ -20,73 +20,84 @@
...
     page_no = 1
-    url_template = "https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={}"  # nopep8
-
     cve_data = None
     while True:
-        current_url = url_template.format(page_no)
+        current_url = f"https://access.redhat.com/hydra/rest/securitydata/cve.json?per_page=10000&page={page_no}"  # nopep8
         try:
-            print(f"Fetching: {current_url}")
             response = requests_session.get(current_url)
```
pombredanne, line 54: IMHO the whole fetching should be abstracted in a function that hides requests/sessions/statsus code and processes all errors. This can be in another PR though

vulnerabilities/importers/redhat.py
```diff
@@ -126,7 +141,7 @@ def to_advisory(advisory_data):
             rhsa_aggregate_severities = []
             if rhsa_data.get("cvrfdoc"):
                 # not all RHSA errata have a corresponding CVRF document
-                value = rhsa_data["cvrfdoc"]["aggregate_severity"]
+                value = get_item(rhsa_data, "cvrfdoc", "aggregate_severity")
```
pombredanne, line 144: Let's handle the cases where there is no value
```suggestion
                value = get_item(rhsa_data, "cvrfdoc", "aggregate_severity")
                if value:
```

vulnerabilities/importers/redhat.py
```diff
@@ -126,7 +141,7 @@ def to_advisory(advisory_data):
...
                 rhsa_aggregate_severities.append(
```
pombredanne, line 145:
```suggestion
                    rhsa_aggregate_severities.append(
```

pombredanne (approved):
LGTM... Just a few minor comments

TG1999:
Thanks @pombredanne , while I was at it, I realised severities were not displayed on UI, also fixed that. Merging it as soon it's all green!

PR #740, opened 2022-05-19
Migrate debian-oval and ubuntu importer

TG1999:


pombredanne (changes requested):
Looking good... do you mind to rebase on the latest code?

pombredanne:
@TG1999 let's rebase one more time before merging!

vulnerabilities/importers/debian_oval.py
```diff
@@ -8,38 +8,29 @@
...
 
 import requests
 
 from vulnerabilities.importer import OvalImporter
-from vulnerabilities.package_managers import DebianVersionAPI
-from vulnerabilities.utils import create_etag
 
 
 class DebianOvalImporter(OvalImporter):
+    spdx_license_expression = "LicenseRef-scancode-unknown"
```
pombredanne, line 19: This should be:
```suggestion
    spdx_license_expression = "LicenseRef-scancode-other-permissive"
```
and we should have the exact the same value for https://github.com/nexB/vulnerablecode/blob/ae609a857c747c45c8391b9b0fd3c54b0b30d921/vulnerabilities/importers/debian.py#L42 ...
The license URL and MIT does not apply here IMHO and we should not have a license URL

And the notice in debian should also exist here: https://github.com/nexB/vulnerablecode/blob/ae609a857c747c45c8391b9b0fd3c54b0b30d921/vulnerabilities/importers/debian.py#L44

vulnerabilities/importers/debian_oval.py
```diff
@@ -8,38 +8,29 @@
...
         super().__init__(*args, **kwargs)
         # we could avoid setting translations, and have it
         # set by default in the OvalParser, but we don't yet know
         # whether all OVAL providers use the same format
         self.translations = {"less than": "<"}
-        self.pkg_manager_api = DebianVersionAPI()
 
     def _fetch(self):
-        releases = self.config.releases
+        releases = ["wheezy", "stretch", "jessie", "buster", "bullseye"]
```
pombredanne, line 29: Please enter an issue to find a way to get alerted when there are new and retied Debian releases

vulnerabilities/utils.py
```diff
@@ -318,7 +318,7 @@ def get_reference_id(url: str):
 def resolve_version_range(
     affected_version_range: VersionRange,
     package_versions: List[str],
-    ignorable_versions: List[str],
+    ignorable_versions: List[str] = [],
```
pombredanne, line 321: This is a global... make it a tuple to avoid issues.

pombredanne (approved):
LGTM except for a small nit... Please merge!

vulnerabilities/importers/debian_oval.py
TG1999: https://github.com/nexB/vulnerablecode/issues/967

Issue #713, opened 2022-04-27
Calculate CVSS score from CVSS vector


Hritik14:
We should have a way to calculate CVSS score from the given CVSS vector in our ``ScoringSystem``. The calculated score should also be saved in the database. For now we can have the CVSS base scores.

See: https://www.first.org/cvss/calculator/3.0

_(via: )_

DennisClark:
this post from NIST might be interesting to those following this issue: 
https://csrc.nist.gov/publications/detail/nistir/8409/final

PR #747, opened 2022-05-21
add support for calculating CVSS score from the CVSS vector

ziadhany:
Reference: #713


pombredanne (changes requested):
Looking good... Can you add some tests?

pombredanne (changes requested):
LGTM! :+1: 
@ziadhany do you mind to rebase on the latest code and also regen the migration on the latest code?
Also there is a need for a data migrations to migrate the current CVSS vector to this new approach

pombredanne:
re: 
>  Also there is a need for a data migrations to migrate the current CVSS vector to this new approach

CVSSV31,
        CVSSV31_VECTOR,

say we have two rows with these data:
- scoring_system: CVSSV31_VECTOR value:"CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:H"
- scoring_system: CVSSV31 value: "8.6"

After the migration(s) I would like to see only one record:
- scoring_system: CVSSV31 value: "8.6" scoring_elements: "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:H"

This is best done in three steps:
1. model schema update adding scoring_elements
2. data migration that searches and adds or computes the score for CVSS severities and add these to one of the two records
2.1 one way: 
starting from this record - scoring_system: CVSSV31_VECTOR value:"CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:H"
update the record this - update scoring_system: CVSSV31, copy value to value to scoring_elements, compute value
2.2 other way: group pairs of records and merge the two in one of the two records
3. data migration to remove the extra record

vulnerabilities/models.py
```diff
@@ -369,7 +367,27 @@ class VulnerabilitySeverity(models.Model):
...
+
+    scoring_elements = models.CharField(
+        max_length=100,
+        null=True,
+        help_text="Supporting a scoring elements as a string "
+        "For example a CVSS vector Important, High, Medium ,Low."
+        "Typically used to compute the value",
+    )
+
+    def save(self, *args, **kwargs):
```
pombredanne, line 383: IMHO we should not save here

pombredanne (changes requested):

vulnerabilities/models.py
TG1999: I think we should compute it in `severity_system.py`

vulnerabilities/migrations/0030_vulnerabilityseverity _rm_extra_records_swap.py
```diff
@@ -0,0 +1,51 @@
+from django.db import migrations
+
+
+def remove_extra_rows(apps, schema_editor):
+    Vuln_severity = apps.get_model('vulnerabilities', 'VulnerabilitySeverity')
+    for vuln_severity in Vuln_severity.objects.all():
```
TG1999, line 6: @ziadhany please check 0025 to see how we are removing duplicates

vulnerabilities/migrations/0030_vulnerabilityserverity_compute_score.py
```diff
@@ -0,0 +1,24 @@
+from django.db import migrations
+
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+
+
+def compute_score(apps, schema_editor):
+    Vuln_severity = apps.get_model('vulnerabilities', 'VulnerabilitySeverity')
+    for vuln_severity in Vuln_severity.objects.all():
```
pombredanne, line 8: Do not iterate all, instead you only care for the cvss so filter on them.

vulnerabilities/migrations/0030_vulnerabilityserverity_compute_score.py
```diff
@@ -0,0 +1,24 @@
...
+        if vuln_severity.value and not vuln_severity.scoring_elements:
+            try:
+                vuln_severity.scoring_elements = SCORING_SYSTEMS[vuln_severity.scoring_system].as_score(vuln_severity.value)
```
pombredanne, line 11: What if we compute a score that's not the same the one we had before?
We should find a way to do what I suggest as a possibility e.g, " 2.2 other way: group pairs of records and merge the two in one of the two records" and ensure that we do not end up computing a score that's not the score we have stored

pombredanne (changes requested):
Thanks for the update! please see some comments inline.

pombredanne:
@ziadhany I pushed these updates:
1. merge the latest main branch including handling overlapping migration numbers
2. update the the migration(s) to use a bulk update/bulk delete approach
3. propagate the use of scoring_elements everywhere including importers, tests, API, etc.
4. added some cosmetic improvements to regen tests fixtures easily

The new migration(s) handle all possible duplicates cases this way:

- keep a mapping of severities keyed by the "unique together" values of the model
- keep a set of severity ids to delete
- for each CVSS-related severity:
  - if this is a vector:
    - update this to move value to scoring elements and compute score in value and use plain cvss scoring system
  - if it does not exist in the mapping
    - add it to this mapping
  - else
    - if needed merge/add vector to the existing severity
    - mark this severity for deletion
- bulk delete the set of ids 
- bulk update severities mapping
 

I need to test run it on the whole DB to validate this is correct

pombredanne:
@ziadhany This is essentially the same approach as yours, but folded in a single migration with bulk updates

pombredanne (approved):

ziadhany:
LGTM ,  I tested it with my local database

TG1999:
@ziadhany please add CHANGEOLG for this PR

Issue #645, opened 2022-03-14
Collect more data from GitHub GraphQL API


TG1999:
Check GitHubGraphQL API for advisories here https://docs.github.com/en/enterprise-server@3.1/graphql/reference/objects#securityadvisory

PR #748, opened 2022-05-21
Add firstPatchedVersion in github API

TG1999:
The GitHub importer failed to parse Composer packages who did not have a slash in their names. We now no longer raise an exception if a Composer package does not have "vendor/name" but only a plain name.

Also we were not getting the first patched version from the GitHub API using GraphQL. This PR is also fixing this.

Reference: https://github.com/nexb/vulnerablecode/issues/645

### DIFF
--- a/vulnerabilities/importers/github.py
+++ b/vulnerabilities/importers/github.py
@@ -320,7 +320,7 @@ def process_response(resp: dict, package_type: str) -> Iterable[AdvisoryData]:
                 logger.error(f"Unknown identifier type {identifier_type!r} and value {value!r}")
 
         yield AdvisoryData(
-            aliases=dedupe(aliases),
+            aliases=sorted(dedupe(aliases)),
             summary=summary,
             references=references,
             affected_packages=affected_packages,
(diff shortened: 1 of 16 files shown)


vulnerabilities/importers/github.py
```diff
@@ -310,12 +319,8 @@ def process_response(resp: dict, package_type: str) -> Iterable[AdvisoryData]:
             else:
                 logger.error(f"Unknown identifier type {identifier_type!r} and value {value!r}")
 
-        date_published = get_item(advisory, "publishedAt")
-        if date_published:
-            date_published = dateparser.parse(date_published)
-
         yield AdvisoryData(
-            aliases=sorted(list(aliases)),
```
pombredanne, line 318: IMHO it is important to get sorting back in there.

pombredanne (approved):
LGTM... Please double check the get item() , may be to ensure you have corner cases testsed.

TG1999:
Thanks! added tests for the corner cases, merging now!

PR #765, opened 2022-06-15
Search from alias in UI

TG1999:


pombredanne (approved):
LGTM!
Do we have aliases in the API too?

Issue #651, opened 2022-03-24
Add support for CWE


pombredanne:
We decided early to keep the data set we track as minimal... but adding CWE would be a great addition. This can be a fairly involved issue since many importers may need to be updated.

ziadhany:
Can we use a library like this https://github.com/Julian-Nash/cwe to get a CWE by ID and add another JSON field in the model?

Hritik14:
CWE could be used for categorization of vulnerabilities. Later on, we might be able to find a mapping between ``OWSAP top 10 -> CWE``, this can be used to prioritize vulnerabilities for the downstream.

To implement CWE (categorization) support a system similar to currently implemented [ScoringSystem](https://github.com/nexB/vulnerablecode/blob/67dda33e1e1926a380abb96023d757acfe45db78/vulnerabilities/severity_systems.py#L32) could be used.

_(via: )_

ziadhany:
I think we have a compressed CSV file containing the fields of the desired Weaknesses related to CWE VIEW: Software Development.
https://cwe.mitre.org/data/definitions/699.html

https://cwe.mitre.org/data/csv/699.csv.zip

PR #782, opened 2022-06-24
Add support for CWE

ziadhany:
Reference: https://github.com/nexB/vulnerablecode/issues/651


ziadhany:
Screenshot from 2022-06-27 20-19-05

vulnerabilities/models.py
```diff
@@ -10,8 +10,9 @@
 import hashlib
 import json
 import logging
-import uuid
 
+from cwe import Database
+from django.contrib.postgres.fields import ArrayField
```
pombredanne, line 15: Do you mind to rather use a JSONField? this is more portable than the postgres-only ArrayField

pombredanne (changes requested):
This would need to be rebased... also you may want to remove the UI part to treat it separately as the UI has changed extensively!
Also could we also get the CWE from the NVD importer? or is this something for later?
Finally two other considerations:
1. the CWE library maintainer has not much replied to data, so if we need this we would have to fok it
2. is there something else beyond CWE that would be about a more general concept of categories?

vulnerabilities/models.py
ziadhany: No Problem, I will change it. but what is the best way to store CWE in the database as ( int or string ) 
{ "cwe_ids" : ['CWE-327', 'CWE-325' ]}
{ "cwe_ids" : [327 ,  325 ]}

ziadhany:
> This would need to be rebased... also you may want to remove the UI part to treat it separately as the UI has changed extensively! Also could we also get the CWE from the NVD importer? or is this something for later? 

ok, no problem I will change this, but what about editing all importers not just NVD, I think I could handle this in a separate pull request.
>     1. the CWE library maintainer has not much replied to data, so if we need this we would have to fok it

oops, yes we can fork it. it isn't a complicated library, all data come from this [ database](https://github.com/Julian-Nash/cwe/blob/master/cwe/database.py), we need to make sure the database is updated so I think we should use this link 
https://cwe.mitre.org/data/downloads.html as our database. 

>     2. is there something else beyond CWE that would be about a more general concept of categories? 
I don't know but the main three categories are  
- **Software Development**
- Hardware Design
- Research Concepts

some External Mappings : 

- **CWE Top 25 (2022)**
- **OWASP Top Ten (2021)**
- **Software Fault Pattern Clusters**

ziadhany:
Screenshot from 2022-09-11 01-38-04

ziadhany:
Screenshot 2022-11-01 at 13-40-57 VulnerableCode Vulnerability Details - VCID-113g-6mxr-aaag

pombredanne:
@ziadhany the cwe2 library is ready now: https://pypi.org/project/cwe2/

vulnerabilities/templates/vulnerability_details.html
```diff
@@ -174,6 +173,36 @@
...
                         </table>
                     </div>
+
+                    <div class="has-text-weight-bold tab-nested-div ml-1 mb-1 mt-6">
+                        Weaknesses ({{ weaknesses|length }})
+                    </div>
+                     <div class="tab-nested-div">
+                        <table class="table is-bordered is-striped is-narrow is-hoverable is-fullwidth gray-header-border">
+                            <tr>
+                                <th> CWE id  </th>
```
pombredanne, line 183: Use plain `CWE`

pombredanne (changes requested):
Here are a few nitpickings for your review

vulnerabilities/improve_runner.py
```diff
@@ -133,6 +134,10 @@ def process_inferences(inferences: List[Inference], advisory: Advisory, improver
                 fix=True,
             ).update_or_create()
 
+        if inference.weaknesses:
+            for cwe_id in inference.weaknesses:
+                Weakness.objects.update_or_create(cwe_id=cwe_id, vulnerabilities=vulnerability)
```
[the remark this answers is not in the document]
ziadhany: I think because we improve vulnerability one by one . Can you suggest any way to do this?

pombredanne:
@ziadhany Could you rebase or merge on the latest main branch?

pombredanne (approved):
LGTM!

PR #784, opened 2022-06-27
Add fixed packages in packages endpoint

TG1999:

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -50,7 +50,34 @@ class MinimalPackageSerializer(serializers.HyperlinkedModelSerializer):
         fields = ["url", "purl"]
 
 
-class MinimalVulnerabilitySerializer(serializers.HyperlinkedModelSerializer):
+class FilteredPackageListSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        params = self.context["request"].query_params
+        name = params.get("name")
+        if name:
+            data = data.filter(name=name)
+        namespace = params.get("namespace")
+        if namespace:
+            data = data.filter(namespace=namespace)
+        type = params.get("type")
+        if type:
+            data = data.filter(type=type)
+        return super(FilteredPackageListSerializer, self).to_representation(data)
+
+
+class FixedPackageSerializer(serializers.ModelSerializer):
+
+    purl = serializers.CharField(source="package_url")
+
+    class Meta:
+        list_serializer_class = FilteredPackageListSerializer
+        model = Package
+        fields = ["url", "purl"]
+
+
+class MinimalVulnerabilitySerializerWithReferencesAndSummary(
+    serializers.HyperlinkedModelSerializer
+):
     """
     Used for nesting inside package focused APIs.
     """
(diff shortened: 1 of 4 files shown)


tdruez (changes requested):
@TG1999 The new code looks fine.
You should add a simple unit test for the `fixed_packages` property.
Model properties should be easy to test and you should always add a unit test when you add new properties.

vulnerabilities/api.py
```diff
@@ -50,7 +50,34 @@ class Meta:
...
+        params = self.context["request"].query_params
+        name = params.get("name")
+        if name:
+            data = data.filter(name=name)
+        namespace = params.get("namespace")
+        if namespace:
+            data = data.filter(namespace=namespace)
+        type = params.get("type")
+        if type:
+            data = data.filter(type=type)
```
tdruez, line 64: Why not using proper filters instead of this duplicated logic?

vulnerabilities/api.py
```diff
@@ -62,6 +89,31 @@ class Meta:
...
+    """
+    Used for nesting inside package focused APIs.
+    """
+
+    class Meta:
+        model = Vulnerability
+        fields = ["url", "vulnerability_id"]
+
+
+class MinimalPackageSerializerWithFixedVulnerabilities(serializers.HyperlinkedModelSerializer):
```
tdruez, line 102: The class name is too long thus not readable.

tdruez (changes requested):
@TG1999 Your code is not tested. How can we know it works?

vulnerabilities/api.py
```diff
@@ -193,7 +224,27 @@ class Meta:
...
+        if type:
+            query &= Q(type=type)
+        queryset = Vulnerability.objects.prefetch_related(
+            Prefetch(
+                "packages",
+                queryset=Package.objects.filter(query, packagerelatedvulnerability__fix=True),
+                to_attr="filtered_fixed_packages",
+            )
+        )
+        return queryset
```
tdruez, line 246: This is still duplicated code and much harder to read and understand than the previous code.
This is not what I meant by "proper filters" but I was rather referring to https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend.
Any reason to not use a FilterSet here?

tdruez (changes requested):

vulnerabilities/api.py
TG1999: The filters applied by me are for Package model and if the viewset is on Vulnerabilities model

tdruez: I understand, but relational fields are not doable on FilterSet?

TG1999: I can do something like this https://github.com/philipn/django-rest-framework-filters#filtering-across-relationships, but the filter will then look like `api?package__name=name` not `api?name=name`

tdruez: > `api?package__name=name`

Is it a problem?

tdruez: @TG1999 you already defined a PackageFilterSet, you should use it, for example:

```
class VulnerabilityViewSet(viewsets.ReadOnlyModelViewSet):
    def get_fixed_packages_qs(self):
        package_filter_data = {"packagerelatedvulnerability__fix": True}

        query_params = self.request.query_params
        for field_name in ["name", "namespace", "type"]:
            value = query_params.get(field_name)
            if value:
                package_filter_data[field_name] = value

        return PackageFilterSet(package_filter_data).qs

    def get_queryset(self):
        queryset = Vulnerability.objects.prefetch_related(
            Prefetch(
                "packages",
                queryset=self.get_fixed_packages_qs(),
                to_attr="filtered_fixed_packages",
            )
        )
        return queryset
```

Also, make sure to add `packagerelatedvulnerability__fix` to the PackageFilterSet.fields.

Lastly, you should always document such methods using the docstring, it is not obvious to the reader what is happening here and why we are prefetching filtered package data.

tdruez (approved):

PR #812, opened 2022-07-26
Fix data migration

TG1999:


pombredanne (changes requested):
I do not think you can change the existing migration that has already been applied. Instead you must create a new migration from scratch that assumes that the data state is that of applying the first migration and then does fix the data accordingly.
Also can you add some detailed explanation of the issue in the PR message?

pombredanne:
Closing in favor of the new #818

Issue #798, opened 2022-07-14
Update VCIO Landing Page


pombredanne:
See also #772

mjherzog:
Update the VCIO landing page to emulate the look and feel of SCIO. Some key points are:
- Narrow the page “width” to emulate the SCIO landing page
- Change the page “header” to emulate the SCIO page header with the items:
  - VulnerableCode.io
   - Packages
   - Vulnerabilities
   - API Documentation (we do not have the same API access as with SCIO)
   - Version (this will need to be the version of VulnerableCode - we can document that on hover)
- Keep the page footer (we are missing a footer on the SCIO landing page - it should have one)
- Show two search sections SCIO style
   - Packages
   - Vulnerabilities
   There is no data to display here as we have for Projects because the only VCIO functionality is to search for a specific Package or Vulnerability.
   - We should display a block of simple help text under each Search box that tells you what to do/expect - 
     - For Packages we might say: “Search for currently known Vulnerabilities for a Package-URL or a Package name.”
     - For Vulnerabilities we might say: “Search for complete information for a VULCOID (VulnerableCode Database ID).”
- Bonus item - add a simple favicon. Just the letters “VC” with a light blue background would suffice for now.

pombredanne:
I think we should revisit the design for search to have a single search box rather than to have multiple search boxes. The content that is searched can help determine easily what we are searching. For instance `CVE` prefix in the search would mean searching for vulnerabilities. Or a pkg prefix would mean this is a purl. Also in general we should be flexible in what we accept for search and not reject anything, possibly suggesting alternative searches like with a "did you mean?"

mjherzog:
A wise man once told me to avoid premature optimization.

johnmhoran:
I'm adding tasks beyond those initially identified above:
- [ ] Move landing page search instructions to icon-hover displays
- [ ] Add relevant search box/capability to the package/purl and vulnerability search results and details pages
- [ ] Add info to package/purl details page reflecting closest version with no unresolved reported vulnerabilities

(More to come.)

johnmhoran:
This has been completed.

PR #813, opened 2022-07-27
Modernize UI #798

johnmhoran:
This PR revamps minimally the UI using the same overall look and feel
as ScanCode.io.

Further work beyond the basics including addressing the reviewers
feedback will take place in:
https://github.com/nexB/vulnerablecode/pull/847

Reference: https://github.com/nexB/vulnerablecode/issues/798
Signed-off-by: John M. Horan <johnmhoran@gmail.com>


pombredanne:
Here is some comments on the current state of this PR... The goal would be to have a minimally new UI working so we can release it in verision 30.0:

## On the landing page

remove template in navbar


## third-party JS

- Remove https://ajax.googleapis.com/ajax/libs/jquery/3.6.0/jquery.min.js https://github.com/nexB/vulnerablecode/blob/441852f29a3d43b5ccf898ebdb1c405cb6e402d5/vulnerabilities/templates/vulnerability.html#L515 we cannot source JS from third-party sites. We would instead keep a Jquery under version control https://github.com/nexB/scancode.io/tree/main/scancodeio/static 

## Package search

- Replace HTML by proper link in the packages search results list page
- In the package detail view hide everything after fixing vulnerabilities

## Vulnerabilities search

- Replace HTML by proper link in the vulnerabilities search results list page

- In the vulnerabilities detail view hide everything after fixed package
  - hide everything after aliases on the essnetial tabs
  - Populate Affected Packages et Fixed Packages

## for later:

  - limit searches to a minimum number of characters

pombredanne:
Here are a few more suggestions:

1. for the package details, we could have no tabs and everything in one view

vcio-package-details-mock


2. for the vulnerabilities details, a more compact view that would also display affected and fixed packages.
- The first few affected packages and first few fixed packages would be always listed on the main tab.
- There would be a ... (dot dot dot) linking to the tab with the full table
- Full details would be in the tabs.

The count of references, and other items would be added to the right of the tab title inside parenthesis as in `References (23)`

vcio-vuln-details-mock

johnmhoran:
Need to fix the sorting for all lists in the UI.

johnmhoran:
Change documentation link in navbar to the RTD.

vulnerabilities/templates/index.html
```diff
@@ -1,14 +1,166 @@
...
+            Search for vulnerable packages
+            <div class="dropdown is-hoverable" style="font-weight: normal;">
+                <div class="dropdown-trigger" style="text-align: right;">
+                    <i class="fa fa-question-circle" style="font-size:20px; margin-left: 10px;"></i>
+                </div>
+                <div class="dropdown-menu" id="dropdown-menu4" role="menu" style="width: 600px;">
+                    <div class="dropdown-content" style="box-shadow: 0px 8px 16px 0px #808080;">
+                        <div class="dropdown-item" style="font-size: 14px;">
+                            <div style="margin: 0px 0px 0px 0px;">
+                                <div style="margin-top: 0px;">Search for currently known Vulnerabilities for a <a href="https://github.com/package-url/purl-spec" target="_blank">Package-URL</a> (aka <span class="inline-code01">purl</span>) or a Package <span class="inline-code01">name</span>.</div>
```
tdruez, line 30: Margin classes are available in the CSS framework for those style, for example:
`class="m-0"` for no margins at all or `class="mt-0"` for no margin on top.

vulnerabilities/templates/base.html
```diff
@@ -1,53 +1,43 @@
...
-                <a class="navbar-item" href="{% url 'package_search' %}">
-                    Packages
-                </a>
-                <a class="navbar-item" href="{% url 'vulnerability_search' %}">
-                    Vulnerabilities
-                </a>
+    <div class="container max-desktop-width is-fullheight">
+        {% block content %}{% endblock %}
+
+        <div style="margin-top: 50px;"></div>
```
tdruez, line 25: It would be better to use the available margin CSS classes to ensure consistency in the spacing instead of using arbitrary values. See https://bulma.io/documentation/helpers/spacing-helpers/

vulnerabilities/templates/packages.html
```diff
@@ -1,64 +1,202 @@
...
-  <div class="column is-one-third is-offset-one-third">
-    <form action="{% url 'package_search' %}" method="get">
-    {% load widget_tweaks %}
-      <div class="field is-horizontal pd-3">
-        <div class="select">
-        {{form.type}}
+
+{% include 'navbar.html' %}
+
+<section class="section01">
```
tdruez, line 12: Use descriptive names instead of numbers.

vulnerabilities/templates/packages.html
```diff
@@ -1,64 +1,202 @@
...
+            {% if packages.number != packages.paginator.num_pages %}
+            {% if packages.next_page_number != packages.paginator.num_pages %}
+            <li>
+                <span class="pagination-ellipsis">&hellip;</span>
+            </li>
+            {% endif %}
+            <li>
+                <a href="?{{ searched_for }}&page={{ packages.paginator.num_pages }}{% if url_params_without_page %}&{{ url_params_without_page }}{% endif %}" class="pagination-link" aria-label="Goto page {{ packages.paginator.num_pages }}">{{ packages.paginator.num_pages|intcomma }}</a>
+            </li>
+            {% endif %}
```
tdruez, line 170: Use indentation after tag such as `if` and `for` to make the code easier to read, especially here where you have imbricated conditions.
See some examples at https://docs.djangoproject.com/en/4.0/ref/templates/language/#tags

vulnerabilities/views.py
```diff
@@ -19,50 +19,118 @@
...
 from django.views.generic.edit import UpdateView
 from django.views.generic.list import ListView
+from packageurl import PackageURL
+from packaging import version
 
 from vulnerabilities import forms
 from vulnerabilities import models
+from vulnerabilities.forms import CVEForm
+from vulnerabilities.forms import PackageForm
+from vulnerablecode import settings
```
tdruez, line 29: https://docs.djangoproject.com/en/4.1/topics/settings/#using-settings-in-python-code

> Also note that your code should not import from either global_settings or your own settings file.

Use `from django.conf import settings` instead.

vulnerabilities/views.py
```diff
@@ -19,50 +19,118 @@
...
         if len(request.GET["type"]):
             package_type = request.GET["type"]
 
         if len(request.GET["name"]):
             package_name = request.GET["name"]
 
+        # Check whether the input value is a syntactically-correct purl
+        try:
+            purl = PackageURL.from_string(package_name)
+            return list(
```
tdruez, line 111: The QS should live outside the try block.

vulnerabilities/views.py
```diff
@@ -100,40 +200,87 @@ def request_to_vulnerabilities(request):
...
+
         resolved_vuln = [i for i in package.resolved_to]
         unresolved_vuln = [i for i in package.vulnerable_to]
 
+        resolved_vuln = sorted(resolved_vuln, key=lambda x: x.vulnerability_id)
+        unresolved_vuln = sorted(unresolved_vuln, key=lambda x: x.vulnerability_id)
+
         return resolved_vuln, unresolved_vuln
 
+    def _related_packages(self):
```
tdruez, line 256: This should be moved to the models.py as a method on the Package QuerySet.

vulnerabilities/views.py
```diff
@@ -19,50 +19,118 @@
...
+
+            if len(request.GET["name"]):
+                package_name = request.GET["name"]
+                context["package_name"] = package_name
+
+            if result_size == 0:
+                context = {
+                    "package_search": "The VCIO DB does not contain a record of the package you entered -- "
+                    + request.GET["name"]
+                    + ".",
```
tdruez, line 69: Use f-strings for better readability:
```
search_string = request.GET["name"]
"package_search": f"The VCIO DB does not contain a record of the package you entered: {search_string}."
```

johnmhoran:
@tdruez Re my using too many `style` attributes, I'll do my best to replace these with Bulma classes as you requested, but it will take a fair amount of time to figure out what the Bulma version is of a specific display style I've fine-tuned, and not all likely are covered by the class choices provided by Bulma.  Working on this now.

vulnerabilities/views.py
johnmhoran: Thanks @tdruez -- changed.

vulnerabilities/views.py
johnmhoran: @tdruez I don't understand -- what's the reason to move the Qs outside the `try` block, and how would that be done?

vulnerabilities/views.py
johnmhoran: @tdruez What's the reason to do this, and how would I implement?

vulnerabilities/views.py
johnmhoran: Perhaps you're suggesting that I move the entire `return()` into a new method, leaving only `purl = PackageURL.from_string(package_name)`, which tests whether the input is a syntactically-correct purl?  Pass `purl` to the new method, keeping the `try` focused on its sole purpose: the `purl` test?

vulnerabilities/views.py
johnmhoran: I've not yet had a chance to give this a careful look.  If our schedule permits (we don't have much overlap, and it might be another day before I could followup again if need be), I'd prefer to do that first and give it a try, although my familiarity with Django models is still rather shallow.  

Re the rationale -- I modelled `def _related_packages(self)` on `def _package_vulnerabilities(self)`, which is just above it inside `class PackageUpdate(UpdateView)`.  Why does one of these belong in `models.py` as a method on the `Package` QuerySet while the other belongs in the view?

And another newb question: does adding it as a QuerySet method just mean adding it as a method to the model, e.g, `def related_packages(self, package_url)` -- no preceding `_` -- and figuring out how to pass this to the view/template?

vulnerabilities/views.py
tdruez: > what's the reason to move the Qs outside the try block

In general, you want to keep only one piece of logic that could fail within a `try` block.
An exception raised by converting a PURL has nothing to do with an exception filtering a QuerySet.

Also, using generic `except:` is bad practice. You should always try to be explicit about the exception you want to catch. Since you want to "Check whether the input value is a syntactically-correct purl", let's catch that specific Exception:
```
>>> PackageURL.from_string('wrong syntax')
  File "<console>", line 1, in <module>
  File "/Volumes/nexB/repos/scancode.io/lib/python3.9/site-packages/packageurl/__init__.py", line 354, in from_string
    raise ValueError(
ValueError: purl is missing the required "pkg" scheme component: {repr(purl)}.
```

The packageurl raise a `ValueError` in that case, that's the exception we want to catch.
You code could become:

```
try:
    purl = PackageURL.from_string(package_name)
except ValueError:
    purl = None

if purl:
    packages = ...
else:
    packages = ...

return list(packages)
```

This way, this are properly grouped, the code is more readable:
1. we deal with the purl
2. we prepare a queryset based on the available data
3. we return the fetched objects

> keeping the try focused on its sole purpose: the purl test?

Yes, bottom line is to always try to keep you `try` block as focused as possible.

johnmhoran: Far out.  Thank you for the detailed explanation @tdruez .  I'll tackle this in the A.M.  👍

vulnerabilities/views.py
tdruez: > What's the reason to do this, and how would I implement?

To keep thing properly organized, re-usable, and easier to test.

In a Django app, code related to the database transaction belongs in the models.

> Why does one of these belong in models.py as a method on the Package QuerySet while the other belongs in the view?

I don't see any QuerySet logic in `_package_vulnerabilities`.

> And another newb question: does adding it as a QuerySet method just mean adding it as a method to the model, e.g, def related_packages(self, package_url) -- no preceding _ -- and figuring out how to pass this to the view/template?

Not exactly, it means adding this on the QuerySet class associated with a model.

Some examples:
https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L1215
https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L960

The first step is to add a QuerySet class to your Package model:

```
class PackageQuerySet(models.QuerySet):    
```

You can then start to break your various filters and annotation into reusable methods on the PackageQuerySet class:

```
return list(
            models.Package.objects.all()
            .filter(
                Q(
                    type=purl.type,
                    namespace=purl.namespace,
                    name=purl.name,
                    subpath=purl.subpath,
                    qualifiers=purl.qualifiers,
                )
            )
            .order_by("version")
            .annotate(
                vulnerability_count=Count(
                    "vulnerabilities",
                    filter=Q(packagerelatedvulnerability__fix=False),
                ),
                # TODO: consider renaming to fixed in the future
                patched_vulnerability_count=Count(
                    "vulnerabilities",
                    filter=Q(packagerelatedvulnerability__fix=True),
                ),
            )
            .prefetch_related()
        )
```

Could become:


```
class PackageQuerySet(models.QuerySet): 
    def for_package_url(self, purl):
        return self.filter(
                type=purl.type,
                namespace=purl.namespace,
                name=purl.name,
                subpath=purl.subpath,
                qualifiers=purl.qualifiers,
        )

    def with_vulnerability_counts(self):
        return self.annotate(
                vulnerability_count=Count(
                    "vulnerabilities",
                    filter=Q(packagerelatedvulnerability__fix=False),
                ),
                patched_vulnerability_count=Count(
                    "vulnerabilities",
                    filter=Q(packagerelatedvulnerability__fix=True),
                ),
            )
```

Then on the view side, you have something much more readable such as:

```
return list(
    models.Package.objects
    .for_package_url(purl)
    .order_by("version")
    .with_vulnerability_counts()
    .prefetch_related()
)
```

Add this on your Package model before the "Meta" class and after all fields, such as https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L1469

```
````

You now have the database logic organized in the models, you can re-use the count annotations without copy pasting the code into other views, and you view logic is more focused and readable.


Note: The code above is not tested, simply here to illustrate, it'll have to be adapted and refined.

File: vulnerabilities/views.py [hunk 9]
```diff
@@ -19,50 +19,118 @@
...
 
         if len(request.GET["name"]):
             package_name = request.GET["name"]
 
+        # Check whether the input value is a syntactically-correct purl
+        try:
+            purl = PackageURL.from_string(package_name)
+            return list(
+                models.Package.objects.all()
+                .filter(Q(type=purl.type, name=purl.name, version=purl.version))
```
tdruez, line 113: `Q` object is not needed here, it is only required for complex query such as applying OR between filters such as https://github.com/nexB/vulnerablecode/pull/813/files#diff-3e8fc96d993d0bd8585642f68239732a30a19b7ec100a36ee5d21e209625ff81R202.
This comment applies to all the place you have a single Q in your filter().

vulnerabilities/views.py
```diff
@@ -19,50 +19,118 @@
...
             package_type = request.GET["type"]
 
         if len(request.GET["name"]):
             package_name = request.GET["name"]
 
+        # Check whether the input value is a syntactically-correct purl
+        try:
+            purl = PackageURL.from_string(package_name)
+            return list(
+                models.Package.objects.all()
```
tdruez, line 112: `.all()` is not needed if you call `.filter()` after.

vulnerabilities/views.py
```diff
@@ -19,50 +19,118 @@
...
+            )
+        except:
+            pass
+
         return list(
             models.Package.objects.all()
-            # FIXME: This filter is wrong and ignoring most of the fields needed for a
-            # proper package lookup: type/namespace/name@version?qualifiers and so on
             .filter(name__icontains=package_name, type__icontains=package_type)
+            .order_by("type", "namespace", "name", "version", "subpath", "qualifiers")
```
tdruez, line 133: This is also a good candidate to move on the PackageQuerySet class:

```
def order_by_purl(self):
    self.order_by("type", "namespace", "name", "version", "subpath", "qualifiers")
```

Then on the view side:

```
models.Package.objects.[...].order_by_purl()
```

tdruez:
- [ ] Fix the labels and wording case consistency

vulnerabilities/templates/index.html
johnmhoran: Done.

vulnerabilities/templates/base.html
johnmhoran: Done.

vulnerabilities/templates/index.html
```diff
@@ -1,14 +1,166 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="panel-block">
+            <div style="padding-bottom: 10px; width: 100%;">
+                <form action="{% url 'package_search' %}" method="get">
+                    {% load widget_tweaks %}
```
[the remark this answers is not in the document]
johnmhoran: Done.

vulnerabilities/templates/index.html
```diff
@@ -1,14 +1,166 @@
...
+                    <div class="field has-addons" style="margin: 10px 0px 0px 0px;">
+                        <div class="control">
+                            <div class="select">
+                                {{package_form.type}}
+                            </div>
+                        </div>
+                        <div class="control" style="width: 100%;">
+                            {{package_form.name|add_class:"input"}}
```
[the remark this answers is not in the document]
johnmhoran: Done.

vulnerabilities/templates/packages.html
johnmhoran: Done.

vulnerabilities/templates/packages.html
johnmhoran: @tdruez I've started applying the indentation to the Jinja if/for tags, only to notice that when I save the file, some "feature" of VSCode is returning all HTML to the indented format structure one would expect if there were no Jinja templates.  Time for some troubleshooting....

johnmhoran: I've installed the Django extension from Baptiste Darthenay and now I can save Jinja tag indents, but my HTML intellisense/auo-complete has disappeared.

johnmhoran: Finished properly indenting all Jinja2 if/for tags.  VSCode problem has been widely discussed on StackOverflow and elsewhere for several years, but I've found no "solutions" that (1) enable the user to save properly-indented Jinja tags -and- (2) provide the usual VSCode HTML intellisense and auto-complete.  (1) is clearly necessary, and perhaps someday (2) will also be available at the same time.  🙂

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/876

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/877

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/878

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/879

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/880

vulnerabilities/views.py
johnmhoran: See new issue https://github.com/nexB/vulnerablecode/issues/880

DennisClark:
The sorting issue described in issue #872  still exists for the package search in VCIO v36.0.0.

PR #830, opened 2022-08-07
Paginate initial listings to display a small number of or records

TG1999:

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -177,7 +177,6 @@ class PackageFilterSet(filters.FilterSet):
 class PackageViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Package.objects.all()
     serializer_class = PackageSerializer
-    paginate_by = 50
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = PackageFilterSet
 
@@ -252,7 +251,6 @@ class VulnerabilityViewSet(viewsets.ReadOnlyModelViewSet):
         )
 
     serializer_class = VulnerabilitySerializer
-    paginate_by = 50
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = VulnerabilityFilterSet
 
@@ -270,7 +268,6 @@ class CPEViewSet(viewsets.ReadOnlyModelViewSet):
         vulnerabilityreference__reference_id__startswith="cpe"
     ).distinct()
     serializer_class = VulnerabilitySerializer
-    paginate_by = 50
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = CPEFilterSet
 
@@ -286,6 +283,5 @@ class AliasFilterSet(filters.FilterSet):
 class AliasViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Vulnerability.objects.all()
     serializer_class = VulnerabilitySerializer
-    paginate_by = 50
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = AliasFilterSet
--- a/vulnerablecode/settings.py
+++ b/vulnerablecode/settings.py
@@ -159,7 +159,7 @@ REST_FRAMEWORK = {
         "django_filters.rest_framework.DjangoFilterBackend",
         "rest_framework.filters.SearchFilter",
     ),
-    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
-    # Reduce to 10 records per page to display data almost instantaneously https://github.com/nexB/vulnerablecode/issues/819
+    "DEFAULT_PAGINATION_CLASS": "vulnerabilities.pagination.SmallResultSetPagination",
+    # Limit the load on the Database returning a small number of records by default. https://github.com/nexB/vulnerablecode/issues/819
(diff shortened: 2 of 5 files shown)


vulnerabilities/api.py
```diff
@@ -175,9 +176,9 @@ def filter_purl(self, queryset, name, value):
 
 
 class PackageViewSet(viewsets.ReadOnlyModelViewSet):
-    queryset = Package.objects.all()
+    queryset = Package.objects.all().order_by("id")
     serializer_class = PackageSerializer
-    paginate_by = 50
+    pagination.PageNumberPagination.page_size = 10
```
tdruez, line 181: This seems quite wrong. It looks like you are overriding a value for a completely different class reference.

vulnerabilities/api.py
```diff
@@ -175,9 +176,9 @@ def filter_purl(self, queryset, name, value):
 
 
 class PackageViewSet(viewsets.ReadOnlyModelViewSet):
-    queryset = Package.objects.all()
+    queryset = Package.objects.all().order_by("id")
```
tdruez, line 179: Why this change? If there is an issue with the ordering, please create a new issue and discuss the changes first.
This may need to be implement on the Model instead of on the API side. But the is no explanation about why you are changing all the ordering in the API context.

vulnerablecode/settings.py
```diff
@@ -160,5 +160,5 @@
         "rest_framework.filters.SearchFilter",
     ),
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
-    "PAGE_SIZE": 100,
+    "PAGE_SIZE": 10,
```
tdruez, line 163: Isn't this enough? The paginate_by on ViewSet is only useful to override the default setting, no?

tdruez (changes requested):
The need for changing ordering in the context of refining the pagination is unclear.
See my inline comments.

Also, could you confirm that `bulk_search` actions is not limited to this 10 items pagination?
If I'm running a bulk search on 100 items, I do not want to have to make 10 requests to get the 100 items as results.

vulnerablecode/settings.py
TG1999: @tdruez yes, you are right, this is enough

vulnerabilities/api.py
TG1999: @tdruez there is an issue reported here for same https://github.com/nexB/vulnerablecode/issues/685, but we can tackle this in a separate PR, will remove unrelated changes

tdruez: 1 issue -> 1 PR.

Also, this fix approach for #685 seems wrong.

TG1999: Alright! will fix this by changing in models side not at the API side

vulnerablecode/settings.py
tdruez, line 163: A comment about why we want to keep the default page size low would be useful.

TG1999: Added the comment

vulnerablecode/settings.py
```diff
@@ -160,5 +160,6 @@
         "rest_framework.filters.SearchFilter",
     ),
     "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
-    "PAGE_SIZE": 100,
+    # Reduce to 10 records per page to display data almost instantaneously https://github.com/nexB/vulnerablecode/issues/819
```
tdruez, line 163: > almost instantaneously

This may not always be true, there are cases where displaying 10 records is not instantaneous.

Maybe something along:
Limit the load on the Database returning a small number of records by default.

tdruez (changes requested):
Since we are already using the [PageNumberPagination](https://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination), I think we should enable the `page_size` feature, so the consumer can decide to get more items per requests when needed.
Let's keep a reasonable `max_page_size` though to avoid overloading the database.

vulnerabilities/api.py
```diff
@@ -177,9 +183,9 @@ def filter_purl(self, queryset, name, value):
 class PackageViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Package.objects.all()
     serializer_class = PackageSerializer
-    paginate_by = 50
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = PackageFilterSet
+    pagination_class = SmallResultSetPagination
```
tdruez, line 188: Why declaring this on the ViewSet instead of globally using the `DEFAULT_PAGINATION_CLASS` setting?

vulnerabilities/api.py
```diff
@@ -24,6 +25,11 @@
 from vulnerabilities.models import VulnerabilitySeverity
 
 
+class SmallResultSetPagination(PageNumberPagination):
+    page_size_query_param = "page_size"
+    max_page_size = 30
```
tdruez, line 30: It could be useful to allow up to 100.

tdruez (changes requested):

tdruez:
@TG1999 We're missing an entry in the CHANGELOG about those changes.

tdruez (approved):
LGTM!

TG1999:
Thanks! Merging

PR #831, opened 2022-08-07
Add fixed packages in vulnerabilities details in packages endpoint.

TG1999:


vulnerabilities/api.py
```diff
@@ -175,9 +203,9 @@ def filter_purl(self, queryset, name, value):
 
 
 class PackageViewSet(viewsets.ReadOnlyModelViewSet):
-    queryset = Package.objects.all()
+    queryset = Package.objects.all().order_by("id")
     serializer_class = PackageSerializer
-    paginate_by = 50
+    pagination.PageNumberPagination.page_size = 10
```
tdruez, line 208: This seems very wrong plus it's unrelated to this PR.

tdruez (changes requested):
It looks like you have included changes completely unrelated to the goal of this PR:
- Pagination
- Ordering

vulnerabilities/api.py
TG1999: I have cut a branch from #830, I will rebase this branch once #830 get's merged

vulnerabilities/api.py
```diff
@@ -128,13 +117,52 @@ def to_representation(self, instance):
...
+            namespace=package.namespace,
+            type=package.type,
+            qualifiers=package.qualifiers,
+            subpath=package.subpath,
+            packagerelatedvulnerability__fix=True,
+        ).distinct()
+
+    def get_affected_vulnerabilities(self, package):
+        fixed_packages = self.get_fixed_packages(package=package)
+        qs = package.vulnerabilities.filter(packagerelatedvulnerability__fix=False)
```
tdruez, line 137: This is a bit confusing, `packagerelatedvulnerability__fix` is used with True in `get_fixed_packages` then overridden with False here?

TG1999: This qs is to get those vulnerabilities on the package that affects this that' why `packagerelatedvulnerability__fix` is set to False, whereas `get_fixed_packages` sends all the packages that matches name, namespace, type, qualifiers of that package and also fixes the vulnerability that's why it is set to True

tdruez: I get that, the confusing part is your approach of reusing the same base qs and overriding an existing lookup.

tdruez (approved):

PR #844, opened 2022-08-15
Add bulk search support for CPEs #808

TG1999:

### DIFF
--- a/vulnerabilities/tests/test_fix_api.py
+++ b/vulnerabilities/tests/test_fix_api.py
@@ -270,3 +270,37 @@ class BulkSearchAPI(TestCase):
             content_type="application/json",
         ).json()
         assert len(response) == 13
+
+
+class BulkSearchAPI(TestCase):
+    def setUp(self):
+        cpes = [
+            "cpe:/a:nginx:1.0.7",
+            "cpe:/a:nginx:1.0.15",
+            "cpe:/a:nginx:1.14.1",
+            "cpe:/a:nginx:1.15.5",
+            "cpe:/a:nginx:1.15.6",
+            "cpe:/a:nginx:1.16.1",
+            "cpe:/a:nginx:1.17.2",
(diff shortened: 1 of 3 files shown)


vulnerabilities/tests/test_fix_api.py
```diff
@@ -270,3 +270,37 @@ def test_api_response(self):
...
+    def test_api_response_with_one_vulnerability(self):
+        request_body = {
+            "cpes": self.cpes,
+        }
+        response = self.client.post(
+            "/api/cpes/bulk_search",
+            data=request_body,
+            content_type="application/json",
+        ).json()
+        assert len(response) == 1
```
tdruez, line 306: You need to test the content of the response. The current assertion is too weak and if the content of the response changes this test may not catch it.

tdruez (changes requested):
The unit test needs some improvements.
Also, I think we are missing a changelog entry.

vulnerabilities/tests/test_fix_api.py
TG1999: Improved assertions

vulnerabilities/tests/test_fix_api.py
```diff
@@ -270,3 +270,40 @@ def test_api_response(self):
...
+        ]
+        self.cpes = cpes
+        vuln = Vulnerability.objects.create(summary="test")
+        for cpe in cpes:
+            ref = VulnerabilityReference.objects.create(reference_id=cpe)
+            VulnerabilityRelatedReference.objects.create(reference=ref, vulnerability=vuln)
+
+    def test_api_response_with_one_vulnerability(self):
+        request_body = {
+            "cpes": self.cpes,
```
tdruez, line 299: A better approach would be to only request a few of the existing cpes along some non existing ones.
Testing on everything is risky as the expecting results can be the same as not filtering at all.

tdruez (approved):

Issue #460, opened 2021-05-22
Add API rate limiting


sbs2001:
Deploying without any guards against API spam, is a bad idea. Especially since we need to do quite a work and API payload is big.

pombredanne:
Using the DRF AnonRateThrottle should be good enough for a start

pombredanne:
At this stage we do not have anythin beyond basic auth

We should enable the "classic" API key system for DRF API access.

For now we barely enabled  "django.contrib.auth", and "rest_framework.authtoken" in the settings https://github.com/nexB/vulnerablecode/blob/479111359070cc09010bde343e210306c4b14e40/vulnerablecode/settings.py ...

But we should emulate what is done in ScanCode.io ... see https://github.com/nexB/scancode.io/pull/368 and https://github.com/nexB/scancode.io/issues/359

pombredanne:
For now, I think we will not need rate limiting once we implement auth... that's a refinement for later

TG1999:
Added authentication here https://github.com/nexB/vulnerablecode/pull/848

pombredanne:
As a first step when we make the public instance public I suggest this:
- no auth for the UI access
- auth for API access with an API key, that can be requested simply by email for a start.

We need to add link/content in the UI to explain how to request an API key.

PR #848, opened 2022-08-16
Add authentication

TG1999:


vulnerablecode/auth.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class ConditionalLoginRequired(UserPassesTestMixin):
+    """
+    CBV mixin for views that checks that the current user is authenticated when
+    authentication is enabled.
+    """
+
+    def test_func(self):
+        return is_authenticated_when_required(self.request.user)
```
tdruez, line 49: Those are not used.

vulnerabilities/templates/base.html
```diff
@@ -49,5 +63,4 @@
     </footer>
 </body>
 {% block javascript %}
-{% endblock %}
-</html>
```
tdruez, line 53: No more <html> tag? Is this valid?

CHANGELOG.rst
```diff
@@ -59,6 +59,8 @@ Version v30.0.0
 
 - Add bulk search support for CPEs.
 
+- Add authentication for API.
```
tdruez, line 62: This could use more details.

vulnerabilities/templates/base.html
```diff
@@ -23,6 +23,20 @@
                 <a class="navbar-item" href="{% url 'vulnerability_search' %}">
                     Vulnerabilities
                 </a>
+                {% if user.is_authenticated %}
```
tdruez, line 26: I think we're missing a Sign In link when the user is anonymous.

tdruez (changes requested):

vulnerabilities/templates/base.html
TG1999: Yes this is valid HTML

tdruez: It may be but it's inconsistent with most of the web.

Also, the w3c validator complains about it:

> Consider adding a lang attribute to the html start tag to declare the language of this document.

TG1999: This html file does not have a starting html tag, but had a closing one, either we could remove the closing tag (that's what I have done right now) or add an opening html tag.

tdruez: > This html file does not have a starting html tag

Sounds like a bug to me.

Looking at https://github.com/nexB/vulnerablecode/blob/main/vulnerabilities/templates/base.html#L1
Someone mixed the DOCTYPE and the <html> tag.

TG1999: I have added an opening html tag now and left the closing tag as it is.

tdruez: You want to remove the `class="has-navbar-fixed-top"` from the DOCTYPE too.

TG1999: Done!

tdruez (approved):

TG1999:
Partially solves issue https://github.com/nexB/vulnerablecode/issues/460

PR #893, opened 2022-09-01
Add namespace filter in packages api

TG1999:

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -190,7 +190,15 @@ class PackageFilterSet(filters.FilterSet):
 
     class Meta:
         model = Package
-        fields = ["name", "type", "version", "subpath", "purl", "packagerelatedvulnerability__fix"]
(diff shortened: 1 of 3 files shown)


vulnerabilities/api.py
```diff
@@ -190,7 +190,15 @@ class PackageFilterSet(filters.FilterSet):
 
     class Meta:
         model = Package
-        fields = ["name", "type", "version", "subpath", "purl", "packagerelatedvulnerability__fix"]
+        fields = [
+            "name",
```
pombredanne, line 194: The list of fields should be ordered this way:
```
            "purl",
            "type",
            "namespace",
            "name",
            "version",
            "subpath",
            "packagerelatedvulnerability__fix",
```

pombredanne (approved):
There is a just a minor nit for your review... I will merge as-is.. but let's make sure we always order fieds in the right purl field hierarchy

Issue #875, opened 2022-08-25
Allow lowercase searches for VULCOID-*


johnmhoran:

johnmhoran:
Closed by https://github.com/nexB/vulnerablecode/pull/898

PR #898, opened 2022-09-05
Allow case insensitive search for VCIDs #875

TG1999:


johnmhoran:
@TG1999 LGTM.  👍 Note that we already had an issue for this: issue #875 .  Do we need/have a test?

johnmhoran (approved):

PR #901, opened 2022-09-05
register available datasources

keshav-space:
automatically register all the datasourse classes available inside `datasources` directory.


TG1999:
IMO it's better to register data sources by hand as done in vulnerablecode rather than this

keshav-space:
> IMO it's better to register data sources by hand as done in vulnerablecode rather than this

It's doing the exact same thing. Just one less thing to worry about for someone adding any new data source.

PR #905, opened 2022-09-07
Improve by adding CVE as Reference id if missing

pombredanne:
This new improver adds missing CVE reference id that were not created by their importer.

Note that the current approach for improvers seems to be restricted to advisories and should be relaxed to better handle the general cases. Here this improver does not deal with advisories at all.


pombredanne:
Merged in #909

PR #910, opened 2022-09-09
Make URL mandatory for reference model

TG1999:
Raise error when URL is not provided for a reference

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -119,7 +119,6 @@ class VulnerabilityReference(models.Model):
         )
         ordering = ["reference_id", "url"]
 
-    
     def save(self, *args, **kwargs):
         if not self.url:
             raise ValidationError("URL cannot be empty")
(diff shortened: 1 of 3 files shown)


vulnerabilities/models.py
```diff
@@ -119,6 +119,11 @@ class Meta:
         )
         ordering = ["reference_id", "url"]
 
+    def save(self, *args, **kwargs):
```
pombredanne, line 122: I am not sure this is the correct way to make a field mandatory.
Rather 
1. you need a data migration first to fix things
2. the model needs to be changed: https://github.com/nexB/vulnerablecode/pull/910/files#diff-7f9f2c92e7163b06d21fa139369d75caed6561d0368b60ea2516cece0220eb5bR108 this should be explicit blank=False, null=False
?

pombredanne:
I am closing this in favor of another approach in #909 that @TG1999  and I worked out together with advices from @tdruez

PR #927, opened 2022-09-26
Add creation and updation date in models

TG1999:


vulnerabilities/models.py
```diff
@@ -144,6 +148,10 @@ class VulnerabilityReference(models.Model):
 
     objects = BaseQuerySet.as_manager()
 
+    date_created = models.DateTimeField(auto_now_add=True, null=True)
```
pombredanne, line 151: How could this ever be null? I would rather have it not null.

pombredanne (changes requested):
Let's try to avoid nulls. Also we need a changelog entry for this.

vulnerabilities/models.py
TG1999: This won't be null ever, because `auto_now_add` is True, I will add a test to show this migration's results

Issue #929, opened 2022-09-28
Provide a way to fetch lists of all vulnerable CPEs and PURLs


tdruez:
Add extra actions on the `packages` and `cpes` API endpoint that would return a single list of all vulnerable PURLs and CPEs.

For examples:

```
GET /api/packages/all/

[
  "purl1",
  "purl2",
  "purl3",
  ...
]
```

```
GET /api/cpes/all/

[
  "cpe1",
  "cpe2",
  "cpe3",
  ...
]
```

Make sure to not include PURLs nor CPEs that are not vulnerable.

tdruez:
> Tushar Goel: should we add pagination for this https://github.com/nexB/vulnerablecode/pull/932 ?

> tomd What’s your reasoning? It kind of defeat the purpose of having the full list of purl in a single request.

> Tushar Goel The reasoning was that when we have a huge volume of purls, the request will take a long time to be fetched, so by pagination we can serve small list of purls quickly.

> tomd We’ll have to see how long on the large PURL dataset. We can then decide the type of pagination we may need. I’m not sure that fetching those 100 at the time will be faster than the large request. My suggestion is to deploy this new action and see the performance on the actual set of data before implementing anything else.

> Tushar Goel Got it!

pombredanne:
we have purls but not CPEs yet

Issue #880, opened 2022-08-25
Use models.py Package QuerySet method in place of views.py method


johnmhoran:
From @tdruez 's comments in https://github.com/nexB/vulnerablecode/issues/875 .

Excerpts:

![image](https://user-images.githubusercontent.com/11096678/186774517-109636a6-ee24-45eb-9c8c-1ff92e57c22d.png)

This should be moved to the models.py as a method on the Package QuerySet.

[me]

Re the rationale -- I modelled `def _related_packages(self)` on `def _package_vulnerabilities(self)`, which is just above it inside `class PackageUpdate(UpdateView)`.  Why does one of these belong in `models.py` as a method on the `Package` QuerySet while the other belongs in the view?

And another newb question: does adding it as a QuerySet method just mean adding it as a method to the model, e.g, `def related_packages(self, package_url)` -- no preceding `_` -- and figuring out how to pass this to the view/template?

[@tdruez]

> What's the reason to do this, and how would I implement?

To keep thing properly organized, re-usable, and easier to test.

In a Django app, code related to the database transaction belongs in the models.

> Why does one of these belong in models.py as a method on the Package QuerySet while the other belongs in the view?

I don't see any QuerySet logic in `_package_vulnerabilities`.

> And another newb question: does adding it as a QuerySet method just mean adding it as a method to the model, e.g, def related_packages(self, package_url) -- no preceding _ -- and figuring out how to pass this to the view/template?

Not exactly, it means adding this on the QuerySet class associated with a model.

Some examples:
https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L1215
https://github.com/nexB/scancode.io/blob/main/scanpipe/models.py#L960

The first step is to add a QuerySet class to your Package model:

```
class PackageQuerySet(models.QuerySet):    
```

You can then start to break your various filters and annotation into reusable methods on the PackageQuerySet class:

```
return list(
            models.Package.objects.all()
            .filter(
                Q(
                    type=purl.type,
                    namespace=purl.namespace,
                    name=purl.name,
                    subpath=purl.subpath,
                    qualifiers=purl.qualifiers,
                )
            )
            .order_by("version")
            .annotate(
                vulnerability_count=Count(
                    "vulnerabilities",
                    filter=Q(packagerelatedvulnerability__fix=False),
                ),
                # TODO: consider renaming to fixed in the future
                patched_vulnerability_count=Count(
                    "vulnerabilities",
[...]

PR #932, opened 2022-09-30
Add API endpoint to get all vulnerable packages #929

TG1999:
Reference: https://github.com/nexB/vulnerablecode/issues/929

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Release notes
 =============
 
+
+Version v31.0.0
+----------------
+
+- We added a new "/packages/all" API endpoint to get all Package URLs know to be vulnerable.
+
+
 Version v30.0.0
 ----------------
 
@@ -79,7 +86,6 @@ This is a major version that is not backward compatible.
 
 - The data license is now CC-BY-SA-4.0 as this is the highest common
   denominator license among all the data sources we collect and aggregate. 
-
 
 Other:
 
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -254,6 +254,12 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
 
         return Response(response)
 
+    @action(detail=False, methods=["get"])
+    def all(self, request):
+        vulnerable_packages = Package.objects.vulnerable().only(*PackageURL._fields)
+        vulnerable_purls = [str(package.purl) for package in vulnerable_packages]
+        return Response(vulnerable_purls)
+
 
 class VulnerabilityFilterSet(filters.FilterSet):
     class Meta:
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -94,7 +94,7 @@ class Vulnerability(models.Model):
         """
         Return packages that are vulnerable to this vulnerability.
         """
-        return self.packages.filter(packagerelatedvulnerability__fix=False)
+        return self.packages.vulnerable()
 
     @property
     def resolved_to(self):
@@ -197,6 +197,12 @@ class PackageQuerySet(BaseQuerySet, PackageURLQuerySet):
 
         else:
             return self.none()
+
+    def vulnerable(self):
+        """
+        Return all vulnerable packages.
+        """
+        return Package.objects.filter(packagerelatedvulnerability__fix=False).distinct()
 
 
 class Package(PackageURLMixin):
--- a/vulnerabilities/tests/test_fix_api.py
+++ b/vulnerabilities/tests/test_fix_api.py
@@ -12,7 +12,6 @@ import json
 from django.contrib.auth import get_user_model
 from django.test import TestCase
 from django.test import TransactionTestCase
-from django.utils.http import int_to_base36
 from packageurl import PackageURL
 from rest_framework import status
 from rest_framework.test import APIClient
@@ -109,6 +108,7 @@ class APITestCasePackage(TestCase):
             summary="test-vuln",
         )
         self.vuln = vuln
+        self.vulnerable_packages = []
         for i in range(0, 10):
             query_kwargs = dict(
                 type="generic",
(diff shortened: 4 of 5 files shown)


CHANGELOG.rst
```diff
@@ -77,6 +77,7 @@ Version v30.0.0
 - The data license is now CC-BY-SA-4.0 as this is the highest common
   denominator license among all the data sources we collect and aggregate. 
 
+- We added a new "/packages/all" endpoint in API to get all vulnerable packages.
```
pombredanne, line 80:
```suggestion
- We added a new "/packages/all" API endpoint to get all Package URLs know to be vulnerable.
```

vulnerabilities/tests/test_fix_api.py
```diff
@@ -258,6 +271,16 @@ def test_api_with_single_vulnerability_and_vulnerable_package(self):
             ],
         }
 
+    def test_api_with_all_vulnerable_packages(self):
+        response = self.csrf_client.get(f"/api/packages/all", format="json").data
+        assert len(response) == 11
+        response_urls = [package["url"] for package in response]
+        for package in self.vulnerable_packages:
+            assert package["url"] in response_urls
+        response_purls = [package["purl"] for package in response]
```
pombredanne, line 280: Why running the test twice?

pombredanne (changes requested):
Thanks! See my comments

vulnerabilities/tests/test_fix_api.py
```diff
@@ -258,6 +271,12 @@ def test_api_with_single_vulnerability_and_vulnerable_package(self):
             ],
         }
 
+    def test_api_with_all_vulnerable_packages(self):
+        response = self.csrf_client.get(f"/api/packages/all", format="json").data
+        assert len(response) == 11
+        for package in self.vulnerable_packages:
+            assert package in response
```
tdruez, line 278: I would be preferable to directly assert the content of response.

vulnerabilities/api.py
```diff
@@ -185,6 +185,17 @@ class Meta:
         ]
 
 
+class VulnerablePackageSerializer(serializers.HyperlinkedModelSerializer):
+    purl = serializers.CharField(source="package_url")
+
+    class Meta:
+        model = Package
+        fields = [
+            "url",
```
tdruez, line 194: Why including the `url` field?

vulnerabilities/api.py
```diff
@@ -254,6 +265,17 @@ def bulk_search(self, request):
 
         return Response(response)
 
+    @action(detail=False, methods=["get"])
+    def all(self, request):
+        vulnerable_packages = Package.objects.filter(
+            packagerelatedvulnerability__fix=False
+        ).distinct()
```
tdruez, line 272: Can you confirm this QS triggers only 1 DB query? It should be asserted in your unit test as this is critical for the performances of those type of specialized actions.

tdruez (changes requested):

vulnerabilities/tests/test_fix_api.py
TG1999: Done!

vulnerabilities/api.py
TG1999: Removed the URL field

vulnerabilities/api.py
TG1999: Asserted the number of queries, only one query is used for getting all vulnerable packages

vulnerabilities/tests/test_fix_api.py
```diff
@@ -124,6 +125,11 @@ def setUp(self):
                 vulnerability=vuln,
                 fix=False,
             )
+            self.vulnerable_packages.append(
+                dict(
+                    purl=str(vuln_package.purl),
+                )
```
tdruez, line 131: Why using `dict()` instead of `{}`?

vulnerabilities/tests/test_fix_api.py
tdruez: Where? You are still testing the len() of the content, not the content itself.

TG1999: @tdruez here

tdruez: Ok, but as a reader of the test, it's hard to figure out the content of the variable.

If I'm not mistaken, the content of the response is something like:
```
[
   {"purl": purl1},
   {"purl": purl2},
   ....
]
```

Could you confirm?

TG1999: I have tried to make the tests more readable, please check if it looks good

tdruez: It does look better. 
Now we have a problem with this data structure.
Returning a list of dicts instead of a simple list of purls make the response content larger than is needs to be.
In the context of returning hundred of thousands of records, this is not ideal.

TG1999: Done!

vulnerabilities/api.py
```diff
@@ -254,6 +254,14 @@ def bulk_search(self, request):
...
```
tdruez, line 261: We could add `.only(*PackageURL._fields)` to get the minimal data required from the DB to build the purls.

Also, `.filter(packagerelatedvulnerability__fix=False)` should be a method on the QuerySet since it's used in multiple places. It would improve code readability by using a simple method name.

tdruez: See also https://github.com/nexB/vulnerablecode/issues/880

TG1999: Done!

vulnerabilities/api.py
```diff
@@ -254,6 +254,24 @@ def bulk_search(self, request):
 
         return Response(response)
 
+    @action(detail=False, methods=["get"])
+    def all(self, _):
```
tdruez, line 258: Could you explain this change? I see the `request` arg is not used in this method, but that's a bad idea to rename function parameter with `_` imo.

vulnerabilities/models.py
```diff
@@ -198,6 +198,12 @@ def for_package_url_object(self, purl):
         else:
             return self.none()
 
+    def vulnerable(self):
```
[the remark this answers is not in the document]
TG1999: Done!

tdruez (approved):
Looking good!

Issue #954, opened 2022-10-18
Speed up vulnerability search in the web UI


TG1999:
Some issues:
- [ ] Searching for a vulnerability takes forever.
- [ ] I cannot open a vulnerability details page as it will timeout

PR #955, opened 2022-10-18
[WIP] Make search for vulnerabilities faster

pombredanne:
This is a PR for https://github.com/nexB/vulnerablecode/issues/954 and attempts to make some small and critically needed adjustments for performance in the web UI.

Some approaches:

1. Cascade queries from exact to approximate searches to avoid full table scans in all cases. This is a band-aid for now. The proper solution will likely require using full text search instead.
2. Avoid iceberg queries with "prefetch related" to limit the number of queries that are needed in the UI
3. Do not recreate querysets from scratch but instead allow these to be chained for simpler and correct code. 
4. Remove extra details from the vulnerability pacge: each package was further listing its related vulnerabilities creating an iceberg query.
5. Enable the django-debug-toolbar with a setting to easily profile queries on demand.


TG1999 (approved):
LGTM!

Issue #978, opened 2022-10-25
Ignore qualifiers and subpath from PURL search lookups


tdruez:
In the VulnerableCode UI, searching for packages with the following purl: `pkg:maven/log4j/log4j@1.2.17` returns 1 result.
Now add any qualifiers, for examples `pkg:maven/log4j/log4j@1.2.17?classifier=sources`: 0 results.

Since it is unlikely that VulnerableCode would reference the infinity of possible qualifiers, we should relax the purl matching to only lookup the `type`+`namespace`+`name`+`version` fields.

This needs to be applied to the API as well.

PR #980, opened 2022-10-26
Ignore qualifiers and subpath from PURL search lookups #978

TG1999:

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -210,7 +210,10 @@ class PackageFilterSet(filters.FilterSet):
                 detail={"error": f'"{purl}" is not a valid Package URL: {ve}'},
             )
 
-        attrs = {k: v for k, v in purl.to_dict().items() if v}
+        ignorable_fields = ["subpath", "qualifiers"]
+
+        attrs = {k: v for k, v in purl.to_dict().items() if v and k not in ignorable_fields}
+
         return self.queryset.filter(**attrs)
 
 
@@ -239,8 +242,13 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
                 purl = PackageURL.from_string(purl).to_dict()
             except ValueError:
                 return Response(status=400, data={"Error": f"Invalid Package URL: {purl}"})
+            ignorable_fields = ["subpath", "qualifiers"]
             purl_data = Package.objects.filter(
-                **{key: value for key, value in purl.items() if value}
+                **{
+                    key: value
+                    for key, value in purl.items()
+                    if value and key not in ignorable_fields
+                }
             )
             purl_response = {}
             if purl_data:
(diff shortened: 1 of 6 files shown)


vulnerabilities/api.py
```diff
@@ -210,7 +210,10 @@ def filter_purl(self, queryset, name, value):
                 detail={"error": f'"{purl}" is not a valid Package URL: {ve}'},
             )
 
-        attrs = {k: v for k, v in purl.to_dict().items() if v}
+        ignorable_fields = ["subpath", "qualifiers"]
+
+        attrs = {k: v for k, v in purl.to_dict().items() if v and k not in ignorable_fields}
+
```
tdruez, line 216: I would suggest:

In the models.py:
```
def get_purl_query_lookups(purl):
    lookup_fields = ["type", "namespace", "name", "version"]
    return {
        field_name: value
        for field_name, value in purl.to_dict().items()
        if value and field_name in lookup_fields
    }
```

Then:
```
lookups = get_purl_query_lookups(purl)
return self.queryset.filter(**lookups)
```

vulnerabilities/api.py
```diff
@@ -239,8 +242,13 @@ def bulk_search(self, request):
...
             except ValueError:
                 return Response(status=400, data={"Error": f"Invalid Package URL: {purl}"})
+            ignorable_fields = ["subpath", "qualifiers"]
             purl_data = Package.objects.filter(
-                **{key: value for key, value in purl.items() if value}
+                **{
+                    key: value
+                    for key, value in purl.items()
+                    if value and key not in ignorable_fields
+                }
```
tdruez, line 251:
```
lookups = get_purl_query_lookups(purl)
purl_data = Package.objects.filter(**lookup)
```

tdruez (changes requested):

vulnerabilities/api.py
```diff
@@ -210,8 +210,17 @@ def filter_purl(self, queryset, name, value):
                 detail={"error": f'"{purl}" is not a valid Package URL: {ve}'},
             )
 
-        attrs = {k: v for k, v in purl.to_dict().items() if v}
-        return self.queryset.filter(**attrs)
+        lookups = get_purl_query_lookups(purl)
+        return self.queryset.filter(**lookups)
+
+
+def get_purl_query_lookups(purl):
```
tdruez, line 217: Not a great location for this function. The model or around the QuerySet would be better.

vulnerabilities/models.py
```diff
@@ -218,6 +218,15 @@ def with_vulnerability_counts(self):
         )
 
 
+def get_purl_query_lookups(purl):
```
tdruez, line 221: Dostring would be nice, with some explanations about why qualifiers and subpath are not included.

tdruez (changes requested):
Missing changelog entry.

PR #984, opened 2022-10-27
fix a bug in postgresql importer

sify21:
I think `pass` on line 83 should be `continue`, or it will cause an exception on line 112 where `cve_id` is referenced but not defined

Signed-off-by: 司芳源 <sify21@163.com>


TG1999:
@sify21 please checkout the postgresql importer after latest migration.

pombredanne:
@sify21 Thanks.... This was fixed separately. You contrition is mucho apperciated!

PR #988, opened 2022-10-29
Enable throttling

TG1999:
Enabled throttling for a basic user: 1000/day
And a staff user can have unlimited access on API

### DIFF
--- a/vulnerabilities/throttling.py
+++ b/vulnerabilities/throttling.py
@@ -13,40 +13,12 @@ from rest_framework.throttling import UserRateThrottle
 User = get_user_model()
 
 
-class ExceptionalUserRateThrottle(UserRateThrottle):
+class StaffUserRateThrottle(UserRateThrottle):
     def allow_request(self, request, view):
         """
-        Give special access to a few special accounts.
-
-        Mirrors code in super class with minor tweaks.
+        Do not apply throttling for superusers and admins.
         """
-        if self.rate is None:
+        if request.user.is_superuser or request.user.is_staff:
             return True
 
-        self.key = self.get_cache_key(request, view)
-        if self.key is None:
-            return True
-
-        self.history = self.cache.get(self.key, [])
-        self.now = self.timer()
-
-        # Adjust if user has special privileges.
-
-        user = User.objects.get(username=request.user.username)
-
-        if user:
-            if user.is_superuser or user.is_staff:
-                # No throttling for superusers or staff.
-                return True
-
-            else:
-                self.num_requests = self.num_requests
-                self.duration = self.duration
-
-        # Drop any requests from the history which have now passed the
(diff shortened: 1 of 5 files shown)


vulnerabilities/throttling.py
```diff
@@ -0,0 +1,52 @@
...
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+from django.contrib.auth import get_user_model
+from rest_framework.throttling import UserRateThrottle
+
+User = get_user_model()
+
+
+class ExceptionalUserRateThrottle(UserRateThrottle):
```
tdruez, line 16: Not a great name. Superusers may not all be "exceptional" :p

vulnerabilities/throttling.py
```diff
@@ -0,0 +1,52 @@
...
+            return True
+
+        self.history = self.cache.get(self.key, [])
+        self.now = self.timer()
+
+        # Adjust if user has special privileges.
+
+        user = User.objects.get(username=request.user.username)
+
+        if user:
```
tdruez, line 37: If that line is reached, there's always a user in the variable. Otherwise the queryset would have raised an DoesNotExist exception.

vulnerabilities/throttling.py
```diff
@@ -0,0 +1,52 @@
...
+
+User = get_user_model()
+
+
+class ExceptionalUserRateThrottle(UserRateThrottle):
+    def allow_request(self, request, view):
+        """
+        Give special access to a few special accounts.
+
+        Mirrors code in super class with minor tweaks.
```
tdruez, line 21: No need to duplicate the code, call super() instead.

vulnerabilities/throttling.py
```diff
@@ -0,0 +1,52 @@
...
+        self.key = self.get_cache_key(request, view)
+        if self.key is None:
+            return True
+
+        self.history = self.cache.get(self.key, [])
+        self.now = self.timer()
+
+        # Adjust if user has special privileges.
+
+        user = User.objects.get(username=request.user.username)
```
tdruez, line 35: Isn't `request.user` already an instance of the user?

vulnerabilities/throttling.py
tdruez: What about:
```
def allow_request(self, request, view):
    """"
    Do not apply throttling for superusers and admins.
    """"
    if request.user.is_superuser or request.user.is_staff:
        return True

    return super().allow_request(request, view)
```

tdruez (changes requested):

vulnerabilities/throttling.py
TG1999: Done!

tdruez (approved):

PR #995, opened 2022-11-07
Added a failing tests to show multiple vulnerabilities are created for single advisory without aliases

TG1999:


vulnerabilities/models.py
```diff
@@ -327,6 +327,14 @@ def get_or_create_from_purl(self, purl: PackageURL):
         package, _ = Package.objects.get_or_create(**purl_fields)
         return package
 
+    def get_from_purl(self, purl: PackageURL):
+        """
+        Return an existing or new Package (created if neeed) given a
```
pombredanne, line 332: "(created if neeed) "?
Are you sure?

vulnerabilities/models.py
```diff
@@ -327,6 +327,14 @@ def get_or_create_from_purl(self, purl: PackageURL):
         package, _ = Package.objects.get_or_create(**purl_fields)
         return package
 
+    def get_from_purl(self, purl: PackageURL):
```
pombredanne, line 330: I wonder if we could do better? may be `for_purl`?

pombredanne (changes requested):

vulnerabilities/models.py
TG1999: You are right, we are not creating here.

vulnerabilities/improve_runner.py
```diff
@@ -218,3 +221,91 @@ def get_or_create_vulnerability_and_aliases(vulnerability_id, alias_names, summa
...
+
+    # This is fallback for cases when we cannot find vulnerability
+    # by references, summary and packages
+    vulnerability = Vulnerability(summary=inference.summary)
+    vulnerability.save()
+
+    return vulnerability
+
+
+# def match_packages(inference, vuln):
```
pombredanne, line 278: Please remove commented code.

pombredanne (changes requested):
Just a nit for now, I am not sure this is ready for review.

vulnerabilities/improve_runner.py
TG1999: Yes this is not ready for review as of now. It's out of scope for now ( this needs model changes, which we will consider after the migration of importers is done )

PR #1012, opened 2022-11-21
Explicitly state app name in TestMigration

JonoYang:
* The code used to determine the name of the django app to perform test migrations on is a bit arcane
    * By explicitly stating the app name, you can move the data migration test file to any location

Signed-off-by: Jono Yang <jyang@nexb.com>


pombredanne (approved):
LGTM. Thanks!

PR #1017, opened 2022-11-25
Make bulk search fast

TG1999:


vulnerabilities/migrations/0034_package_package.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('vulnerabilities', '0033_alter_vulnerabilityseverity_scoring_system'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='package',
```
pombredanne, line 14: This means you are storing the concrete purl, and this makes 100% sense. But then call it purl or pakage_url and replace/override anything we have already.

vulnerabilities/migrations/0034_package_package.py
```diff
@@ -0,0 +1,18 @@
...
+            name='package',
+            field=models.CharField(blank=True, help_text='The Package URL for this package.', max_length=255),
```
pombredanne, line 16: This should not be blank ever once migrated, so you would need a third migration once the data is updated.

vulnerabilities/models.py
```diff
@@ -531,8 +531,19 @@ class Package(PackageURLMixin):
         to="Vulnerability", through="PackageRelatedVulnerability"
     )
 
+    package = models.CharField(
+        max_length=255,
```
pombredanne, line 535: It would have to be indexed

vulnerabilities/models.py
```diff
@@ -531,8 +531,19 @@ class Package(PackageURLMixin):
...
+        blank=True,
+        null=False,
+        help_text="The Package URL for this package.",
+    )
+
     objects = PackageQuerySet.as_manager()
 
+    def save(self, *args, **kwargs):
+        self.package = self.purl
```
pombredanne, line 544: as explained above this would replace entirely the self.purl.

vulnerabilities/migrations/0035_auto_20221125_1257.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 4.0.7 on 2022-11-25 12:57
```
pombredanne, line 1: You need a better migration script name as this is not a generated migration, is it?

vulnerabilities/migrations/0035_auto_20221125_1257.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations
+from packageurl import PackageURL
+
+class Migration(migrations.Migration):
+
+    def save_package(apps, schema_editor):
+        Package = apps.get_model("vulnerabilities", "Package")
+        for package in Package.objects.all():
```
pombredanne, line 10: you need to create a bulk migration, not one at a time

pombredanne (changes requested):
Thanks!
See feedback inline.

TG1999:
@pombredanne the tests where we ignore qualifiers and subpath are failing because we are matching purls directly without removing qualifiers and subpath, what I suggest as a solution is to store `purl_without_qualifiers_subpath` as a different field and remove qualifiers and subpath from incoming purls and match the list like we are doing now.

vulnerabilities/api.py
```diff
@@ -238,40 +238,34 @@ def bulk_search(self, request):
...
+
+        query = Package.objects.filter(package_url__in=purls)
+
+        if not purl_only:
+            return Response(
+                PackageSerializer(query.distinct(), many=True, context={"request": request}).data
+            )
+
+        vulnerable_purls = (
+            query.filter(packagerelatedvulnerability__fix=False).only("package").distinct()
```
tdruez, line 258: I'm wondering why you do not use the same syntax as in the `all` action?
ie: `.vulnerable().only("package_url")`
Are you fetching something different here?

tdruez:
> @pombredanne the tests where we ignore qualifiers and subpath are failing because we are matching purls directly without removing qualifiers and subpath, what I suggest as a solution is to store purl_without_qualifiers_subpath as a different field and remove qualifiers and subpath from incoming purls and match the list like we are doing now.

Yes, you should store both full purl and plain purl and allow to run the matching on the plain purl.

tdruez (approved):

vulnerabilities/models.py
```diff
@@ -531,8 +531,41 @@ class Package(PackageURLMixin):
...
     )
 
+    package_url = models.CharField(
+        max_length=1000,
+        null=False,
+        help_text="The Package URL for this package.",
+        db_index=True,
+    )
+
+    plain_package_url = models.CharField(
```
pombredanne, line 541: I am not in love with the `plain_package_url` name but this makes sense for now. We will need to find a better way.
I will merge for now!

pombredanne (approved):
Thanks! This will need to be refined name wise but this is a good fix for now.
Merging

PR #1021, opened 2022-11-30
Migrate kaybee importer #1011

TG1999:


vulnerabilities/importers/kaybee.py
```diff
@@ -7,56 +7,98 @@
...
+                affected_version_range=affected_version_range,
+                fixed_version=fixed_version,
+            )
+        )
 
-    for fix in data.get("fixes", []):
-        for commit in fix["commits"]:
+    for fix in data.get("fixes") or []:
+        for commit in fix.get("commits") or []:
             references.append(Reference(url=f"{commit['repository']}/{commit['id']}"))
```
pombredanne, line 97: Would there be a way to state that this is fix commit explicitly?

pombredanne (changes requested):
Thanks! Iwould like to make sure we have a proper way to identify a fix commit reference

TG1999:
Closing this PR for now, will re-open once models are open for storing commits

Issue #1032, opened 2022-12-07
Package search does not handle Package "name@version"


mjherzog:
Using the current (22.12.06) VCIO v31.0.0, I am not able to Package Search on name@version. Some examples:
- Search on "pkg:pypi/jinja2@2.11.3" returns one record
   Search on "pypi/jinja2@2.11.3" returns none
   Search on "jinja2@2.11.3" returns none
- Search on "pkg:rpm/redhat/expat@2.1.0-14" returns one record
   Search on "pkg:rpm/redhat/expat@2.1.0" returns none
   Search on "rpm/redhat/expat@2.1.0-14" returns none
   Search on "expat@2.1.0-14" returns none

The documentation for the Package Search currently says: "Search for vulnerable packages by Package URL (aka. purl) such as pkg:maven/org.apache.logging.log4j/log4j@2.0 or purl prefix fragment such as pkg:alpine or by package name." I interpreted these to be examples not a list of 3 specific options. 

There are two aspects to my use case:
1. I would like to enter name@version without entering the "pkg:" or "pkg/type/' prefix for ease of use
2. In some cases I would like to look up the same package name@version across different package types - e.g. alpine, deb and redhat.

We need to either enhance Package Search to handle more types of purl "fragments" or update the documentation to specify the syntax options.

TG1999:
This has been fixed in the latest release please try the latest release and let us know.

PR #1033, opened 2022-12-07
Handle purl fragments in package search #1032

TG1999:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,9 @@ Version v31.1.0
 
 - We re-enabled support for the NPM vulnerabilities advisories importer. 
 - We re-enabled support for the Retiredotnet vulnerabilities advisories importer.
-- We are now handling purl fragments in package search.
+- We are now handling purl fragments in package search. For example:
+  you can now serch using queries in the UI like this : `cherrypy@2.1.1`,
+  `cherrypy` or `pkg:pypi`.
 
 
 Version v31.0.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Version v31.1.0
 
 - We re-enabled support for the NPM vulnerabilities advisories importer. 
 - We re-enabled support for the Retiredotnet vulnerabilities advisories importer.
+- We are now handling purl fragments in package search.
 
 
 Version v31.0.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@ Release notes
 =============
 
 
+Version v31.1.0
+----------------
+
+- We are now handling purl fragments in package search.
```
pombredanne, line 8: Please elaborate on what this means for a user

vulnerabilities/models.py
```diff
@@ -412,61 +412,22 @@ def search(self, query=None):
...
-                qs = qs.filter(version__iexact=version)
-
-        return qs
+
+        try:
+            # if it's a valid purl, use it as is
+            purl = PackageURL.from_string(query)
+            return self.for_purl(purl, with_qualifiers_and_subpath=False)
+        except ValueError:
+            return qs.filter(package_url__icontains=query)
```
pombredanne, line 422: Are you sure this provides the same kind of search results as before?

vulnerabilities/models.py
```diff
@@ -412,61 +412,22 @@ def search(self, query=None):
...
 
     def for_purl(self, purl, with_qualifiers_and_subpath=True):
         """
         Return a queryset matching the ``purl`` Package URL.
         """
         if not isinstance(purl, PackageURL):
             purl = PackageURL.from_string(purl)
-        purl = purl.to_dict()
+        purl = purl_to_dict(purl)
```
pombredanne, line 430: What is purl_to_dict() doing?

vulnerabilities/tests/test_view.py
```diff
@@ -57,6 +57,12 @@ def test_package_view(self):
         self.assertEqual(len(pkgs), 1)
         self.assertEqual(pkgs[0].purl, "pkg:nginx/nginx@1.0.15")
 
+    def test_package_view_with_purl_fragment(self):
```
pombredanne, line 60: We need many more tests IMHO to assert the search results quality before and ensure this does not regress.

pombredanne (changes requested):
Thanks. Some nits for your review.

vulnerabilities/tests/test_view.py
TG1999: @pombredanne we already had tests for the previous implementation and none of them fail, please tell me any particular corner case that should we check for ?

vulnerabilities/models.py
TG1999: This returns a dict of purl components for use in the queryset.

pombredanne (approved):
LGTM! Thanks.

PR #1075, opened 2023-01-10
use public VulnerableCode instance in VulnTotal 

keshav-space:
requires VCIO_TOKEN in .env file

Signed-off-by: Keshav Priyadarshi <git@keshav.space>

### DIFF
--- a/vulntotal/datasources/vulnerablecode.py
+++ b/vulntotal/datasources/vulnerablecode.py
@@ -8,10 +8,12 @@
 #
 
 import logging
+import os
 from typing import Iterable
 from urllib.parse import urljoin
 
 import requests
+from dotenv import load_dotenv
 from packageurl import PackageURL
 
 from vulntotal.validator import DataSource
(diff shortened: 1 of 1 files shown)


vulntotal/datasources/vulnerablecode.py
```diff
@@ -85,3 +85,26 @@ def parse_advisory(fetched_advisory) -> VendorData:
...
         aliases=aliases, affected_versions=affected_versions, fixed_versions=fixed_versions
     )
+
+
+class VCIOTokenError(Exception):
+    pass
+
+
+def fetch_vulnerablecode_query(url: str, payload: dict):
+    load_dotenv()
```
TG1999, line 95: Any particular reason to use `load_dotenv()` here, we have accessed the environment variables without this as well in past for example here https://github.com/nexB/vulnerablecode/blob/main/vulnerabilities/utils.py#L238 ?

vulntotal/datasources/vulnerablecode.py
```diff
@@ -85,3 +85,26 @@ def parse_advisory(fetched_advisory) -> VendorData:
...
+    vcio_token = os.environ.get("VCIO_TOKEN", None)
+    if not vcio_token:
+        msg = "Cannot call VulnerableCode API without a token set in the VCIO_TOKEN environment variable."
+        raise VCIOTokenError(msg)
+
+    response = (
+        requests.post(url, headers={"Authorization": vcio_token}, json=payload)
```
TG1999, line 102: Add a comment showing what is expected in token in .env , AFAIK either the user can store just the token and you have to do something like this in that case `headers={"Authorization": "Token f{vcio_token}"}` or add a comment explaining the env should have something like this : `Token <YOUR_VCIO_TOKEN>`

TG1999 (changes requested):
Thakns!
minor nitpickings for your review

vulntotal/datasources/vulnerablecode.py
keshav-space: VulnTotal is not a Django app, it cannot access the env variable loaded by Django. Hence the need to use `load_dotenv()`

TG1999 (approved):
LGTM! Thanks

PR #1087, opened 2023-01-20
Add migrations to remove corrupted advisories

TG1999:


vulnerabilities/tests/test_data_migrations.py
```diff
@@ -533,3 +533,38 @@ def test_merge_rows(self):
         ]
 
         assert severities == expected
+
+
+class RemoveCorrupteAdvisories(TestMigrations):
+    app_name = "vulnerabilities"
+    migrate_from = "0036_alter_package_package_url_and_more"
+    migrate_to = "0037_auto_20230120_1204"
```
pombredanne, line 541: Can we give this a better name?
What about renaming the file to `0037_remove_corrupted_advisories_with_incorrect_refs_and_severity

pombredanne (approved):
LGTM. There is only a small nit picking for your consideration.

pombredanne:
You also need to renumber your migration to 038

vulnerabilities/tests/test_data_migrations.py
TG1999: Thanks! done

Issue #1126, opened 2023-02-12
Run latest release of Black


johnmhoran:
See https://github.com/psf/black/blob/main/CHANGES.md#highlights-1 ("This is the first release of 2023, and following our [stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy), it comes with a number of improvements to our stable style, including improvements to empty line handling . . . .")

PR #1085, opened 2023-01-19
Create SUSE OVAL importer

johnmhoran:
Implements #1079

PR #1127, opened 2023-02-12
Run latest release of Black

johnmhoran:
Reference: https://github.com/nexB/vulnerablecode/issues/1126

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -97,7 +97,6 @@ class VulnSerializerRefsAndSummary(serializers.HyperlinkedModelSerializer):
 
 
 class VulnerabilitySerializer(serializers.HyperlinkedModelSerializer):
-
     fixed_packages = MinimalPackageSerializer(
         many=True, source="filtered_fixed_packages", read_only=True
     )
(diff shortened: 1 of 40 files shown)


johnmhoran:
@pombredanne @TG1999 It looks like we have a `BaseTests.test_codestyle` failure here because our test suite uses Black 22.3.0

```
Collecting black>=22.3.0
  Downloading black-22.3.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.5 MB)
```

while I have Black 23.1.0 (released 2023-02-04 -- see https://github.com/psf/black/blob/main/CHANGES.md) in my local vulnerablecode `venv`, which produces a similar failure when I run `make test` (and `make valid` updates 40 files locally).  Can we update Black in our GH testing to 23.1.0?

johnmhoran:
This PR remains open, awaiting guidance/decision re how we use Black and `make valid` and deal with the line-spacing issue.  Perhaps that has been resolved, but to my knowledge that's not the case.  See also https://github.com/nexB/vulnerablecode/pull/1085.

pombredanne:
We now use a pinned version so this should no longer be an issue https://github.com/nexB/vulnerablecode/blob/0d9c9b536a5c6281b92bf7b816f8904a7c499632/requirements.txt#L12

johnmhoran:
Thanks.  Note that this has been awaiting a response since February.  I'll take a look when time permits.

TG1999:
We are using the version 23.12.0 of `black` and it's from skeleton.

PR #1149, opened 2023-03-10
Fix github importer

TG1999:


TG1999:
Trivial change merging this

PR #1155, opened 2023-03-21
Add support for conan advisories

TG1999:


vulnerabilities/package_managers.py
```diff
@@ -556,6 +556,22 @@ def fetch(self, pkg: str) -> Iterable[PackageVersion]:
             )
 
 
+class ConanVersionAPI(VersionAPI):
+    """
+    Fetch versions of Erlang packages from the hex API
```
pombredanne, line 561: Erlang and hex?

pombredanne (changes requested):
Thanks... See a nit inlined and you need to add a CHANGELOG entry too!

vulnerabilities/package_managers.py
TG1999: Done!

pombredanne (approved):
LGTM! Thanks.

Issue #1148, opened 2023-03-09
Update documentation for v32


DennisClark:
Review existing documentation for correctness 
* installation instructions
* usage instructions
* API usage information 

make sure the Changelog is complete
are there any Upgrade instructions

TG1999:
User-
- Using the UI and API
- Explain how to use API via openAPI documentation
- A Table to show endpoint what to query it with and what you can expect from the output. 
- Explain WebUI
  - Search by vulnerabilities and packages
  - PURL or fragment of PURL
  - Drill on details and all the thing shown on the package details page.

Developer - 
Move importer and improver to CONTRIBUTING section

Local Instance - 
- Installation
- API usage

PR #1169, opened 2023-04-03
Add documentation for v32.0.0

TG1999:
We want to make this documentation friendly for a totally new user who is going to use vulnerablecode for the first time. As discussed here https://github.com/nexB/vulnerablecode/issues/1148#issuecomment-1491677552


docs/source/api.rst
```diff
@@ -48,3 +59,153 @@ There are two primary endpoints:
...
+the page parameter.
+
+.. list-table:: Table for the API endpoints
+   :widths: 30 40 30
+   :header-rows: 1
+
+   * - Endpoint
+     - Query Parameters
+     - Expected Output
+   * - /api/vulnerabilities/
```
AyanSinhaMahapatra, line 73: `/api/vulnerabilities/` similarly and for all occurances below

docs/source/api.rst
```diff
@@ -48,3 +59,153 @@ There are two primary endpoints:
...
+       - page (integer) = page number of the response
+       - page_size (integer) = number of vulnerabilities in each page
+     - Returns a list of vulnerabilities
+   * - /api/vulnerabilities/{id}
+     -
+       - id = internal primary id of the vulnerability
+     - Returns a vulnerability with the given id
+   * - /api/packages
+     -
+       - type (string) = type of the package
```
AyanSinhaMahapatra, line 85: these variables should also be like `type`

docs/source/api.rst
```diff
@@ -48,3 +59,153 @@ There are two primary endpoints:
...
+
+
+.. _Package Bulk Search:
+
+Package Bulk Search
+---------------------
+
+
+The package bulk search endpoint allows you to search for packages in bulk. You can
+pass a list of packages in the request body and the endpoint will return a list of
```
AyanSinhaMahapatra, line 136: packages -> purls ?

docs/source/tutorial_add_new_importer.rst
```diff
@@ -1,301 +0,0 @@
-.. _tutorial_add_a_new_importer:
-
```
AyanSinhaMahapatra, line 2: why are we deleting these?

docs/source/user-interface.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The search by vulnerabilities is a very powerful feature of the
+VulnerableCode. It allows you to search for vulnerabilities by the
+VCID itself. It also allows you to search for
+vulnerabilities by the CVE, GHSA, CPEs etc or by the
+fragment of these indentifiers like ``CVE-2021``.
+
+The search by vulnerabilities is available at the following URL:
+
+    https://public.vulnerablecode.io/vulnerabilities/search
```
AyanSinhaMahapatra, line 19: also have this hyperlinked (while keeping the text showed same)

AyanSinhaMahapatra (changes requested):
@TG1999 looking good, thanks! A couple of nits for your consideration here.

docs/source/api.rst
johnmhoran, line 1: docs/source/api.rst

Line 144: Change `recieved` to `received`.

Line 145:  In `allows you to query API` add `the` before `API`.

Lines 172-173:  In the passage

The response will be a list of packages with vulnerabilities
they are affected by or fixing any vulnerability.

it's not clear what we mean by `they are affected by or fixing any vulnerability`.  

Do we mean something like this?  `-- these are packages that are affected by and/or that fix a vulnerability.`

docs/source/contributing.rst
johnmhoran, line 1: docs/source/contributing.rst

Line 177: "All importers are located in the :file:`vulnerabilites/importers` package."  Do we mean `directory` rather than `package`?  Or does the presence of `__init__.py` lead us to use the term `package` instead?

Line 232: Change `targetting` to `targeting`.

Line 314: Use ``make valid`` to format your new code using black and isort automatically.

Don't we have a problem with the most recent version of Black changing certain types of line spacing, e.g., immediately after a function definition?  If so, how (if at all) should we document our resolution of that problem?

Line 315: Use ``make check`` to check for formatting errrors.

Change `errrors` to `errors`.

docs/source/user-interface.rst
johnmhoran, line 1: docs/source/user-interface.rst

Line 11: In `very powerful feature of the` delete `the`.

Line 15: Change `indentifiers` to `identifiers`.

Line 51: In `very powerful feature of the` delete `the`.

docs/source/api.rst
```diff
@@ -48,3 +59,153 @@ There are two primary endpoints:
 And two secondary endpoints, used to query vulnerability aliases (such as CVEs)
 and vulnerability by CPEs: cpes/ and aliases/
 
+The API is paginated and the default page size is 100. You can change the page size
+by passing the page_size parameter. You can also change the page number by passing
```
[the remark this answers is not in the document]
TG1999: Done!

docs/source/api.rst
TG1999: Done!

docs/source/api.rst
TG1999: Done!

docs/source/api.rst
TG1999: Done!

docs/source/api.rst
TG1999: Done!

docs/source/api.rst
```diff
@@ -11,6 +11,17 @@ Browse the Open API documentation
...
+--------------------------------------
+
+The API documentation is available at https://public.vulnerablecode.io/api/docs/.
+To use the endpoints you need to authenticate with an API key. Request your API key
+from https://public.vulnerablecode.io/account/request_api_key/. Once you have
+your API key, click on the ``Authorize`` button on the top right of the page and enter
+your API key in the ``value`` field with ``Token`` prefix, so if your token is "1234567890abcdef"
+then you have to enter this: ``Token 1234567890abcdef``.
+
+
```
pombredanne, line 24: The lines below "Enable the API key authentication" and "Create an API key-only user" are irrelevant for end users as they do not have access to this. These should be moved to a new section related to API usage administration for on-premise deployments, either at the bottom, or a completely new and different page

docs/source/api.rst
```diff
@@ -34,10 +45,121 @@ Access the API using curl
 
     curl -X GET -H 'Authorization: Token <YOUR TOKEN>' https://public.vulnerablecode.io/api/
```
pombredanne, line 46: IMHO this should be part of the examples, as an alternative to using Python. Here this is not the best place.

pombredanne (changes requested):
Some more feedback!

PR #1177, opened 2023-04-11
Dark mode toggle added for documentation.

swastkk:
Added the Dark Mode toggle in the documentation.
Reference - https://github.com/MrDogeBro/sphinx_rtd_dark_mode

With light mode-
Screenshot_2023-04-11_18-32-40

With Dark Mode-
Screenshot_2023-04-11_18-32-58

PR #1178, opened 2023-04-11
Add dark mode toggle in documentation.

swastkk:
Reference- https://github.com/nexB/vulnerablecode/pull/1177

Signed-off-by - swastik <swastkk@gmail.com>


swastkk:
@TG1999 can merge this PR instead .

docs/source/conf.py
```diff
@@ -63,3 +63,7 @@
 html_static_path = []
 
 master_doc = "index"
+
+# user starts in light mode
+
+default_dark_mode = False
```
AyanSinhaMahapatra, line 69: Yes, this is what we want, light stays as default, as that is the case without this plugin.

AyanSinhaMahapatra (approved):
LGTM! Thanks! I wonder though whether we should add the same to https://github.com/nexB/skeleton, we should have this everywhere.

docs/source/conf.py
swastkk: > Yes, this is what we want, light stays as default, as that is the case without this plugin.

Yep ✨

swastkk:
Will add this in the skeleton as soon as possible ✨🚀

TG1999:
@swastkk I am closing this PR in favour of https://github.com/nexB/skeleton/commit/5ab9b3a15e0095a8186ca3870cb9f9eade83833d, thanks again for raising this PR

PR #1209, opened 2023-06-04
purl-sync project

ziadhany:


purl_sync/review/views.py
```diff
@@ -0,0 +1,616 @@
...
+        headers={"content-type": "application/x-www-form-urlencoded"},
+        data={
+            "grant_type": "password",
+            "username": payload["username"],
+            "password": payload["password"],
+            "client_id": CLIENT_ID,
+            "client_secret": CLIENT_SECRET,
+        },
+    )
+    return JsonResponse(json.loads(r.content), status=r.status_code, content_type=AP_CONTENT_TYPE)
```
ziadhany, line 585: @TG1999 @pombredanne Is there any better way to pass client_id and client_secret instead of duplicate the requests again .

ziadhany:
Moved to a new Git repository

PR #1222, opened 2023-07-05
Drop duplicated fields from API response and add missing quotes for href values in template

TG1999:


tdruez (approved):

TG1999:
follow up issue https://github.com/nexB/vulnerablecode/issues/1223

PR #1232, opened 2023-07-12
Mark advisories status according to NVD advisory

### LINKED ISSUE #1281: Dealing with junk CVEs
See https://daniel.haxx.se/blog/2023/08/26/cve-2020-19909-is-everything-that-is-wrong-with-cves/
We need to find a way to ensure we can filter and tag junk CVEs like this quickly, especially on popular packages like cURL where the junk will be a cause of major churn.

TG1999:
Fixes #1281


vulnerabilities/models.py
```diff
@@ -152,6 +148,14 @@ def with_package_counts(self):
         )
 
 
+class VulnerabilityStatusType(models.IntegerChoices):
+    """List of vulnerability statuses."""
+
+    PUBLISHED = 1, "published"
```
pombredanne, line 154: The code in https://github.com/nexB/vulnerablecode/pull/1232/files#diff-3e8fc96d993d0bd8585642f68239732a30a19b7ec100a36ee5d21e209625ff81R120 seem redundant. Could we have a single label, either as "published" or "Published" but not both?

vulnerabilities/views.py
```diff
@@ -115,12 +115,21 @@ class VulnerabilityDetails(DetailView):
     def get_queryset(self):
         return super().get_queryset().prefetch_related("references", "aliases", "weaknesses")
 
+    def get_status(self, status):
```
pombredanne, line 118: This method should be moved to VulnerabilityStatusType IMHO

pombredanne (approved):
I added some nits... please merge when you feel comfy!

PR #1245, opened 2023-07-24
Add weakness in unique content ID in advisories

TG1999:
The NVD importer is failing to import due to this problem, since all other data is identical except weaknesses which is a new field on advisory, NVD importer is not able to create advisories.

### DIFF
--- a/vulnerabilities/import_runner.py
+++ b/vulnerabilities/import_runner.py
@@ -54,19 +54,23 @@ def process_advisories(advisory_datas: Iterable[AdvisoryData], importer_name: st
     for data in advisory_datas:
         # https://nvd.nist.gov/vuln/detail/CVE-2013-4314
         # https://github.com/cms-dev/cms/issues/888#issuecomment-516977572
-        data.summary = data.summary.replace("\x00", "\uFFFD")
-        obj, created = Advisory.objects.get_or_create(
-            aliases=data.aliases,
-            summary=data.summary,
-            affected_packages=[pkg.to_dict() for pkg in data.affected_packages],
-            references=[ref.to_dict() for ref in data.references],
-            date_published=data.date_published,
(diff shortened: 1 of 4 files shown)


vulnerabilities/import_runner.py
```diff
@@ -54,19 +54,23 @@ def process_advisories(advisory_datas: Iterable[AdvisoryData], importer_name: st
...
+                references=[ref.to_dict() for ref in data.references],
+                date_published=data.date_published,
+                weaknesses=data.weaknesses,
+                defaults={
+                    "created_by": importer_name,
+                    "date_collected": datetime.datetime.now(tz=datetime.timezone.utc),
+                },
+            )
+        except Exception as e:
+            logger.error(f"Error while processing {data!r} with aliases {data.aliases!r}: {e}")
```
pombredanne, line 72: Are you getting every data you need to debug with this message?

vulnerabilities/models.py
```diff
@@ -836,7 +836,7 @@ class Meta:
 
     def save(self, *args, **kwargs):
         checksum = hashlib.md5()
-        for field in (self.summary, self.affected_packages, self.references):
+        for field in (self.summary, self.affected_packages, self.references, self.weaknesses):
```
pombredanne, line 839: Why not dumping only once?
```
key_data = [self.summary, self.affected_packages, self.references, self.weaknesses]
dumped = json.dumps(key_data, separators=(",", ":")).encode("utf-8")
checksum = hashlib.md5(dumped)
```

pombredanne (approved):
Thanks... some minor nits for your consideration! You can apply or ignore.

vulnerabilities/models.py
TG1999: This will change the checksum IMO, we should do it separately as a migration first and then change the function on models.

PR #1250, opened 2023-08-01
Add vulnerability only option on package API endpoint

TG1999:


vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
     Lookup software package using Package URLs
     """
 
+    def to_representation(self, instance):
```
tdruez, line 127: The code would benefit from using the line spaces to enhance readability.

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        data = super().to_representation(instance)
+        request = self.context["request"]
+        if not hasattr(request, "data"):
+            return data
+        vulnerabilities_only = request.data.get("vulnerabilities_only", False)
+        if not vulnerabilities_only:
+            return data
+        print(data["affected_by_vulnerabilities"])
```
tdruez, line 135: Leftover statement.

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        vuln_only_data = {}
+        vuln_only_data["affected_by_vulnerabilities"] = [
+            vulnerability for vulnerability in data["affected_by_vulnerabilities"]
```
tdruez, line 138: This appears to have no effect.

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        ]
+        vuln_only_data["affected_by_vulnerabilities"] = self.get_unique_vulnerabilities(
+            vuln_only_data["affected_by_vulnerabilities"]
+        )
+        vuln_only_data["fixing_vulnerabilities"] = [
+            vulnerability for vulnerability in data["fixing_vulnerabilities"]
```
tdruez, line 144: This appears to have no effect.

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        ]
+        vuln_only_data["fixing_vulnerabilities"] = self.get_unique_vulnerabilities(
+            vuln_only_data["fixing_vulnerabilities"]
+        )
+        return vuln_only_data
```
tdruez, line 149: Suggestions:
```
affected_by = data["affected_by_vulnerabilities"]
fixing = data["fixing_vulnerabilities"]
        
return {
    "affected_by_vulnerabilities": self.get_unique_vulnerabilities(affected_by),
    "fixing_vulnerabilities": self.get_unique_vulnerabilities(fixing),
}
```

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        if not hasattr(request, "data"):
+            return data
+        vulnerabilities_only = request.data.get("vulnerabilities_only", False)
+        if not vulnerabilities_only:
+            return data
+        print(data["affected_by_vulnerabilities"])
+        vuln_only_data = {}
+        vuln_only_data["affected_by_vulnerabilities"] = [
+            vulnerability for vulnerability in data["affected_by_vulnerabilities"]
+        ]
```
tdruez, line 139: This appears to have no effect.

vulnerabilities/api.py
```diff
@@ -124,6 +124,34 @@ class PackageSerializer(serializers.HyperlinkedModelSerializer):
...
+        )
+        vuln_only_data["fixing_vulnerabilities"] = [
+            vulnerability for vulnerability in data["fixing_vulnerabilities"]
+        ]
+        vuln_only_data["fixing_vulnerabilities"] = self.get_unique_vulnerabilities(
+            vuln_only_data["fixing_vulnerabilities"]
+        )
+        return vuln_only_data
+
+    def get_unique_vulnerabilities(self, vuln_list):
```
tdruez, line 151: Method 'get_unique_vulnerabilities' may be 'static'

tdruez (changes requested):
@TG1999 See the comments.
Also, why not use a proper stripped-down (including only the required fields) `PackageSerializer` class to handle this instead of this data manipulation?

vulnerabilities/api.py
```diff
@@ -272,12 +268,18 @@ def bulk_search(self, request):
         purls = request.data.get("purls", []) or []
         purl_only = request.data.get("purl_only", False)
         plain_purl = request.data.get("plain_purl", False)
+        vulnerabilities_only = request.data.get("vulnerabilities_only", False)
```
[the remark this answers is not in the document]
TG1999: @tdruez what will you like to suggest here? how should we take the request data  input here ?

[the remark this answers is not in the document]
tdruez: My point is that instead of having 1 URL with many exceptions and variations based on the request data, shouldn't we have different endpoints? I'm not saying that this is the best way, I'm just wondering how we can avoid to much complexity within a single endpoint.

TG1999:
@tdruez added a sample test.

tdruez (approved):

PR #1256, opened 2023-08-02
Remove weaknesses from view which are not in DB

TG1999:
The problem with the CWE data is that there is no error handling when we try to access weakness using this ``weakness = self.db.get(self.cwe_id)``. For say CWE-189 which is a weakness associated with CVE-2009-3389 is not present in cwe2.database.

### DIFF
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -276,16 +276,26 @@ class Weakness(models.Model):
     db = Database()
 
     @property
+    def weakness(self):
+        """
+        Return a queryset of Weakness for this vulnerability.
+        """
+        try:
+            weakness = self.db.get(self.cwe_id)
+            return weakness
+        except Exception as e:
+            logger.warning(f"Could not find CWE {self.cwe_id}: {e}")
+            return None
+
+    @property
     def name(self):
         """Return the weakness's name."""
-        weakness = self.db.get(self.cwe_id)
-        return weakness.name
+        return self.weakness.name if self.weakness else ""
 
     @property
     def description(self):
         """Return the weakness's description."""
-        weakness = self.db.get(self.cwe_id)
-        return weakness.description
+        return self.weakness.description if self.weakness else ""
 
 
 class VulnerabilityReferenceQuerySet(BaseQuerySet):
--- a/vulnerabilities/models.py
+++ b/vulnerabilities/models.py
@@ -285,7 +285,6 @@ class Weakness(models.Model):
             return weakness
         except Exception as e:
             logger.warning(f"Could not find CWE {self.cwe_id}: {e}")
-            return None
 
     @property
     def name(self):
--- a/vulnerabilities/views.py
+++ b/vulnerabilities/views.py
@@ -117,7 +117,9 @@ class VulnerabilityDetails(DetailView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         weaknesses = self.object.weaknesses.all()
-        weaknesses_present_in_db = [weakness for weakness in weaknesses if weakness.weakness]
+        weaknesses_present_in_db = [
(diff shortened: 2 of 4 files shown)


vulnerabilities/models.py
```diff
@@ -275,17 +275,27 @@ class Weakness(models.Model):
...
+    def weakness(self):
+        """
+        Return a queryset of Weakness for this vulnerability.
+        """
+        try:
+            weakness = self.db.get(self.cwe_id)
+            return weakness
+        except Exception as e:
+            logger.warning(f"Could not find CWE {self.cwe_id}: {e}")
+            return None
```
tdruez, line 288: `return None` is equivalent to `return`.
Also, no need to return here since it's the last line of the function.

vulnerabilities/models.py
```diff
@@ -275,17 +275,27 @@ class Weakness(models.Model):
...
 
+    @property
+    def weakness(self):
+        """
+        Return a queryset of Weakness for this vulnerability.
+        """
+        try:
+            weakness = self.db.get(self.cwe_id)
+            return weakness
+        except Exception as e:
```
tdruez, line 286: Avoid too broad `Exception`.

vulnerabilities/views.py
```diff
@@ -116,6 +116,8 @@ def get_queryset(self):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        weaknesses = self.object.weaknesses.all()
+        weaknesses_present_in_db = [weakness for weakness in weaknesses if weakness.weakness]
```
tdruez, line 120: `[weakness for weakness in weaknesses if weakness.weakness]`
This naming is not great.

tdruez (changes requested):

vulnerabilities/models.py
TG1999: @tdruez , the exception raised by cwe2 is this:
```
55             break
     57 if not cwe_obj:
---> 58     raise Exception(f"Invalid CWE ID {cwe_id} - {category or ''}")
     60 return Weakness(*cwe_obj)
```
What should be the best solution here?

tdruez: @TG1999  In that case, you're right, we don't have a choice.
It's a bad design on the cwe2 side.

vulnerabilities/models.py
TG1999: done

vulnerabilities/views.py
TG1999: done

tdruez: I was trying to point out a design issue. We can change variable names, but we're simply hiding the underlying issues.

tdruez (approved):

vulnerabilities/views.py
TG1999: @tdruez In that case, we first need a migration, that will remove the CWEs that are stored in advisories and CWE table. Then overwrite save, to verify if a CWE exists on CWE2 side before storing it into DB again.

TG1999: @ziadhany FYI ^^

vulnerabilities/models.py
ziadhany: > @TG1999 In that case, you're right, we don't have a choice. It's a bad design on the cwe2 side.

@tdruez How can I improve this design ?

vulnerabilities/views.py
ziadhany: @TG1999 
According to cwe.mitre.org:  [CWE-189](https://cwe.mitre.org/data/definitions/189.html)  is CWE CATEGORY  .
`Vulnerability Mapping Notes : (this CWE ID must not be used to map to real-world vulnerabilities)`
so I think we should remove all CWE that have this Note .

Issue #1263, opened 2023-08-08
Add a sticky div to the 'Package details' page that will keep the name of the Package visible when the user scrolls


johnmhoran:
The title says it all.  I already wrote the HTML and CSS code for this earlier today -- for an initial draft, at least -- and will start with that when I tackle this issue.

One question: do we want or need a non-white background?  A thick bottom border or a bottom box-shadow will make the div more visible when it rests at the top of the page, but we will also have two more rows below it, newly added in my work on issue #1228, which I had contemplated displaying as a 3-row table -- we might have a design/display conflict to resolve when we reach this point.

pombredanne:
We could either make the "purl" row sticky... our we could also add the purl to the blue "package details" bar and make this bar sticky?
Would you have some screenshot to paste as I reckon you made some prototype?

johnmhoran:
Thanks @pombredanne and good idea.  Here are several screenshots of the current prototype.  Please treat this as just a working draft intended to prompt comments and ideas.  

This one combines the sticky div (the top row with `purl pkg:npm/sequelize@3.19.3`) -- I changed the yellow background to white for discussion purposes -- with the two new rows I created for the closest non-vuln and the latest non-vuln.  (I agree that it does look cleaner with a white background. ;-))

This is the page at rest (before scrolling):  
![image](https://github.com/nexB/vulnerablecode/assets/11096678/0e348ced-fc5c-45be-91b1-7ccfde80d84b)



. . . and after a little scrolling:  
![image](https://github.com/nexB/vulnerablecode/assets/11096678/b0e250fc-175c-493a-b1ec-1ff64424616b)

johnmhoran:
Since there have been no further comments in the past month, let's see if we can put this to bed.  I'll open a PR now.

johnmhoran:
I spoke too soon about opening a PR -- I'd done the exploratory work in my 1228 branch I think, so I'll either open a new branch for this issue or tackle it in the 1228 branch.  @pombredanne Do you have a preference?

pombredanne:
@johnmhoran my preference is for smaller focused PRs as they are easier to merge

johnmhoran:
Thanks @pombredanne -- I'll handle the sticky div with a separate branch and a separate PR.

PR #1294, opened 2023-09-07
Add sticky div to Package details template

johnmhoran:
Reference: https://github.com/nexB/vulnerablecode/issues/1263

### DIFF
--- a/vulnerabilities/templates/package_details.html
+++ b/vulnerabilities/templates/package_details.html
@@ -20,11 +20,11 @@ VulnerableCode Package Details - {{ package.purl }}
                 </div>
             </article>
 
-            <div class="pl-3 pr-3 mb-5">
+            <div class="pl-3 pr-3 mb-5 floating-purl">
                 <table class="table vcio-table width-100-pct mt-2">
                     <tbody>
                         <tr>
-                            <td class="two-col-left">
+                            <td class="two-col-left" style="width: 175px;">
                                 <span
                                     class="has-tooltip-multiline has-tooltip-black has-tooltip-arrow has-tooltip-text-left"
                                     data-tooltip="The package url or purl is a URL string used to identify and locate a software package."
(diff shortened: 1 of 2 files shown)


TG1999:
@johnmhoran thanks ++, can we have some before and after screengrabs for changes done by this PR ?

johnmhoran:
@TG1999 Here are before and after screenshots.  I'll also update `main` and merge into my 1263 branch and push that shortly.

Before at top:

image

Before when scrolled:

image

After at top:

image

After when scrolled:

image

TG1999:
@johnmhoran LGTM! I was thinking if we could use `rem` instead of `px` https://austingil.com/px-or-rem-in-css/#:~:text=Pixels%20are%20an%20absolute%20unit,the%20equivalent%20of%2024%20pixels.

johnmhoran:
@TG1999 Well, with `1263-package-details-sticky-div` checked out, in `custom.css` we have 96 instances of `px` being used, and 5 instances of `px` for widths in the template `package_details.html`.

I could take the time to calculate the `rem` equivalent (no idea how straightforward that process is), make the changes, then test the UI to check for any odd results -- I gather `rem` is relative rather than absolute so there likely will be visual changes as a result.  Not sure that's a good use of my time right now as I'm trying to finally put my 1228 branch to bed but will turn to the px-rem work if that's what you and @pombredanne want me to focus on now.  Please let me know.

johnmhoran:
Presumably I'd then also need to make the same sorts of changes in my 1228 branch, to the extent not covered in the sticky div branch, right?

johnmhoran:
@TG1999 Looking more broadly in the files, there seem to be several thousand `px` references.  Change all of them to `rem`?

TG1999:
@johnmhoran leave px to rem for now, it can be addressed separately

TG1999 (approved):
LGTM!

johnmhoran:
@TG1999 OK.  But if we need to convert from `px` to `rem` then I do want to correct my use of `px`.  Is there a relatively easy and sane way to do that conversion?

PR #1307, opened 2023-09-22
Add pagination to valid versions improver

TG1999:
We are having the same problem that we had with the default improver, the volume of advisories is too high and it eats up a lot of RAM, by paginating the advisories we will be able to save a lot of RAM.


TG1999:
Trivial change merging this

PR #1337, opened 2023-11-10
Add RTD build configuration from skeleton

AyanSinhaMahapatra:
RTD builds were failing as we were not using the `build.os` field which is required.
This adds the RTD build conf from skeleton to fix the issue.


TG1999:
Thanks++ @AyanSinhaMahapatra

TG1999 (approved):
LGTM!

Issue #977, opened 2022-10-20
Add documentation for VC database fields


mjherzog:
We need to add Reference documentation for the VulnerableCode database fields. Perhaps we can regenerate the diagrams that @johnmhoran created about a month ago (copies attached). We may need to figure out how to create them in a narrower width format because is it annoying to scroll for this type of diagram - see https://vulnerablecode.readthedocs.io/en/latest/reference_framework_overview.html.
This may be the time to widen our RTD page format.

mjherzog:
![VC_DB_core_diagram](https://user-images.githubusercontent.com/4853767/197075504-d4e432c6-4e67-4ad5-98c4-ee47fff09d93.png)
![VC_DB_diagram](https://user-images.githubusercontent.com/4853767/197075507-034f2fd4-dcdc-4d1e-93b7-00d594e25a4f.png)

pombredanne:
As an example we generate models docs in SCIO here https://scancodeio.readthedocs.io/en/latest/data-models.html

mjherzog:
@johnmhoran We need to widen this page in the RTD as you did for the Audit/Analysis/SPATS RTD. There are already 2 tables there that you cannot read in portrait-only format so we need landscape even before considering how to display a DB diagram. It would be best if you can just widen the page for now since we have not polled the team on changing look & feel overall.

johnmhoran:
Will do @mjherzog .

johnmhoran:
@mjherzog I've added some simple CSS and folder structure etc. and the vulnerablecode RTD page now has the wide look of the other RTD.  The latter has much more CSS for various widgets, dropdowns, a navbar etc. and I'm not proposing we add all of that to the vulnerablecode RTD.  However, before I commit and push I could show you what we now have locally and get any immediate additional feedback you might have.  Comparisons pasted below.  How would you like to proceed?

Before:

![image](https://github.com/nexB/vulnerablecode/assets/11096678/f5190737-106b-482b-9462-ede913d86ba6)

After:

![image](https://github.com/nexB/vulnerablecode/assets/11096678/549c7279-6c24-4f03-9797-a8c62f1c9920)

mjherzog:
Please proceed with the width change for now. I think that we will want to make landscape and a default and add some or all of the other features from the private RTDs, but we need to engage a broader group who are interested in our doc style. I am guessing Ayan, Dennis, Thomas and Philippe, but we need to open the floor to all nexB people and likely the AboutCode community people who are interested.

[...]

PR #1339, opened 2023-11-15
Widen the RTD page #977

johnmhoran:
Reference: https://github.com/nexB/vulnerablecode/issues/977

Note: I just noticed that I named the branch with 997 but the issue is 977.

### DIFF
--- a/docs/source/_static/theme_overrides.css
+++ b/docs/source/_static/theme_overrides.css
@@ -1,0 +1,20 @@
+/* this is the container for the pages */
+.wy-nav-content {
+    max-width: 100%;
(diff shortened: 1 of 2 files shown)


AyanSinhaMahapatra (approved):
Thanks++ @jmhoran looks great now with the latest changes, on laptop/mobile/larger monitors. Readability is actually greatly improved on larger screens. 
We definately want this in other RTD pages too, so please also add this at https://github.com/nexB/skeleton

TG1999:
@johnmhoran @AyanSinhaMahapatra thanks ++, merging this

PR #1354, opened 2023-11-30
Drop package_managers in favour of fetchcode.package_versions

### LINKED ISSUE #1349: Drop package_managers module in favor of pcakage_versions from FetchCode
`package_managers` module is in the process of migration to FetchCode https://github.com/nexB/fetchcode/pull/93 also going forward the package_managers will be renamed and referred to as `package_versions`.

keshav-space:
- fixes #1349


vulnerabilities/improvers/valid_versions.py
```diff
@@ -74,21 +66,16 @@ def get_package_versions(
...
-            logger.error(f"Could not get versions for {package_url!r}")
-            return []
-        versions_fetcher = self.versions_fetcher_by_purl.get(package_url)
-        if not versions_fetcher:
-            versions_fetcher = get_version_fetcher(package_url)
-            self.versions_fetcher_by_purl[package_url] = versions_fetcher()
-
-        versions_fetcher = self.versions_fetcher_by_purl[package_url]
+        versions = package_versions.versions(str(package_url))
+        versions_before_until = set()
```
pombredanne, line 72: The function signatures tell that we return a list not a set.... Do we really need a set?

vulnerabilities/improvers/valid_versions.py
```diff
@@ -74,21 +66,16 @@ def get_package_versions(
...
+        for version in versions or []:
```
pombredanne, line 73: The `[]` may not be needed because we yield but this is safe to keep it.

pombredanne (approved):
LGTM.... just a minor nit for your consideration wrt. set(s)

PR #1359, opened 2023-12-06
Add endpoint for purl lookup

TG1999:
fixes : #1357

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -352,14 +352,36 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
         """
         Return the response for exact PackageURL requested for.
         """
-        purl = request.data.get("purl", []) or None
+        purl = request.data.get("purl")
         if not purl:
             return Response(
                 status=400,
                 data={"Error": "A 'purl' is required."},
             )
-        query = Package.objects.filter(package_url=purl).distinct()
-        return Response(PackageSerializer(query, many=True, context={"request": request}).data)
+        return Response(
+            PackageSerializer(
+                Package.objects.for_purls([purl]), many=True, context={"request": request}
+            ).data
+        )
+
+    @action(detail=False, methods=["post"])
+    def bulk_lookup(self, request):
+        """
+        Return the response for exact PackageURLs requested for.
+        """
+        purls = request.data.get("purls") or []
+        if not purls:
+            return Response(
+                status=400,
+                data={"Error": "A non-empty 'purls' list of PURLs is required."},
+            )
+        return Response(
+            PackageSerializer(
+                PackageSerializer(Package.objects.for_purls(purls)),
+                many=True,
+                context={"request": request},
+            ).data
+        )
 
 
 class VulnerabilityFilterSet(filters.FilterSet):
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -377,7 +377,7 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
             )
         return Response(
             PackageSerializer(
-                PackageSerializer(Package.objects.for_purls(purls)),
+                Package.objects.for_purls(purls),
                 many=True,
                 context={"request": request},
             ).data
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -347,6 +347,20 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
         vulnerable_purls = [str(package.package_url) for package in vulnerable_packages]
         return Response(vulnerable_purls)
 
+    @action(detail=False, methods=["post"])
+    def lookup(self, request):
+        """
+        Return the response for exact PackageURL requested for.
+        """
+        purl = request.data.get("purl", []) or None
+        if not purl:
+            return Response(
+                status=400,
+                data={"Error": "A 'purl' is required."},
+            )
+        query = Package.objects.filter(package_url=purl).distinct()
+        return Response(PackageSerializer(query, many=True, context={"request": request}).data)
+
 
 class VulnerabilityFilterSet(filters.FilterSet):
     class Meta:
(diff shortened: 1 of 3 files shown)


johnmhoran:
@TG1999 I tried to test this new PURL lookup endpoint but cannot find the branch `lookup_endpoint` anywhere in the nexB repo.  How do I access that branch?  Or is that branch not in the nexB vulnerablecode repo (where I expect it should be)?  It almost looks like the branch is in a  fork on your own repo, but I don't know why that would be the case -- all of our work should be in the nexB vulnerablecode repo, right?

johnmhoran:
@TG1999   BTW, I've been doing some initial testing of our API bulk PURL search endpoint and other search endpoints/methods -- using `pkg:nginx/nginx@0.8.9?os=windows` which we know has two duplicate records with the same PURL -- and have found that:

- my local .py single-PURL API search returns 3 PURLs: 2 dupes of `pkg:nginx/nginx@0.8.9?os=windows` plus `pkg:nginx/nginx@0.8.9`

- my local .py bulk-PURL API search returns 2 PURLs: 2 dupes of `pkg:nginx/nginx@0.8.9?os=windows` -- just what we want (except for the duplication) i.e., only the PURL I searched for -- works for a single-PURL search so maybe we don't need a new lookup endpoint?

- public Package UI single-PURL search returns 3 PURLs: 2 dupes of `pkg:nginx/nginx@0.8.9?os=windows` plus `pkg:nginx/nginx@0.8.9`

- public API UI single-PURL search (GET /api/packages/) returns 3 PURLs: 2 dupes of `pkg:nginx/nginx@0.8.9?os=windows` plus `pkg:nginx/nginx@0.8.9`

- public API UI bulk-PURL search (POST /api/packages/bulk_search/): throws an error

Unable to get this to work.

The required input template differs from the structure that works in my local .py file.

```
{
  "purl": "string",
  "type": "string",
  "namespace": "string",
  "name": "string",
  "version": "string",
  "qualifiers": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  },
  "subpath": "string"
}
```

I tried this without success:

```
{
  "purl": "pkg:nginx/nginx@0.8.9?os=windows",
  "type": "nginx",
  "namespace": "",
  "name": "nginx",
  "version": "0.8.9",
  "qualifiers": {
    "os": "windows"
  },
  "subpath": ""
}
```

Response:

```
{
  "Error": "A non-empty 'purls' list of PURLs is required."
}
```

I also tried the following structure without success (but this DOES work in my local .py file AND is recommended in our RTD (https://vulnerablecode.readthedocs.io/en/latest/api.html#package-bulk-search)): 

```
{
    "purls": ["pkg:nginx/nginx@0.8.9?os=windows"],
    "purl_only": False,
    "plain_purl": False,
}
```

Response:

```
{
  "detail": "JSON parse error - Expecting value: line 3 column 18 (char 70)"
}
```

johnmhoran:
Does the public API UI bulk-PURL search (POST /api/packages/bulk_search/) actually work and I just can't follow directions?  ;-)

TG1999:
@johnmhoran see https://github.com/nexB/dejacode/blob/2bdb96d1fefca88e10db4c4e0c8efdcecb05864f/dejacode_toolkit/vulnerablecode.py#L69 how bulk_search API is used here

johnmhoran:
@TG1999 I don't think you read my comments very closely.  I already have the bulk search API working in my Python code -- but I am unable to get the Public API UI bulk search working (here: https://public.vulnerablecode.io/api/docs/), and I explained what the experience was with that.  I also do not know where your lookup branch is.  All of that is detailed in my comments above, including my successful .py use of this structure:

```
{
    "purls": ["pkg:nginx/nginx@0.8.9?os=windows"],
    "purl_only": False,
    "plain_purl": False,
}
```

vulnerabilities/api.py
```diff
@@ -347,6 +347,20 @@ def all(self, request):
         vulnerable_purls = [str(package.package_url) for package in vulnerable_packages]
         return Response(vulnerable_purls)
 
+    @action(detail=False, methods=["post"])
+    def lookup(self, request):
+        """
+        Return the response for exact PackageURL requested for.
+        """
+        purl = request.data.get("purl", []) or None
```
TG1999, line 355:
```suggestion
        purl = request.data.get("purl")
```

vulnerabilities/api.py
```diff
@@ -347,6 +347,20 @@ def all(self, request):
         vulnerable_purls = [str(package.package_url) for package in vulnerable_packages]
```
TG1999, line 347: Also add `bulk_lookup`

TG1999 (commented):
Some nits for my consideration

TG1999:
Added `bulk_lookup` endpoint with tests, please review

johnmhoran:
@TG1999 I've added your repo as a new remote and can now access your `lookup_endpoint` branch.  So far I've focused only on the new `lookup` endpoint.

Based on the [tests](https://github.com/nexB/vulnerablecode/pull/1359/files#diff-d3ca0948dc3b5eb0b1adecaa9da9d7854628b0b6bbcf5f515bed6cab4d894339R754-R831), I expected this structure to work with my local instance:

```
headers = {"Authorization": f"Token {token}"}

request_body = {"purl": "pkg:nginx/nginx@0.8.9?os=windows"}

response = requests.post(
    "http://127.0.0.1:8001/api/packages/lookup",
    data=json.dumps(request_body),
    content_type="application/json",
    headers=headers,
).json()
```

However, this threw an error:

```
TypeError: Session.request() got an unexpected keyword argument 'content_type'
```

Not sure why that works in your tests but not for me.  Googling was unhelpful, and the variations I tried in the key 'content_type' did not resolve the error.  I finally deleted that line from the `requests.post()` content and instead modified my `headers` like this:

```
headers = {"Authorization": f"Token {token}", "Content-Type": "application/json"}
```

and the following returned the expected list containing two duplicate `pkg:nginx/nginx@0.8.9?os=windows` PURLs:

```
request_body = {"purl": "pkg:nginx/nginx@0.8.9?os=windows"}

response = requests.post(
    "http://127.0.0.1:8001/api/packages/lookup",
    data=json.dumps(request_body),
    headers=headers,
).json()
```

I'll continue my exploration of the `lookup` endpoint and then turn to the `bulk_lookup` endpoint.

johnmhoran:
@TG1999 Very nice!  👍

johnmhoran (approved):
LGTM!!

PR #1366, opened 2023-12-13
Fix swagger API docs generation

keshav-space:
- Use `extend_schema` to override the request body, which was not being properly discovered.

- drf-spectacular relies on `get_serializer_class()` and `get_serializer()`, and it works well for view functions that purely deal with ModelSerializer. For anything else, it gets a bit murky, and it is advised to provide proper overrides in the `extend_schema` decorator.

- Override erroneous pagination and filter backend caused due to response containing multiple serializer object https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-action-is-erroneously-paginated-or-has-filter-parameters-that-i-do-not-want

### DIFF
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -11,8 +11,10 @@ from urllib.parse import unquote
 
 from django.db.models import Prefetch
 from django_filters import rest_framework as filters
+from drf_spectacular.utils import extend_schema
 from packageurl import PackageURL
 from rest_framework import serializers
+from rest_framework import status
 from rest_framework import viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


vulnerabilities/api.py
```diff
@@ -283,21 +295,36 @@ class PackageViewSet(viewsets.ReadOnlyModelViewSet):
...
         """
         Lookup for vulnerable packages using many Package URLs at once.
         """
-
-        purls = request.data.get("purls", []) or []
-        purl_only = request.data.get("purl_only", False)
-        plain_purl = request.data.get("plain_purl", False)
-        if not purls or not isinstance(purls, list):
+        serializer = self.serializer_class(data=request.data)
+        if not serializer.is_valid():
```
TG1999, line 316: Do this function also check for if purls list is empty or not? Or is it needed to be checked after accessing validated_data ?

TG1999 (changes requested):

vulnerabilities/tests/test_api.py
```diff
@@ -647,6 +647,40 @@ def test_bulk_api_with_purl_only_option(self):
...
+        ).json()
+
+        expected = {
+            "error": {"purls": ["This field may not be null."]},
+            "message": "A non-empty 'purls' list of PURLs is required.",
+        }
+
+        self.assertEqual(response, expected)
+
+    def test_bulk_api_without_purls_empty_list(self):
```
TG1999, line 667: @keshav-space thanks, one test with an empty request body as well, just to verify what happens if "purls" property is not provided in the body

TG1999 (approved):
LGTM!

TG1999 (changes requested):
@keshav-space thanks! can you also check for lookup and bulk_lookup endpoint as well

TG1999 (approved):
LGTM!

Issue #1327, opened 2023-11-01
Remove redundant package-urls from VCIO


TG1999:
Currently, we store duplicated package-urls because of the way they are different structurally but similar when constructed as a purl whole, for example:

1st Scenario:

Purl A-
type-"pypi", namespace- "", name - "foo/bar"

Purl B-
type-"pypi", namespace-"foo", name - "bar"

They are structurally different, but their purl will be identical i.e "pkg:pypi/foo/bar"

2nd Scenario:

https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#meeting-on-tuesday-2023-10-24-at-1600-utc discussed here in the weekly meeting

TG1999:
As per my discussion with @pombredanne 
- Take a snapshot of the current database.
- Iterate on vulnerabilities and delete all the packages for vulnerabilities that have an alias

Issue #1278, opened 2023-08-23
Some UI package queries return duplicate copies of the same Package URL


johnmhoran:
Some package queries using the UI return duplicate copies of the same Package URL.  For example, a search for `pkg:deb/debian/jackson-databind@2.12.1-1?distro=sid` returns

```
pkg:deb/debian/jackson-databind@2.12.1-1?distro=stretch
pkg:deb/debian/jackson-databind@2.12.1-1?distro=sid
pkg:deb/debian/jackson-databind@2.12.1-1?distro=stretch
pkg:deb/debian/jackson-databind@2.12.1-1?distro=sid
pkg:deb/debian/jackson-databind@2.12.1-1%2Bdeb11u1?distro=sid
pkg:deb/debian/jackson-databind@2.12.1-1%2Bdeb11u1
pkg:deb/debian/jackson-databind@2.12.1-1%2Bdeb11u1?distro=trixie
pkg:deb/debian/jackson-databind@2.12.1-1?distro=trixie
```

![image](https://github.com/nexB/vulnerablecode/assets/11096678/700c806e-35b9-4690-9235-40e67d0178b1)

johnmhoran:
See also, e.g., `pkg:deb/debian/sqlite3@3.27.2-3%2Bdeb10u1` -- but note that unlike the prior example, the numbers of affected by/fixing vulnerabilities are not 100% identical across the duplicates.

![image](https://github.com/nexB/vulnerablecode/assets/11096678/7683ed0c-fe82-4a00-9f35-eab575e3737d)

johnmhoran:
@pombredanne @TG1999 Last week Philippe suggested I learn how to use the Django shell and search the DB for PURLs groupedby `type`, `namespace`, `name` and `version` -- hoping that the output would include patterns that might shed light on how many duplicates/bad records the DB contains.

I started with a count of packages in my local DB instance -- 596,745, a subset of the public DB.

I initially ran the following to generate the result Philippe suggested.

```
from django.db.models import Count

from vulnerabilities.models import Package

duplicate_packages = Package.objects.values('type','namespace','name','version').annotate(type_count=Count('type'),namespace_count = Count('namespace'),name_count = Count('name'),version_count = Count('version')).filter(type_count__gt=1,namespace_count__gt=1,name_count__gt=1,version_count__gt=1)

duplicate_packages
```

This gave me a large output (36,428 records in the queryset) that began like this:

```
[...]

PR #1368, opened 2023-12-19
Package purl model updates

TG1999:
This PR does two major changes:

1. internally we now store Package URL qualifiers as a string and not a JSON field anymore. The API is unchanged for now but will change in the future and will also return a string and not a dict/object.

2. Because of qualifiers storage inconsistencies, we had several data issues that the schema changes addresses. We also have namespaces inconsistencies issues described in https://github.com/nexB/vulnerablecode/issues/1327 and we use a sequence of data and schema migrations to fix these issues and normalize the storage of name and namespaces for all records now and in the future. 

Therefore, this PR fixes:
- https://github.com/nexB/vulnerablecode/issues/1278
- https://github.com/nexB/vulnerablecode/issues/1327


vulnerabilities/models.py
```diff
@@ -561,33 +509,44 @@ def for_purls(self, purls=[]):
...
+    plain_purl = utils.plain_purl(purl=purl)
+    return purl_to_dict(plain_purl, with_empty=False)
 
 
 class Package(PackageURLMixin):
     """
     A software package with related vulnerabilities.
     """
 
+    """
```
pombredanne, line 525: These comments may no longer be needed now.

pombredanne (approved):
LGTM... Just a minor nit wrt. dangling comments that you may want to remove.

Issue #1403, opened 2024-01-25
Encoding issue in  `vulnerabilities.importers.fireeye.FireyeImporter`


harsh098:
I was receiving this small error on running the importer `vulnerabilities.importers.fireeye.FireyeImporter` 
turns out this was caused by wrong encoding type. 

## Error
```
Importing data using vulnerabilities.importers.fireeye.FireyeImporter
        Fetching Advisories █████∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙ 17%
Traceback (most recent call last):
  File "/home/harsh/vulnerablecode/vulnerabilities/management/commands/import.py", line 64, in import_data
    ImportRunner(importer).run()
  File "/home/harsh/vulnerablecode/vulnerabilities/import_runner.py", line 61, in run
    count = self.process_advisories(advisory_datas=advisory_datas, importer_name=importer_name)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harsh/vulnerablecode/vulnerabilities/import_runner.py", line 95, in process_advisories
    for data in advisory_datas:
  File "/home/harsh/vulnerablecode/vulnerabilities/importers/fireeye.py", line 57, in advisory_data
    yield parse_advisory_data(raw_data=f.read(), file=file, base_path=base_path)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harsh/vulnerablecode/vulnerabilities/importers/fireeye.py", line 79, in parse_advisory_data
    md_dict = md_list_to_dict(md_list)
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harsh/vulnerablecode/vulnerabilities/importers/fireeye.py", line 153, in md_list_to_dict
    md_dict[md_key].append(md_line)
    ~~~~~~~^^^^^^^^
KeyError: ''
Failed to run importer vulnerabilities.importers.fireeye.FireyeImporter. Continuing...
Fetching Data from Databases |████████████████████████████████| 1/1
CommandError: 1 failed!: vulnerabilities.importers.fireeye.FireyeImporter

```  
---
# How to reproduce 
```
[...]

PR #1404, opened 2024-01-25
Fix Encoding Type in Fireeye Importer

harsh098:
This PR fixes issue #1403


vulnerabilities/importers/fireeye.py
```diff
@@ -44,7 +44,7 @@ def advisory_data(self) -> Iterable[AdvisoryData]:
                 if Path(file).stem == "README":
                     continue
                 try:
-                    with open(file) as f:
+                    with open(file, encoding="utf-8-sig") as f:
```
TG1999, line 47: @harsh098 thanks for your PR, please add tests

TG1999 (changes requested):
Please add tests for verifying the code change :)

vulnerabilities/importers/fireeye.py
harsh098: ok @TG1999

harsh098: Should I add this test for `md_list_to_dict` or `parse_advisory`.
I am currently going for `md_list_to_dict` as the goal to accomplish is to check if
`md_list_to_dict` works with the `utf-8-sig` encoding or not

TG1999 (approved):
LGTM!

Issue #1419, opened 2024-02-15
Add URL to the vulnerability details view in the API serializers


tdruez:
For example, in the https://public.vulnerablecode.io/api/vulnerabilities/14104?format=json results, I'd like to have a field that includes the URL https://public.vulnerablecode.io/vulnerabilities/VCID-2v89-k9c8-aaah value.

This is required to provide a link to resources when displaying the API results, see https://github.com/nexB/dejacode/issues/4

The field name could be something along `resource_url` or `view_url`.

Note that this should be added to all the API serializers that have a details view in the main UI, like packages.

barkhaaroraa:
could i work on this issue?

pombredanne:
@barkhaaroraa I think that @TG1999 has it mostly solved

TG1999:
closed by https://github.com/nexB/vulnerablecode/pull/1423

PR #1423, opened 2024-02-19
Add URL to the vulnerability and package details view in the API serializers

TG1999:
Reference: https://github.com/nexB/vulnerablecode/issues/1419


tdruez (changes requested):
- Use a common naming convention instead of `package_url` and `vulnerability_url`
- The URL build logic should live on the model for a global usage, not on the API, see https://docs.djangoproject.com/en/5.0/ref/models/instances/#get-absolute-url

TG1999:
@tdruez we already have `get_absolute_url` method on both Package and Vulnerability models, but it returns `/packages/pkg:foo/bar` and not `public.vulnerablecode.io/packages/pkg:foo/bar`, I have added a new method get_details_url for the needed purpose. Please review, thanks!

tdruez:
@TG1999 LGTM but the documentation check is failing. Ready to merge otherwise.

PR #1445, opened 2024-03-21
Fix #1443 Update import in vulntotal datasources

### LINKED ISSUE #1443: AppRegistryNotReady error when setting up vulntotal
### Description

I encountered an `AppRegistryNotReady` error when trying to set up vulntotal using Docker. This issue arises at the final step when running `vulntotal --help`.

### Steps to Reproduce

1. Ran `docker-compose up` to start the services.
2. Executed `docker-compose exec vulnerablecode pip install -r requirements.txt` to install Python dependencies.
3. Ran `docker-compose exec vulnerablecode pip install -e .` to install the current package.
4. Attempted to run `docker-compose exec vulnerablecode vulntotal --help` to check if vulntotal was correctly set up.

At the last step, I encountered the following error:

```console
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

Upon investigating, I found that the root cause is an import statement in `github.py`, which tries to import from `vulnerabilities import utils`. Specifically, `vulnerabilities.utils` contains the line `from packageurl.contrib.django.models import without_empty_values`, which seems to trigger the error. It appears that vulntotal's operation is being impeded by its dependency on Django's app loading state.

### Expected Behavior

I expected to be able to run `vulntotal --help` without encountering an `AppRegistryNotReady` error, regardless of the Django app's state.

### Actual Behavior

An `AppRegistryNotReady` error is raised due to an import statement dependent on Django being fully initialized.

### Possible Solution

It might be beneficial to refactor the code to remove the dependency on Django's app loading state, especially for utility scripts like `vulntotal` that should ideally operate independently.

poju3185:
Fix #1443 

Change the usage of `vulnerabilities.utils.fetch_github_graphql_query` to `fetchcode.package_versions.github_response` in `github.py`


keshav-space:
@poju3185 please add signoff to your commit, and you may want to go through this https://aboutcode.readthedocs.io/en/latest/contributing/writing_good_commit_messages.html.

poju3185:
I've updated my commit with the sign-off. Thanks for the reminder!

keshav-space (approved):
Thanks @poju3185, merging!

PR #1517, opened 2024-07-22
Revert "Add support for reference_type"

TG1999:
Reverts nexB/vulnerablecode#1502

PR #1481, opened 2024-05-19
Add Support to EPSS

ziadhany:
issue #850 #1474 

Screenshot from 2024-06-01 02-59-57

Screenshot from 2024-06-01 03-00-46

PR #1518, opened 2024-07-22
Add support for reference_type

TG1999:
Reverts nexB/vulnerablecode#1517


TG1999:
@ziadhany please see CI is faling here with Rust changes, please check this, so I can merge this and #1481 . Thanks!

ziadhany:
@TG1999 
Update the file `vulnerabilities/tests/test_data/github_api/cargo-expected.json` with the version from EPSS #1481.

TG1999:
@ziadhany please can you push that here? You have access for same ?

ziadhany:
@TG1999 I don't have the permission 

`remote: Permission to nexB/vulnerablecode.git denied to ziadhany.`

this is a patch to fix the test 

[Fix_cargo_test.zip](https://github.com/user-attachments/files/16337156/Fix_cargo_test.zip)

TG1999:
Try pushing again

TG1999 (commented):
LGTM!

ziadhany:
@TG1999 @johnmhoran I've encountered a strange issue in CI where the `apache_kafka` and `apache_tomcat` tests are failing intermittently. However, when I run the tests locally using the same Python version, they all pass. This might be due to the CI running tests in parallel. Do you have any suggestions on how we can resolve this?

keshav-space:
> I've encountered a strange issue in CI where the `apache_kafka` and `apache_tomcat` tests are failing intermittently. However, when I run the tests locally using the same Python version, they all pass. This might be due to the CI running tests in parallel. Do you have any suggestions on how we can resolve this?

@ziadhany the issue is resolved, just needed to regenerate the apache kafka test fixture properly.




Also @ziadhany

- Should we allow the `reference_type` to be empty, or should we use "UNKNOWN" when we cannot properly ascertain the reference type? 
- Additionally, what should we do with the already existing references? Should we assign them the reference type `Advisory`? 
- We will also need to update the existing importers to use the reference type. Do we have any follow-up issues for that?

ziadhany:
@keshav-space Thanks for solving this issue.

>   Should we allow the `reference_type` to be empty, or should we use "UNKNOWN" when we cannot properly ascertain the reference type?

I believe it makes more sense to allow `reference_type` to be empty.

>   Additionally, what should we do with the already existing references? Should we assign them the reference type `Advisory`?

I believe we shouldn't assign a reference type to them since not all references are `Advisory`. Instead, I think we should edit the `process_inferences` in `improver_runner.py` to merge the new references that have a `reference_type`.

>   We will also need to update the existing importers to use the reference type. Do we have any follow-up issues for that?

Yes, I will create a few issues that we need to track to successfully update the reference type.

TG1999:
@ziadhany is it good to be merged ?

ziadhany:
@TG1999 Yes, we can go ahead and merge.

Issue #1453, opened 2024-04-05
Collect exploits from exploitdb https://github.com/offensive-security/exploitdb and https://gitlab.com/exploit-database/exploitdb/-/blob/main/files_exploits.csv


pombredanne:

DennisClark:
@pombredanne  @ziadhany   I do not understand the value of collecting exploits from exploitdb in the context of VulnerableCode.  I don't see any PURLs or other identifiers in that database (perhaps I am missing something) that can relate or connect those bits of information to objects in Vulnerablecode.  We need some kind of explanation here to clarify the strategy.

ziadhany:
@DennisClark 
> I do not understand the value of collecting exploits from exploitdb in the context of VulnerableCode.

Exploit-DB is a well-known online public database and platform that offers detailed information about security vulnerabilities exploits, and associated proof-of-concept (PoC) code. It's widely used by penetration testers. 

In the context of VulnerableCode, I tried to store the PoC as a reference. I believe this is an Improver, as it allows us to gather more detailed information (exploit URL, PoC source URL, ... ) about the vulnerabilities we have.

> I don't see any PURLs or other identifiers in that database (perhaps I am missing something) that can relate or connect those bits of information to objects in Vulnerablecode. 

Yes, Exploit-DB doesn’t provide any PURLs, but it does include an alias like `CVE-xxxx-xxxx` ( **codes** column). We can iterate over these aliases and update the reference list by appending the exploit URL, PoC source URL, and other relevant links.

Here's a sample of the CSV file:
| id    | file                       | description                                                                           | date_published | author          | type | platform | port | date_added | date_updated | verified | codes                         | tags                      | aliases                | screenshot_url | application_url | source_url                                                         |
|-------|----------------------------|---------------------------------------------------------------------------------------|----------------|-----------------|------|----------|------|-------------|--------------|----------|------------------------------|---------------------------|------------------------|----------------|-----------------|--------------------------------------------------------------------|
| 16929 | exploits/aix/dos/16929.rb   | AIX Calendar Manager Service Daemon (rpc.cmsd) Opcode 21 - Buffer Overflow (Metasploit) | 2010-11-11     | Metasploit      | dos  | aix      |      | 2010-11-11  | 2011-03-06   | 1        | CVE-2009-3699;OSVDB-58726    | Metasploit Framework (MSF) |                        |                |                 | http://aix.software.ibm.com/aix/efixes/security/cmsd_advisory.asc |
[...]

PR #1529, opened 2024-07-30
Collect exploit from exploitdb

ziadhany:
issue #1453
depends on #1518


ziadhany:
Close and merge into PR #1562.

Issue #753, opened 2022-05-23
Collect rockylinux advisories


pombredanne:
This is a RH derivative (following CentOS changes) see https://rockylinux.org
https://errata.rockylinux.org/ has the list ... same look as Alma linux #750  but different API https://errata.rockylinux.org/api/advisories and https://errata.rockylinux.org/api/advisories/RLSA-2022:2199

pombredanne:
From today's call: https://errata.rockylinux.org/api/v2/advisories?filters.product=&filters.fetchRelated=true (then paginate &page=0&limit=25)

pombredanne:
@ambuj-1211 FYI

pombredanne:
See:
- https://github.com/rocky-linux/peridot/issues/191

pombredanne:
(I likely filed the peridot issue in the wrong repo) .... Should be in https://github.com/resf/distro-tools/tree/main/apollo instead

AyanSinhaMahapatra:
Btw, from above ^

> Hey - No worries about filing it here. The license for the errata data is the same as RH, that is, CC-BY-4.0.

>    The data resources linked on this page as well as their alternative representations available through the [Security Data API](https://access.redhat.com/documentation/en-us/red_hat_security_data_api/1.0/html/red_hat_security_data_api/index) are licensed under the [Creative Commons Attribution 4.0 International License](https://creativecommons.org/licenses/by/4.0/). If you distribute this content or a modified version of it, you must provide attribution to Red Hat, Inc. and provide a link to the original.

> https://access.redhat.com/security/data

pombredanne:
Per https://github.com/rocky-linux/peridot/issues/191#issuecomment-2248933377

> Hey - No worries about filing it here. The license for the errata data is the same as RH, that is, CC-BY-4.0.
> 
> > The data resources linked on this page as well as their alternative representations available through the Security Data API are licensed under the Creative Commons Attribution 4.0 International License. If you distribute this content or a modified version of it, you must provide attribution to Red Hat, Inc. and provide a link to the original.
> 
> https://access.redhat.com/security/data

PR #1535, opened 2024-08-05
Add rockylinux advisories

ambuj-1211:
fixes: #753

### DIFF
--- a/vulnerabilities/importers/rockylinux.py
+++ b/vulnerabilities/importers/rockylinux.py
@@ -1,0 +1,162 @@
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# VulnerableCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/nexB/vulnerablecode for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+import logging
+import re
+from typing import Dict
+from typing import Iterable
+from typing import List
+
+import requests
+from packageurl import PackageURL
+from univers.version_range import RpmVersionRange
+
+from vulnerabilities import severity_systems
+from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
+from vulnerabilities.importer import Importer
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.rpm_utils import rpm_to_purl
+from vulnerabilities.utils import get_cwe_id
+from vulnerabilities.utils import get_item
+from vulnerabilities.utils import requests_with_5xx_retry
+
+logger = logging.getLogger(__name__)
+
+# FIXME: we should use a centralized retry
+requests_session = requests_with_5xx_retry(max_retries=5, backoff_factor=1)
+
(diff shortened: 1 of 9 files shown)


ambuj-1211:
@ziadhany please review this pr, I have added the Rockylinux importer and tests for it.

vulnerabilities/importers/rockylinux.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def advisory_data(self) -> Iterable[AdvisoryData]:
+
+        for rockylinux_cve in fetch_cves():
+            yield to_advisory(rockylinux_cve)
+
+
+def to_advisory(advisory_data):
+    aliases = advisory_data.get("name") or ""
+    date_published = dateparser.parse(advisory_data.get("publishedAt", ""))
```
ziadhany, line 73: @ambuj-1211 Could you split this into two lines, please?

vulnerabilities/importers/rockylinux.py
```diff
@@ -0,0 +1,199 @@
...
+
+    for ref in advisory_data.get("cves") or []:
+
+        name = ref.get("name", "")
+        if not isinstance(name, str):
+            logger.error(f"Invalid advisory type {name}")
+            continue
+
+        if "CVE" in name.upper():
+            severity_vector_pattern = r"CVSS:3\.1/([A-Z:/]+)"
```
ziadhany, line 118: @ambuj-1211 Why are you using pattern matching? If your goal is to avoid `UNKNOWN`, you should use an `if` statement instead, as pattern matching can be time-consuming.

vulnerabilities/importers/rockylinux.py
```diff
@@ -0,0 +1,199 @@
...
+                            name=purl.name,
+                            namespace=purl.namespace,
+                            qualifiers=purl.qualifiers,
+                            subpath=purl.subpath,
+                        ),
+                        affected_version_range=affected_version_range,
+                        fixed_version=None,
+                    )
+                )
+            except Exception as e:
```
ziadhany, line 100: @ambuj-1211 Please avoid using a general exception and use the appropriate exception for this case instead.

ziadhany (changes requested):
Thanks @ambuj-1211, please see few nits below for your consideration

vulnerabilities/importers/rockylinux.py
ambuj-1211: using pattern matching because the cwe strings were also like this (CWE-273, CWE-123, CWE-546) or they may have other things too like CWE-123 and CWE-345

ambuj-1211: because of this reason I used pattern matching I think if condition won't work what do you say?

ziadhany (changes requested):
@ambuj-1211 Thank you! Please use the new pipeline and address the severity and reference issues.

ambuj-1211:
@ziadhany I have made the changes and created the rockylinux pipeline code, here are the logs for it.
[rockylinux_importer_pipeline_logs.txt](https://github.com/user-attachments/files/17531129/rockylinux_importer_pipeline_logs.txt)

ambuj-1211:
@ziadhany please check this one also I have already made the suggested changes, please tell me if there are any other changes needed.

ambuj-1211:
@ziadhany please review this one I have done most of the things please tell me what changes should I make.

Issue #1454, opened 2024-04-05
Collect exploits from metasploit


pombredanne:
Collect exploits from metasploit https://github.com/rapid7/metasploit-framework and https://www.metasploit.com/

See discussion document at https://docs.google.com/document/d/1XtMmxthmANhr-IqXsyMgFnrOq5fTGfsE/edit?usp=sharing&ouid=117241222429542576816&rtpof=true&sd=true
See work-in-progress normalized model spreadsheet at https://docs.google.com/spreadsheets/d/1J2t2T_s015pnAouy5ss-AA0SI4e2xjT4uICjlL_Aa38/edit?usp=sharing

ziadhany:
I think this link is a good start to get all CVEs that metasploit uses: https://github.com/rapid7/metasploit-framework/discussions/16415

ziadhany:
@DennisClark 
For the [Metasploit Git repository](https://github.com/rapid7/metasploit-framework), I think we are primarily interested in a single file: [modules_metadata_base.json](https://github.com/rapid7/metasploit-framework/blob/master/db/modules_metadata_base.json). This file includes the exploits path and detailed information about each exploit.

We can create a ref from a file path like this  "**/modules/auxiliary/admin/2wire/xslt_password_reset.rb**" -> https://github.com/rapid7/metasploit-framework/tree/master/modules/auxiliary/admin/2wire/xslt_password_reset.rb

also, we can make use of the reference field ( 'https://seclists.org/bugtraq/2007/Aug/225' ), but what type should this reference be categorized as? Should it be set as other, left empty, or classified as an exploit? I think it should be left empty.

And I'm also a bit confused about the licensing because it seems they have different licenses for various files: https://github.com/rapid7/metasploit-framework?tab=License-1-ov-file. What is the license in this situation?

```
{
  "auxiliary_admin/2wire/xslt_password_reset": {
    "name": "2Wire Cross-Site Request Forgery Password Reset Vulnerability",
    "fullname": "auxiliary/admin/2wire/xslt_password_reset",
    "aliases": [

    ],
    "rank": 300,
    "disclosure_date": "2007-08-15",
    "type": "auxiliary",
    "author": [
      "hkm <hkm@hakim.ws>",
      "Travis Phillips"
    ],
    "description": "This module will reset the admin password on a 2Wire wireless router.  This is\n        done by using the /xslt page where authentication is not required, thus allowing\n        configuration changes (such as resetting the password) as administrators.",
[...]

PR #1539, opened 2024-08-06
Add support to metasploit

ziadhany:
issue #1454


ziadhany:
Close and merge into PR #1562.

Issue #1541, opened 2024-08-09
nginx importer fails


pombredanne:
The change at https://github.com/nexB/vulnerablecode/commit/0fe73ef18b5e2af63251eadf9809a03d0a803907#diff-11856d929972b1aa71dcd476298a831d6b598b291c0638057af6d62050ba4a15R192 makes many importers crash.

PR #1542, opened 2024-08-09
Fix importer crash #1541

pombredanne:
This PR is correction for the crash in #1541
The issue was introduced by this commit https://github.com/nexB/vulnerablecode/commit/0fe73ef18b5e2af63251eadf9809a03d0a803907#diff-11856d929972b1aa71dcd476298a831d6b598b291c0638057af6d62050ba4a15R192

We now check for empty published_at before using it. Otherwise the VulnerabilitySeverity.published_at data will be an invalid "None" string resulting from `str(None)`.

Also improve logging.


pombredanne:
@ziadhany @TG1999 FYI, review welcome

vulnerabilities/import_runner.py
```diff
@@ -70,17 +70,23 @@ def do_import(self, advisories) -> None:
...
+                inferences = advisory_importer.get_inferences(advisory_data=advisory_data)
                 process_inferences(
                     inferences=inferences,
                     advisory=advisory,
                     improver_name=importer_name,
                 )
-            except Exception as e:
-                logger.info(f"Failed to process advisory: {advisory!r} with error {e!r}")
+            except Exception:
+                from pprint import pformat
```
ziadhany, line 84: @pombredanne Would it be better to move the `pprint` library import to the top of the code?

ziadhany (approved):
LGTM, and I apologize for introducing this bug.

TG1999 (approved):
LGTM!

PR #1545, opened 2024-08-11
Add prefetch in models

TG1999:


vulnerabilities/api.py
```diff
@@ -1,12 +1,3 @@
-#
```
pombredanne, line 1: You may not want to strip the header, nor change the import style?

Issue #1538, opened 2024-08-06
VCIO-next: API performance issues  (packages endpoints)


tdruez:
Let's focus on the Package-related endpoints for now as those are the ones used to collect vulnerability data in DejaCode.

Those tests were run on a clean install of VCIO with only the `nginx.NginxImporter` data set.
```
Package.objects.all().count()  # 88
Vulnerability.objects.count()  # 39
```

It's a very small amount of data but somehow looking at a single Package triggers over a thousand queries.

- Package list `/api/packages/` (on only 88 packages) -> **6,124 queries**: 5706 similar queries.  Duplicated 73 times.
- Package details `/api/packages/63` -> **1,329 queries**: 1230 similar queries.  Duplicated 16 times.
- Bulk search `/api/packages/bulk_search` (providing the 88 purl): **39,925** queries.

This is quite problematic in the context of batch data collection using the VCIO API.
The `PackageSerializer` and related QuerySets require optimization.
Once done, make sure to implement unit test using the `assertNumQueries` to make sure that future code change do not add uncontrolled queries back.

Related issues:
- https://github.com/aboutcode-org/vulnerablecode/issues/1492
- https://github.com/aboutcode-org/dejacode/issues/94#issuecomment-2298445423 
- https://github.com/aboutcode-org/vulnerablecode/issues/1549

For bulk lookup, we track this here:
- [ ] #1561

pombredanne:
First batch of improvements went in with this PR:
- https://github.com/aboutcode-org/vulnerablecode/pull/1547

More to come

pombredanne:
The performance is now acceptable. We need to9 make sure we save a baseline of performances.

pombredanne:
FYI, the key code to get this fixed is the new API design in #1572

pombredanne:
@tdruez @TG1999 is this completed and closable? If yes, please do!

tdruez:
@TG1999 I've run some data collection using the previous and the new implementation (api v1).

Fetching for 133485 Packages, using bulk_search, with 100 PURL entries per request, so about 1,334 HTTP requests total

- Previous: Completed in 39.5 minutes
- New: Completed in 21.0 minutes

Seems like we double the performance on the package endpoint.

PR #1547, opened 2024-08-12
Improve performance of API calls #1538

pombredanne:
This PR tries to improve the API lookup calls.

One hotspot is the the `Package.is_vulnerable` property that runs an exists() on a queryset
in a repeated manner.

To avoid this, the approach is to annotate the Package queryset(s) with an `is_vulnerable`
attribute.

The results with a single API call to http://127.0.0.1:8001/api/packages/63 are like this
using pyinstrument and django-toolbar on a laptop:

**Before**
* pyinstrument:
```
Absolute time
Duration: 1.46 seconds
Samples: 1376
CPU time: 1.3 seconds
```
* django-toolbar:
```
Time
Resource 	Value
User CPU time 	3992.892 msec
System CPU time 	69.885 msec
Total CPU time 	4062.777 msec
Elapsed time 	4162.812 msec
Context switches 	13364 voluntary, 52 involuntary

SQL
224.49 ms (1329 queries including 1327 similar and 1299 duplicates ) 
```

The pyinstrument hotspot are calls to:
```
Package.is_vulnerable
vulnerabilities/models.py:647
```

**After**:

* pyinstrument:
```
Absolute time
Duration: 0.288 seconds
Samples: 276
CPU time: 0.263 seconds
```

* django-toolbar:

```
Time
Resource 	Value
User CPU time 	395.396 msec
System CPU time 	20.186 msec
Total CPU time 	415.582 msec
Elapsed time 	439.888 msec
Context switches 	2588 voluntary, 17 involuntary

SQL
83.03 ms (99 queries including 96 similar and 68 duplicates ) 
```

Or an improvement looking like this:

| tool           | old       | new      | improvement 
|------          |--------   |----      |---
| pyinstrument   | 1.46 s    | 0.288 s  | x 5.06
| toolbar        | 4.16 s    | 0.439 s  | x 9.47

```


Reference:https://github.com/nexB/vulnerablecode/issues/1538

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -118,6 +118,7 @@ dev =
     commoncode
     # debug
     django-debug-toolbar
+    pyinstrument
 
 [options.entry_points]
 console_scripts =
--- a/vulnerabilities/api.py
+++ b/vulnerabilities/api.py
@@ -108,6 +108,8 @@ class MinimalPackageSerializer(BaseResourceSerializer):
 
     purl = serializers.CharField(source="package_url")
 
+    is_vulnerable = serializers.BooleanField()
+
     class Meta:
         model = Package
         fields = ["url", "purl", "is_vulnerable", "affected_by_vulnerabilities"]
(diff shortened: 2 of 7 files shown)


TG1999 (approved):
LGTM!

PR #1559, opened 2024-08-20
Add base pipeline for importers and migrate PyPa importer to aboutcode pipeline

keshav-space:
- Reference: https://github.com/aboutcode-org/vulnerablecode/issues/1509


vulnerabilities/importers/pypa.py
```diff
@@ -1,66 +0,0 @@
-#
-# Copyright (c) nexB Inc. and others. All rights reserved.
```
pombredanne, line 2: In wonder if this should not be git moved to the new importer, to keep some history?

vulnerabilities/pipelines/__init__.py
```diff
@@ -32,3 +40,90 @@ def qualified_name(cls):
...
+        try:
+            import_advisory(
+                advisory=advisory,
+                pipeline_name=self.qualified_name,
+                confidence=self.advisory_confidence,
+                logger=self.log,
+            )
+        except Exception as e:
+            self.log(
+                f"Failed to process advisory: {advisory!r} with error {e!r}",
```
pombredanne, line 127: Use a full stack trace with traceback otherwise we really cannot diagnose easily

pombredanne, line 127:
```suggestion
                f"Failed to import advisory: {advisory!r} with error {e!r}",
```

vulnerabilities/pipelines/pypa_importer.py
```diff
@@ -0,0 +1,73 @@
...
+            cls.clean_downloads,
+        )
+
+    def clone(self):
+        self.log(f"Cloning `{self.repo_url}`")
+        self.vcs_response = fetch_via_vcs(self.repo_url)
+
+    def advisories_count(self):
+        vulns_path = Path(self.vcs_response.dest_dir) / "vulns"
+        return sum(1 for _ in vulns_path.rglob("*.yaml"))
```
pombredanne, line 49: What about this too? or is this to avoid using too much memory?
```suggestion
        return len(list(vulns_path.rglob("*.yaml")))
```

vulnerabilities/pipelines/pypa_importer.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def collect_advisories(self) -> Iterable[AdvisoryData]:
+        base_directory = self.vcs_response.dest_dir
+        base_path = Path(base_directory)
+        vulns_path = base_path / "vulns"
+        revision = Git.get_revision(base_directory) or "main"
+
+        for advisory in vulns_path.rglob("*.yaml"):
+            advisory_url = get_advisory_url(
```
pombredanne, line 58: get_advisory_url() could get a docstring... it's code is not obvious to me

vulnerabilities/tests/__init__.py
```diff
@@ -3,6 +3,53 @@
 # VulnerableCode is a trademark of nexB Inc.
 # SPDX-License-Identifier: Apache-2.0
 # See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
-# See https://github.com/nexB/vulnerablecode for support or download.
+# See https://github.com/aboutcode-org/vulnerablecode for support or download.
 # See https://aboutcode.org for more information about nexB OSS projects.
 #
+
+from django.utils import timezone
```
pombredanne, line 10: Getting these "fixtures" in the __init__ looks a little bit weird to me. Why not move these directly in the modules you use them? Also it is OK to have these duplicated in each module IMHO as it may help with understanding the code and avoid dependencies between tests?

pombredanne (commented):
Thank. Some nits for your consideration!

vulnerabilities/importers/pypa.py
keshav-space: Done, we're preserving git history now https://github.com/aboutcode-org/vulnerablecode/commits/ec59d7acb0726c33076367a1a4b085d01f862d3b/vulnerabilities/importers/pypa.py?browsing_rename_history=true&new_path=vulnerabilities/pipelines/pypa_importer.py&original_branch=1509-pypa-importer-pipeline

vulnerabilities/pipelines/pypa_importer.py
keshav-space: Yes, did that to avoid keeping all the Path objects in memory at the same time.

vulnerabilities/pipelines/pypa_importer.py
keshav-space: Done.

vulnerabilities/tests/__init__.py
keshav-space: Done, moved fixtures to individual test modules.

pombredanne (approved):
LGTM!

PR #1571, opened 2024-08-27
Improve export command

pombredanne:
This PR improves the "export" command for FederatedCode
1. the export now uses less memory and does not load the whole DB at once
2. we no longer export DB ids
3. the output schema is simplified
4. there is a new "aboutcode.hashid" package that is meant to be reused in FederatedCode

This is ready for review. I will likely use Serializers in the future


keshav-space (approved):
Thanks @pombredanne, LGTM!

PR #1599, opened 2024-09-24
Use correct regex for CVE

pombredanne:
Per the CVE JSON schema we had not the correct regex.
We4 now use the official regex from the CVE schema


pombredanne:
@TG1999 could you review and merge? Thanks!

Issue #1618, opened 2024-10-16
Implement better pagination widgets in ScanCode.io


pombredanne:

nilaygit-10721:
please assign me this issue

Issue #1617, opened 2024-10-16
Make the "pager" usable when browsing search results


pombredanne:
Shared by a shy user through a private email:

> BTW, I just ran into the pager on Vulnerable Code, this could be improved:
>
> https://public.vulnerablecode.io/packages/search?page=2&search=sudo
>
> If you want to go to page 45 of results, you have to visit each and every single page before it by pressing next 44 times...

We should have better widgets for next/previous pagination... and also:

- [ ] Implement better pagination widgets in ScanCode.io 

This is related to:
-  https://github.com/aboutcode-org/vulnerablecode/issues/1616

Rishi-source:
In my approach for solving this issue 

the following improvements to our pagination functionality are recommended:

1. Direct Page Navigation:
   Implement an input field allowing users to specify and navigate directly to a desired page number.

2. Last Page Navigation:
   Add functionality to navigate immediately to the final page of results.

3. Enhanced Page Number Display:
   Implement a more comprehensive page number display system, showing a wider range of page numbers (e.g., 1, 2, 3, ... 
   10, 11, 12, ... 98, 99, 100) to facilitate easier navigation across the result set.

4. Alternative Content Loading Methods:
   Consider implementing either infinite scrolling or a "Load More" button as an alternative to traditional pagination.

5. Adjustable Items Per Page:
   Provide users with the option to increase the number of items displayed per page, thereby reducing the total number of 
    pages.

6. Keyboard Navigation:
   Implement keyboard shortcuts to enable efficient navigation through search results using keyboard inputs.

Please Do let me know if you have some other ideas...

pombredanne:
and here is some private feedback I received:

> I think 2 is the bare minimum, and 4 is obvious.
> 
> All other options cater to specific usage patterns, but depend on
> scripting.
> 
> One more option (also requires scripting) is to filter as you
> type (showing only the subset of results containing the string
> you've typed). So that if I type "main" or "armhf" I only see those.

pombredanne:
Based on this starting with 2. using a standard Bulma would be best IMHO. And we can expand later

Rishi-source:
Ok I'll Start scripting with the 2 and 4.

Issue #1616, opened 2024-10-16
Display more results per page when searching for packages


pombredanne:
It would be great to display more results per page when searching for packages. 20 is not much. 
But at the same time, a generic search is not the main use case. Instead a PURL-based search is the main use case.

PR #1625, opened 2024-10-19
Pagination Updated - Items per page option added

Rishi-source:
Add pagination to VCIO & SCIO #1616 #1617 #1618

This change aligns Package Search functionality and Vulnerability Search, improving consistency across the application and enhancing user control over result display.

### **For Package Search:**

Screenshot 2024-10-18 at 4 42 58 PM

### **For Vulnerability Search:**

Screenshot 2024-10-18 at 4 43 39 PM

Signed-off-by: Rishi Garg [rishigarg2503@gmail.com](mailto:rishigarg2503@gmail.com)

### DIFF
--- a/vulnerabilities/pagination.py
+++ b/vulnerabilities/pagination.py
@@ -1,6 +1,145 @@
+import logging
+import re
+
+from django.core.paginator import Paginator
+from django.db.models.query import QuerySet
 from rest_framework.pagination import PageNumberPagination
+
+logger = logging.getLogger(__name__)
 
 
 class SmallResultSetPagination(PageNumberPagination):
     page_size_query_param = "page_size"
     max_page_size = 100
+
+class PaginatedListViewMixin:
+    """
+    A mixin that adds pagination functionality to ListView-based views.
+    """
+
+    paginate_by = 20
+    max_page_size = 100
+    PAGE_SIZE_CHOICES = [
+        {"value": 20, "label": "20 per page"},
+        {"value": 50, "label": "50 per page"},
+        {"value": 100, "label": "100 per page"},
+    ]
+
+    def get_queryset(self):
+        """
+        Ensure a queryset is always available
+        """
+        try:
+            queryset = super().get_queryset()
+            if not queryset:
+                queryset = self.model.objects.all()
+            if not isinstance(queryset, QuerySet):
+                queryset = self.model.objects.all()
+            return queryset
+        except Exception as e:
+            logger.error(f"Error in get_queryset: {e}")
+            return self.model.objects.all()
+
+    def sanitize_page_size(self, raw_page_size):
+        """
+        Sanitize page size input to prevent XSS and injection attempts.
+        """
+        if not raw_page_size:
+            return self.paginate_by
+        clean_page_size = re.sub(r"\D", "", str(raw_page_size))
+        try:
+            page_size = int(clean_page_size) if clean_page_size else self.paginate_by
+            valid_sizes = {choice["value"] for choice in self.PAGE_SIZE_CHOICES}
+            if page_size not in valid_sizes:
+                logger.warning(f"Attempted to use unauthorized page size: {page_size}")
+                return self.paginate_by
+            return page_size
+        except (ValueError, TypeError):
+            logger.info("Empty or invalid page_size input attempted")
+            return self.paginate_by
+
+    def get_paginate_by(self, queryset=None):
+        """
+        Get the number of items to paginate by from the request.
+        """
+        raw_page_size = self.request.GET.get("page_size")
+        return self.sanitize_page_size(raw_page_size)
+
+    def get_page_range(self, paginator, page_obj):
+        """
+        Generate a list of page numbers for navigation
+        """
+        num_pages = paginator.num_pages
+        current_page = page_obj.number
+        if num_pages <= 7:
+            return list(range(1, num_pages + 1))
+        pages = []
+        pages.append(1)
+        if current_page > 4:
+            pages.append("...")
+        start = max(2, current_page - 2)
+        end = min(num_pages - 1, current_page + 2)
+        pages.extend(range(start, end + 1))
+        if current_page < num_pages - 3:
+            pages.append("...")
+        if num_pages > 1:
+            pages.append(num_pages)
+        return [str(p) for p in pages]
+
+    def paginate_queryset(self, queryset, page_size):
+        """
+        Robust pagination with comprehensive error handling
+        """
+        try:
+            if not queryset or queryset.count() == 0:
+                queryset = self.model.objects.all()
+            paginator = Paginator(queryset, page_size)
+            page_params = self.request.GET.getlist("page")
+            page_number = page_params[-1] if page_params else "1"
+            try:
+                page_number = int(re.sub(r"\D", "", str(page_number)))
+                if not page_number:
+                    page_number = 1
+            except (ValueError, TypeError):
+                page_number = 1
+            page_number = max(1, min(page_number, paginator.num_pages))
+            page = paginator.page(page_number)
+            return (paginator, page, page.object_list, page.has_other_pages())
+        except Exception as e:
+            logger.error(f"Pagination error: {e}")
+            queryset = self.model.objects.all()
+            paginator = Paginator(queryset, page_size)
+            page = paginator.page(1)
+            return (paginator, page, page.object_list, page.has_other_pages())
+
+    def get_context_data(self, **kwargs):
+        """
+        Return a mapping of pagination-related context data, preserving filters.
+        """
+        queryset = self.get_queryset()
+        page_size = self.get_paginate_by()
+        paginator, page, object_list, is_paginated = self.paginate_queryset(queryset, page_size)
+        page_range = self.get_page_range(paginator, page)
+
+        search = self.request.GET.get("search", "")
+
+        context = super().get_context_data(
+            object_list=object_list,
+            page_obj=page,
+            paginator=paginator,
+            is_paginated=is_paginated,
+            **kwargs,
+        )
+
+        context.update(
+            {
+                "current_page_size": page_size,
+                "page_size_choices": self.PAGE_SIZE_CHOICES,
+                "total_count": paginator.count,
+                "page_range": page_range,
+                "search": search,
+                "previous_page_url": page.previous_page_number() if page.has_previous() else None,
+                "next_page_url": page.next_page_number() if page.has_next() else None,
+            }
+        )
+        return context
--- a/vulnerabilities/pagination.py
+++ b/vulnerabilities/pagination.py
@@ -12,6 +12,7 @@ class SmallResultSetPagination(PageNumberPagination):
     page_size_query_param = "page_size"
     max_page_size = 100
 
+
 class PaginatedListViewMixin:
     """
     A mixin that adds pagination functionality to ListView-based views.
--- a/vulnerabilities/templates/includes/pagination.html
+++ b/vulnerabilities/templates/includes/pagination.html
@@ -1,39 +1,76 @@
-<nav class="pagination is-centered is-small" aria-label="pagination">
-  {% if page_obj.has_previous %}
-  <a href="?page={{ page_obj.previous_page_number }}&search={{ search|urlencode }}" class="pagination-previous">Previous</a>
-  {% else %}
-    <a class="pagination-previous" disabled>Previous</a>
-  {% endif %}
+<div class="pagination-controls mb-4">
+  <div class="is-flex is-justify-content-center mb-3">
+      <div class="select is-small {% if total_count < current_page_size %}is-disabled{% endif %}">
+          <select onchange="handlePageSizeChange(this.value)" 
+                  {% if total_count < current_page_size %}disabled="disabled"{% endif %}>
+              {% for choice in page_size_choices %}
+                  <option value="{{ choice.value }}" 
+                          {% if choice.value == current_page_size %}selected{% endif %}>
+                      {{ choice.label }}
+                  </option>
+              {% endfor %}
+          </select>
+      </div>
+  </div>
 
-  {% if page_obj.has_next %}
-    <a href="?page={{ page_obj.next_page_number }}&search={{ search|urlencode }}" class="pagination-next">Next</a>
-  {% else %}
-    <a class="pagination-next" disabled>Next</a>
-  {% endif %}
+  {% if page_obj.paginator.num_pages > 1 %}
+  <nav class="pagination is-centered" role="navigation" aria-label="pagination">
+      {% if page_obj.has_previous %}
+          <a href="?page={{ page_obj.previous_page_number }}&search={{ search|urlencode }}&page_size={{ current_page_size }}" 
+             class="pagination-previous">Previous</a>
+      {% else %}
+          <button class="pagination-previous" disabled>Previous</button>
+      {% endif %}
 
-  <ul class="pagination-list">
-    {% if page_obj.number != 1%}
-      <li>
-        <a href="?page=1&search={{ search|urlencode }}" class="pagination-link" aria-label="Goto page 1">1</a>
-      </li>
-      {% if page_obj.number > 2 %}
-        <li>
-          <span class="pagination-ellipsis">&hellip;</span>
-        </li>
+      {% if page_obj.has_next %}
+          <a href="?page={{ page_obj.next_page_number }}&search={{ search|urlencode }}&page_size={{ current_page_size }}" 
+             class="pagination-next">Next</a>
+      {% else %}
+          <button class="pagination-next" disabled>Next</button>
       {% endif %}
-    {% endif %}
-    <li>
-      <a class="pagination-link is-current" aria-label="Page {{ page_obj.number }}" aria-current="page">{{ page_obj.number }}</a>
-    </li>
-    {% if page_obj.number != page_obj.paginator.num_pages %}
-      {% if page_obj.next_page_number != page_obj.paginator.num_pages %}
-        <li>
-          <span class="pagination-ellipsis">&hellip;</span>
-        </li>
-      {% endif %}
-      <li>
-        <a href="?page={{ page_obj.paginator.num_pages }}&search={{ search|urlencode }}" class="pagination-link" aria-label="Goto page {{ page_obj.paginator.num_pages }}">{{ page_obj.paginator.num_pages }}</a>
-      </li>
-    {% endif %}
-  </ul>
-</nav>
+      <ul class="pagination-list">
+          {% for page_num in page_range %}
+              {% if page_num == '...' %}
+                  <li><span class="pagination-ellipsis">&hellip;</span></li>
+              {% else %}
+                  <li>
+                      <a href="?page={{ page_num }}&search={{ search|urlencode }}&page_size={{ current_page_size }}" 
+                         class="pagination-link {% if page_num == page_obj.number %}is-current{% endif %}"
+                         aria-label="Go to page {{ page_num }}"
+                         {% if page_num == page_obj.number %}aria-current="page"{% endif %}>
+                          {{ page_num }}
+                      </a>
+                  </li>
+              {% endif %}
+          {% endfor %}
+      </ul>
+  </nav>
+  {% endif %}
+</div>
+
+<style>
+.select.is-disabled {
+  opacity: 0.7;
+  cursor: not-allowed;
+}
+.select.is-disabled select {
+  cursor: not-allowed;
+}
+</style>
+
+<script>
+function handlePageSizeChange(value) {
+  const url = new URL(window.location.href);
+  const params = new URLSearchParams(url.search);
+  params.set('page_size', value);
+  params.delete('page');
+  
+  const search = params.get('search');
+  if (search) {
+      params.set('search', search);
+  }
+  
+  const newUrl = `${window.location.pathname}?${params.toString()}`;
+  window.location.href = newUrl;
+}
+</script>
--- a/vulnerabilities/views.py
+++ b/vulnerabilities/views.py
@@ -40,7 +40,7 @@ from .pagination import PaginatedListViewMixin
 PAGE_SIZE = 20
 
 
-class PackageSearch(PaginatedListViewMixin,ListView):
+class PackageSearch(PaginatedListViewMixin, ListView):
     model = models.Package
     template_name = "packages.html"
     ordering = ["type", "namespace", "name", "version"]
@@ -68,7 +68,7 @@ class PackageSearch(PaginatedListViewMixin,ListView):
         )
 
 
-class VulnerabilitySearch(PaginatedListViewMixin,ListView):
+class VulnerabilitySearch(PaginatedListViewMixin, ListView):
     model = models.Vulnerability
     template_name = "vulnerabilities.html"
     ordering = ["vulnerability_id"]
--- a/vulnerabilities/views.py
+++ b/vulnerabilities/views.py
@@ -35,10 +35,12 @@ from vulnerabilities.severity_systems import SCORING_SYSTEMS
 from vulnerablecode import __version__ as VULNERABLECODE_VERSION
 from vulnerablecode.settings import env
 
+from .pagination import PaginatedListViewMixin
+
 PAGE_SIZE = 20
 
 
-class PackageSearch(ListView):
+class PackageSearch(PaginatedListViewMixin,ListView):
     model = models.Package
     template_name = "packages.html"
     ordering = ["type", "namespace", "name", "version"]
@@ -66,7 +68,7 @@ class PackageSearch(ListView):
         )
 
 
-class VulnerabilitySearch(ListView):
+class VulnerabilitySearch(PaginatedListViewMixin,ListView):
     model = models.Vulnerability
     template_name = "vulnerabilities.html"
     ordering = ["vulnerability_id"]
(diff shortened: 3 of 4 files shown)


Rishi-source:
Hi @pombredanne @TG1999 @keshav-space @Hritik14  , Can you please review this PR It has passed all workflow tests and was also working fine on my local server. Thank you

pombredanne:
@TG1999 ping .... @tdruez input welcomed too, as eventually we want to use the same approach across all projects.

vulnerabilities/templates/vulnerabilities.html
```diff
@@ -18,7 +18,17 @@
...
+                    <div class="select is-small">
+                        <select id="itemsPerPage" onchange="changeItemsPerPage(this.value)">
+                            <option value="20" {% if page_obj.paginator.per_page == 20 %}selected{% endif %}>20 per page</option>
+                            <option value="50" {% if page_obj.paginator.per_page == 50 %}selected{% endif %}>50 per page</option>
+                            <option value="100" {% if page_obj.paginator.per_page == 100 %}selected{% endif %}>100 per page</option>
+                            <option value="200" {% if page_obj.paginator.per_page == 200 %}selected{% endif %}>200 per page</option>
+                        </select>
+                    </div>
+                </div>
+                {% if is_paginated %}
```
tdruez, line 31: Duplicated code.

vulnerabilities/templates/vulnerabilities.html
```diff
@@ -77,5 +87,13 @@
...
-
-{% endblock %}
+<script>
+    function changeItemsPerPage(pageSize) {
+        var urlParams = new URLSearchParams(window.location.search);
+        urlParams.set('page_size', pageSize);
+        window.location.search = urlParams.toString();
+    }
+    
+</script>
```
tdruez, line 97: Duplicated code.

vulnerabilities/views.py
```diff
@@ -96,17 +104,34 @@ class VulnerabilitySearch(ListView):
     ordering = ["vulnerability_id"]
     paginate_by = PAGE_SIZE
 
+    def get_paginate_by(self, queryset):
```
tdruez, line 107: Why provide the queryset as an arg? It does not seems to be used.

vulnerabilities/views.py
```diff
@@ -96,17 +104,34 @@ class VulnerabilitySearch(ListView):
...
+        page_size = self.request.GET.get("page_size", "")
```
tdruez, line 108: The user provide value for the page_size should be validated to prevent a user from provide a too large number.

vulnerabilities/views.py
```diff
@@ -96,17 +104,34 @@ class VulnerabilitySearch(ListView):
...
         context["vulnerability_search_form"] = VulnerabilitySearchForm(request_query)
         context["search"] = request_query.get("search")
+        context["page_size"] = self.get_paginate_by(self.get_queryset())
         return context
 
-    def get_queryset(self, query=None):
-        query = query or self.request.GET.get("search") or ""
+    def get_queryset(self):
+        query = self.request.GET.get("search") or ""
         return self.model.objects.search(query=query).with_package_counts()
```
tdruez, line 121: That logic and everything related to search/ordering/filering should be handle by the *SearchForm and not by the view itself.

vulnerabilities/views.py
```diff
@@ -96,17 +104,34 @@ class VulnerabilitySearch(ListView):
...
+        page = self.request.GET.get("page", "1")
+        try:
+            page_number = int(page)
+            page_obj = paginator.page(page_number)
+        except (ValueError, PageNotAnInteger):
+            page_obj = paginator.page(1)
+        except EmptyPage:
+            page_obj = paginator.page(paginator.num_pages)
+        return (paginator, page_obj, page_obj.object_list, page_obj.has_other_pages())
+
```
tdruez, line 134: Why the need to override the default Django `paginate_queryset`? This should be documented.

vulnerabilities/templates/packages.html
```diff
@@ -18,6 +18,16 @@
...
                 </div>
+                <div class="is-flex is-justify-content-center mb-2">
+                    <div class="select is-small">
+                        <select id="itemsPerPage" onchange="changeItemsPerPage(this.value)">
+                            <option value="20" {% if page_obj.paginator.per_page == 20 %}selected{% endif %}>20 per page</option>
+                            <option value="50" {% if page_obj.paginator.per_page == 50 %}selected{% endif %}>50 per page</option>
+                            <option value="100" {% if page_obj.paginator.per_page == 100 %}selected{% endif %}>100 per page</option>
+                            <option value="200" {% if page_obj.paginator.per_page == 200 %}selected{% endif %}>200 per page</option>
+                        </select>
+                    </div>
```
tdruez, line 29: This should be implemented as a Django form for better re-usability.

tdruez (changes requested):

Rishi-source:
@tdruez  please see the changes now workflow tests are also successful .

vulnerabilities/templates/vulnerabilities.html
Rishi-source:
```
    PAGE_CHOICES = [
        ("20", "20 per page"),
        ("50", "50 per page"),
        ("100", "100 per page"),
    ]

```
Now In the latest commit it is address via forms.py.

vulnerabilities/templates/vulnerabilities.html
Rishi-source: Now I have made a `pagination.js` file in `static/js ` and certainly used `<script src="{% static 'js/pagination.js' %}"></script>` to avoid duplication of code.

vulnerabilities/templates/packages.html
Rishi-source: Handled via forms now.

vulnerabilities/views.py
Rishi-source: It’s included to match the expected method signature from Django, which sometimes passes the queryset by default.

vulnerabilities/views.py
Rishi-source: Added the search/ordering/filtering functionality in `BaseSearchForm` , `PackageSearchForm` , `VulnerabilitySearchForm`in forms.py.

vulnerabilities/views.py
Rishi-source: Now I have changed the code there is not overriding of default Django `paginate_queryset`.

Rishi-source:
@tdruez I would request you to please review the PR I have made changes In the code after you requested some changes after first review.

vulnerabilities/forms.py
```diff
@@ -12,26 +12,87 @@
...
+    ]
+
+    page_size = forms.ChoiceField(
+        choices=PAGE_CHOICES,
+        initial="20",
+        required=False,
+        widget=forms.Select(
+            attrs={
+                "class": "select is-small",
+                "onchange": "handlePageSizeChange(this.value)",
```
tdruez, line 34: Avoid inline JavaScript.

vulnerabilities/forms.py
```diff
@@ -12,26 +12,87 @@
...
+    )
+
+
+class BaseSearchForm(forms.Form):
+    """Base form for implementing search functionality."""
+
+    search = forms.CharField(required=True)
+
+    def clean_search(self):
+        return self.cleaned_data.get("search", "")
```
tdruez, line 47: What is the purpose of this?

vulnerabilities/forms.py
```diff
@@ -12,26 +12,87 @@
...
+        Args:
+            query (str, optional): Direct query for testing
+        """
+        if query is not None:
+            return self._search(query)
+
+        if not self.is_valid():
+            return self.model.objects.none()
+
+        return self._search(self.clean_search())
```
tdruez, line 61: What's the need for this logic?

vulnerabilities/forms.py
```diff
@@ -12,26 +12,87 @@
...
     )
 
+    def _search(self, query):
+        """Execute package-specific search logic."""
+        return (
+            self.model.objects.search(query)
+            .with_vulnerability_counts()
+            .prefetch_related()
+            .order_by("package_url")
+        )
```
tdruez, line 80: This is not following the Django forms conventions.

vulnerabilities/forms.py
```diff
@@ -12,26 +12,87 @@
...
     search = forms.CharField(
         required=True,
         widget=forms.TextInput(
             attrs={"placeholder": "Vulnerability id or alias such as CVE or GHSA"}
         ),
     )
 
+    def _search(self, query):
+        """Execute vulnerability-specific search logic."""
+        return self.model.objects.search(query=query).with_package_counts()
```
tdruez, line 94: Same comment as above.

vulnerabilities/templates/includes/pagination.html
```diff
@@ -1,39 +1,56 @@
...
-      {% if page_obj.number > 2 %}
-        <li>
-          <span class="pagination-ellipsis">&hellip;</span>
-        </li>
-      {% endif %}
+{% if is_paginated %}
+<nav class="pagination is-centered" role="navigation" aria-label="pagination">
+    {% if page_obj.has_previous %}
+        <a href="?page={{ page_obj.previous_page_number }}&search={{ search|urlencode }}&page_size={{ page_obj.paginator.per_page }}" 
+           class="pagination-previous">Previous</a>
```
tdruez, line 5: What happens if other filters are active, those are lost when using the pagination?

vulnerabilities/templates/includes/pagination.html
```diff
@@ -1,39 +1,56 @@
...
+    {% if page_obj.has_next %}
+        <a href="?page={{ page_obj.next_page_number }}&search={{ search|urlencode }}&page_size={{ page_obj.paginator.per_page }}" 
+           class="pagination-next">Next</a>
+    {% else %}
+        <span class="pagination-next" disabled>Next</span>
     {% endif %}
-  </ul>
-</nav>
\ No newline at end of file
+
```
tdruez, line 16: All the complexity of those templates should likely be handled in Python.

vulnerabilities/views.py
```diff
@@ -62,51 +63,99 @@ def get_purl_version_class(purl: models.Package):
     return purl_version_class
 
 
-class PackageSearch(ListView):
-    model = models.Package
-    template_name = "packages.html"
-    ordering = ["type", "namespace", "name", "version"]
+class BaseSearchView(ListView):
```
tdruez, line 66: Could we use the `django_filters.views.FilterView` instead of re-implementing it?

tdruez:
> @tdruez what do you mean by "reusable" here. I would code the implementation again to make it more simpler and satisfy the required needs.

> @pombredanne  as eventually we want to use the same approach across all projects.

@Rishi-source Let's say I want to add this pagination on another Django view, could you document the few steps I need to do with this current implementation? This would help to figure out where we can improve the "reusability".

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,58 @@
...
+    paginate_by = 20
+    max_page_size = 100
+
+    PAGE_SIZE_CHOICES = [
+        {"value": 20, "label": "20 per page"},
+        {"value": 50, "label": "50 per page"},
+        {"value": 100, "label": "100 per page"},
+    ]
+
+    def get_paginate_by(self, queryset=None):
```
Rishi-source, line 11: This function handles page size from URL parameters. Sets default to 20 items, max of 100. Validates to ensure positive integers only. Returns default if invalid values provided.

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,58 @@
...
+        try:
+            page_size = int(self.request.GET.get("page_size", self.paginate_by))
+            if page_size <= 0:
+                return self.paginate_by
+            return min(page_size, self.max_page_size)
+        except (ValueError, TypeError):
+            return self.paginate_by
+
+    def get_context_data(self, **kwargs):
```
Rishi-source, line 20: It adds pagination data to template context - current page size, size choices (can be handled by altering PAGE_SIZE_CHOICES) , total count and page range. it also preserves search params across pagination.

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,58 @@
...
+                "page_range": self._get_page_range(
+                    context["paginator"], context["page_obj"].number
+                ),
+                "search": self.request.GET.get("search", ""),
+            }
+        )
+
+        return context
+
+    def _get_page_range(self, paginator, current_page, window=2):
```
Rishi-source, line 40: it creates page number sequence. Shows all numbers for small sets (<5 pages). For larger sets, shows first/last pages around current page with (...) for gaps.

vulnerabilities/views.py
```diff
@@ -36,6 +36,8 @@
 from vulnerabilities.utils import get_severity_range
 from vulnerablecode.settings import env
 
+from .pagination_mixin import PaginatedListViewMixin
```
Rishi-source, line 39: Import the Mixin and the PaginatedListViewMixin class.

vulnerabilities/views.py
```diff
@@ -62,25 +64,21 @@ def get_purl_version_class(purl: models.Package):
     return purl_version_class
 
 
-class PackageSearch(ListView):
+class PackageSearch(PaginatedListViewMixin, ListView):
```
Rishi-source, line 67: Add the PaginatedListViewMixin in it and this would convert any ListView into paginated class.

Rishi-source:
Hi @tdruez , I have altered code to improve the reusability for that I have made a file `pagination_mixin.py` which contains `PaginatedListViewMixin` class . This logic can convert and ListView into a paginated class and can improve the reusability as now you would have to just add this class to you views.

vulnerabilities/templates/includes/pagination.html
```diff
@@ -1,39 +1,77 @@
...
+    
+    const search = params.get('search');
+    if (search) {
+        params.set('search', search);
+    }
+    
+    const newUrl = `${window.location.pathname}?${params.toString()}`;
+    window.location.href = newUrl;
+}
+</script>
```
Rishi-source, line 77: This js function manages URL parameters when changing pagination size. It update the page size parameter while preserving search terms, and remove the current page parameter to reset pagination.

tdruez:
@Rishi-source Thanks but looking at multiple comments is hard to follow. Could you also provide a combined documentation of the few steps required to reuse this pagination in another app?

Rishi-source:
# Pagination Implementation Documentation

## Core Components of Code

1. **Pagination Mixin** (in `pagination_mixins.py`):
2. **Pagination Template** (in `templates/includes/pagination.html`):

## Implementation Steps

1. **Copy the Mixin**
   - Create `pagination_mixins.py` in your app directory
   - Copy the `PaginatedListViewMixin` class code into it

2. **Create the Template** 
   - Create `pagination.html` in your templates directory
   - Copy the pagination template code into it

4. **Use PaginatedListViewMixin in Your View**
```python
from django.views.generic import ListView
from .pagination_mixins import PaginatedListViewMixin # Make sure to import the mixin

class MyListView(PaginatedListViewMixin, ListView):
    model = MyModel
    template_name = "my_template.html"
    
    def get_queryset(self):
        return super().get_queryset().order_by('id') 
```

5. **Include in Your Pagination Template**
```html
{% extends "base.html" %}

{% block content %}

    {# Like This #}
    {% if is_paginated %}
        {% include 'pagination.html' with page_obj=page_obj %}
    {% endif %}

{% endblock %}
```

## Customization Options

1. **Change Page Sizes**
```python
class MyListView(PaginatedListViewMixin, ListView):
    PAGE_SIZE_CHOICES = [
        {"value": 10, "label": "10 per page"},
        {"value": 25, "label": "25 per page"},
        {"value": 50, "label": "50 per page"},
    ]
    paginate_by = 20  # Set the Default size
```

2. **Adjust Window Size**
```python
def get_context_data(self, **kwargs):
    # Change window variable to show more/fewer pages around current page
    context = super().get_context_data(**kwargs)
    context['page_range'] = self._get_page_range(
        context['paginator'], 
        context['page_obj'].number,
        window=3  # Show 3 pages on each side i.e. left and right   
    return context
```

## Notes

- JavaScript is embedded directly in the template via the `onchange` handler
- All pagination state is managed through URL parameters
- Search terms are preserved across pagination
- The mixin automatically handles invalid page numbers and sizes

## Common Issues and Solutions

1. **Missing Page Numbers**
   - Check that `page_range` is being passed correctly in context

2. **URL Parameters Not Preserved**
   - Check template URLs include all necessary parameters
   - Verify parameter names match in views and templates

3. **Page Size Not Working**
   - Verify `page_size` parameter is being passed correctly
   - Check `get_paginate_by` implementation in mixin

4. **Styling Issues**
   - Ensure Bulma CSS is included if using default styling
   - Or customize template classes to match your CSS framework

Remember that this implementation assumes:
- You're using Django's class-based views
- You want URL parameter-based pagination
- You're handling search through GET parameters

Rishi-source:
@tdruez Here’s the documentation on implementing pagination using this approach. Please review and let me know if any changes are required.

Rishi-source:
Hi @tdruez and @pombredanne , I have implemented this pagination code in both VCIO and SCIO and it is working nice.It preserved all the filter and add a option for the user to select the items per page. I have already shared the screenshot for the VCIO pagination view here are the screenshot for SCIO pagination.

@tdruez you just have to follow the steps which I have provided in the [Documentation](https://github.com/aboutcode-org/vulnerablecode/pull/1625#issuecomment-2527411407) for implementing this pagination.

Screenshot 2024-12-10 at 7 41 46 PM
Screenshot 2024-12-10 at 7 42 17 PM

Rishi-source:
Hi @tdruez and @TG1999, Whenever you have some time, could you please review my PR? Your feedback would be greatly appreciated. Thank you!

Rishi-source:
@tdruez @TG1999 Please review my pull request so that I can get to know if any further changes are required or not.

Rishi-source:
Hi @tdruez , Please review this PR .

Rishi-source:
Hi @TG1999 @tdruez , I request you to please review this PR I have completed the code .

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        {"value": 50, "label": "50 per page"},
+        {"value": 100, "label": "100 per page"},
+    ]
+
+    def get_paginate_by(self, queryset=None):
+        """
+        Get the number of items to paginate by from the request.
+        """
+        try:
+            page_size = int(self.request.GET.get("page_size", self.paginate_by))
```
pombredanne, line 20: Is this the standard approach to get a request arg?

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+            if page_size <= 0:
+                return self.paginate_by
+            return min(page_size, self.max_page_size)
+        except (ValueError, TypeError):
+            return self.paginate_by
+
+    def get_pagination_context(self, paginator, page_obj):
+        """
+        Generate pagination-related context data, preserving filters.
```
pombredanne, line 29:
```suggestion
        Return a mapping of pagination-related context data, preserving filters.
```

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        if not paginator or not page_obj:
+            return {}
+
+        current_page_size = self.get_paginate_by()
+        total_count = paginator.count
+
+        query_params = self.request.GET.copy()
+        query_params.pop("page", None)
+
+        base_query_string = query_params.urlencode()
```
pombredanne, line 40: Is this the right way to generate a URL query string? Is there a better way?

vulnerabilities/templates/includes/pagination.html
pombredanne: is there any risk of XSS there?

vulnerabilities/pagination_mixin.py
```diff
@@ -0,0 +1,87 @@
+class PaginatedListViewMixin:
+    """
+    A mixin that adds pagination functionality to ListView-based views.
```
pombredanne, line 3: I would call the file pagination.py instead

pombredanne (changes requested):
Thanks
Here is some feedback!
The PR title should be using imperative style:
Not `Pagination Updated - Items per page option added`

But rather:
"Improve pagination in web ui"
then add details in the body.


Could you also add tests? PaginatedListViewMixin is not trivial code 

Also as @tdruez mentioned in late November, the template should have little to no logic and instead the logic should be in the Python code.

vulnerabilities/pagination.py
```diff
@@ -1,6 +1,146 @@
...
+
+
+class PaginatedListViewMixin:
+    """
+    A mixin that adds pagination functionality to ListView-based views.
+    """
+
+    paginate_by = 20
+    max_page_size = 100
+    PAGE_SIZE_CHOICES = [
```
tdruez, line 23: Why this option is uppercase?

vulnerabilities/pagination.py
```diff
@@ -1,6 +1,146 @@
...
+        """
+        Ensure a queryset is always available
+        """
+        try:
+            queryset = super().get_queryset()
+            if not queryset:
+                queryset = self.model.objects.all()
+            if not isinstance(queryset, QuerySet):
+                queryset = self.model.objects.all()
+            return queryset
```
tdruez, line 39: You should only have 1 instruction in your try block wrapping the thing you are catching the exception for.

tdruez: Also, what's the purpose of overriding this method?

vulnerabilities/pagination.py
```diff
@@ -1,6 +1,146 @@
...
+        if not raw_page_size:
+            return self.paginate_by
+        clean_page_size = re.sub(r"\D", "", str(raw_page_size))
+        try:
+            page_size = int(clean_page_size) if clean_page_size else self.paginate_by
+            valid_sizes = {choice["value"] for choice in self.PAGE_SIZE_CHOICES}
+            if page_size not in valid_sizes:
+                logger.warning(f"Attempted to use unauthorized page size: {page_size}")
+                return self.paginate_by
+            return page_size
```
tdruez, line 57: Same as above, do not put all you logic in a try block, just the code that may raise the exception.

vulnerabilities/pagination.py
```diff
@@ -1,6 +1,146 @@
...
+        pages.append(1)
+        if current_page > 4:
+            pages.append("...")
+        start = max(2, current_page - 2)
+        end = min(num_pages - 1, current_page + 2)
+        pages.extend(range(start, end + 1))
+        if current_page < num_pages - 3:
+            pages.append("...")
+        if num_pages > 1:
+            pages.append(num_pages)
```
tdruez, line 87: The values should not be hardcoded but configurable.

vulnerabilities/pagination.py
```diff
@@ -1,6 +1,146 @@
...
+                page_number = 1
+            page_number = max(1, min(page_number, paginator.num_pages))
+            page = paginator.page(page_number)
+            return (paginator, page, page.object_list, page.has_other_pages())
+        except Exception as e:
+            logger.error(f"Pagination error: {e}")
+            queryset = self.model.objects.all()
+            paginator = Paginator(queryset, page_size)
+            page = paginator.page(1)
+            return (paginator, page, page.object_list, page.has_other_pages())
```
tdruez, line 114: 1. **Inefficient Queryset Handling**
   - `queryset.count() == 0` performs an unnecessary database query.
   - Suggested fix: Use `queryset.exists()` instead.

2. **Overcomplicated Page Number Extraction**
   - `re.sub(r"\D", "", str(page_number))` is unnecessary.
   - Suggested fix: Use `page_number = int(self.request.GET.get("page", "1"))`.

3. **Nested Try-Except Blocks**
   - The current structure makes debugging harder.
   - Suggested fix: Use a single `try-except` block.

4. **Broad Exception Handling**
   - Catching `Exception` hides real bugs.
   - Suggested fix: Handle `EmptyPage` and `InvalidPage` explicitly.

5. **Redundant Queryset Reassignment in Exception Handling**
   - Fetching `self.model.objects.all()` again is unnecessary.
   - Suggested fix: Log the error and return a fallback page.

vulnerabilities/templates/includes/pagination.html
```diff
@@ -1,39 +1,76 @@
...
+  
+  const search = params.get('search');
+  if (search) {
+      params.set('search', search);
+  }
+  
+  const newUrl = `${window.location.pathname}?${params.toString()}`;
+  window.location.href = newUrl;
+}
+</script>
```
tdruez, line 76: A no-JS approach would work better.

tdruez (changes requested):

vulnerabilities/pagination.py
Rishi-source: The method is overridden to ensure that a valid QuerySet is always available for pagination. It includes checks to handle cases where the parent class returns None which is not a QuerySet and throws an exception. Without this pagination could fail.

vulnerabilities/pagination.py
Rishi-source: This part shows numbers and ellipses (1 2 3 4 ... 10 11 12 13) based on hardcoded values. I can replace these with configurable variables to make the display pattern easily adjustable in the future but it will still be hardcoded but in a variable form.

vulnerabilities/templates/includes/pagination.html
Rishi-source: Sure I will remove this JS code and handle is via form tag

Rishi-source:
@tdruez All the necessary changes have been made. Please review the updated code at your convenience.

Rishi-source:
Hello @tdruez, please review this pull request I have made all the changes required.

Rishi-source:
Hi @tdruez , It would be great if you could provide some insights on the code.

Rishi-source:
Hi @pombredanne please review this PR I have added tests and also addressed all the given review points.

PR #1630, opened 2024-10-28
Avoid memory exhaustion during data migration

keshav-space:
There are nearly 15 million package vulnerability relationships in VCIO, and loading them all at once during data migration consumes all the memory.


TG1999:
@keshav-space thanks for this, can we use our own paginate here?

keshav-space:
> @keshav-space thanks for this, can we use our own paginate here?

@TG1999 We cannot use `.paginated()` in data migrations for `PackagerRelatedVulnerability`, as this model was created without our custom queryset manager. https://github.com/aboutcode-org/vulnerablecode/blob/289f4b823b6be636216bae04a637b770a71d1f29/vulnerabilities/models.py#L874-L972
In this case it would be better to use the built-in queryset `iterator`.

keshav-space:
Merging this now!

PR #1636, opened 2024-11-06
Refactor severity score model and fix incorrect suses scores

keshav-space:
- Resolves https://github.com/aboutcode-org/vulnerablecode/issues/1597


vulnerabilities/api_v2.py
```diff
@@ -41,11 +40,24 @@ class Meta:
         fields = ["url", "reference_type", "reference_id"]
 
 
+class VulnerabilitySeverityV2Serializer(serializers.ModelSerializer):
+    class Meta:
+        model = VulnerabilitySeverity
+        fields = ["url", "value", "scoring_system", "scoring_elements", "published_at"]
+
+    def to_representation(self, instance):
```
TG1999, line 48: Why we need `to_representation` method here? if published_at is None, let it be sent as None

vulnerabilities/migrations/0078_alter_vulnerabilityseverity_options_and_more.py
```diff
@@ -0,0 +1,220 @@
+# Generated by Django 4.2.16 on 2024-11-01 12:40
+
```
TG1999, line 2: Did you run this migration locally, how much time did this take ?

vulnerabilities/api_v2.py
keshav-space: Added this just to preserve the response structure we're already using in APIv2 https://github.com/aboutcode-org/vulnerablecode/blob/8a68c97dfa369ad048de3ece14cc1b3cf40591cc/vulnerabilities/api_v2.py#L16C33-L16C64.

IMO yes we should simpy send `None` when we don't have published date.

TG1999 (approved):
LGTM!

vulnerabilities/api.py
```diff
@@ -214,10 +220,21 @@ def to_representation(self, instance):
 
         return data
 
+    def get_references(self, vulnerability):
```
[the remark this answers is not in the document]
keshav-space: In our old API, `references` used to contain the nested `severity`, as severity and reference were related through a foreignkey relationship. In our new model we have dissociated reference and severity. To ensure we maintain compatibility for existing users of old API we're manually crafting the references to include the relevant severity.

PR #1651, opened 2024-11-13
Add `on_failure` to handle cleanup during pipeline failure

keshav-space:
- Resolves https://github.com/aboutcode-org/vulnerablecode/issues/1639


TG1999:
@keshav-space can we have some sorts of tests, to check this ? If on_failure runs if any step fails ?

keshav-space:
> can we have some sorts of tests, to check this ? If on_failure runs if any step fails ?

@TG1999 done!

PR #1675, opened 2024-11-20
Add scores in bulk search V1 API

### LINKED ISSUE #1674: Add `exploitability`, `weighted_severity`, `risk_score` to API v1 Package bulk_search
Those new fields are missing form the vulnerability data structure returned by the package bulk_search in v1.

Those fields seems available in bulk_search v2 but since DejaCode has not migrated to v2 yet we need those in v1 as well.

TG1999:
Fixes #1674


tdruez:
Changes looks fine but tests are failing.

Issue #1492, opened 2024-07-02
VCIO-next: API performance issues (vulnerabilities endpoint)


tdruez:
A simple Vulnerability details view can take over 3min to be returned, for example https://public.vulnerablecode.io/api/vulnerabilities/516832

This needs to be optimized to make the API usable.

See also these related issues:
- https://github.com/aboutcode-org/vulnerablecode/issues/1538
- https://github.com/aboutcode-org/dejacode/issues/94#issuecomment-2298445423

tdruez:
While page 1 loads in a couple of seconds: https://public.vulnerablecode.io/api/vulnerabilities

Going to page 2 or 3 get stuck and never render.

pombredanne:
FYI, the key code to get this fixed is the new API design in #1572

TG1999:
This is done now!

PRs for references:
https://github.com/aboutcode-org/vulnerablecode/pull/1701
https://github.com/aboutcode-org/vulnerablecode/pull/1631
https://github.com/aboutcode-org/vulnerablecode/issues/1561

To test this

We have a new endpoint deployed and live on https://public.vulnerablecode.io/api/v2/

/api/v2/vulnerabilities- https://public.vulnerablecode.io/api/v2/vulnerabilties

Vulnerabilities endpoint-
This endpoint has two filters
- alias - where we pass an alias and get vulnerability IDs associated with that alis
- vulnerability_id - where we pass vulnerability ID and get info for the VCID

Format:

https://public.vulnerablecode.io/api/v2/vulnerabilities

- In results, we now provide "vulnerabilities" map which has key as VCID and value in this format

```
{
                "vulnerability_id": "VCID-111c-u9bh-aaac",
                "url": "http://public.vulnerablecode.io/api/v2/vulnerabilities/VCID-111c-u9bh-aaac"
}
```
- The above "url" points to information about that vulnerability ID
- For a single Vulnerability ID http://public.vulnerablecode.io/api/v2/vulnerabilities/VCID-111c-u9bh-aaac format looks like this

```
{
    "vulnerability_id": "VCID-111c-u9bh-aaac",
    "aliases": [
        "CVE-2017-1000136"
    ],
    "summary": "Mahara 1.8 before 1.8.6 and 1.9 before 1.9.4 and 1.10 before 1.10.1 and 15.04 before 15.04.0 are vulnerable to old sessions not being invalidated after a password change.",
    "severities": [
        {
[...]

PR #1701, opened 2024-12-17
Add indexes for models

TG1999:
Reference: #1492 #1538


vulnerabilities/models.py
```diff
@@ -709,10 +714,26 @@ class Package(PackageURLMixin):
...
         default=0,
+        db_index=True,
     )
 
     objects = PackageQuerySet.as_manager()
 
+    class Meta:
+        unique_together = ["type", "namespace", "name", "version", "qualifiers", "subpath"]
+        ordering = ["type", "namespace", "name", "version_rank", "version", "qualifiers", "subpath"]
+        indexes = [
```
TG1999, line 725: Add a index for type, namespace, name, version

PR #1704, opened 2024-12-20
Add models for CodeFix

TG1999:
References:
- https://github.com/aboutcode-org/vulnerablecode/issues/207
- https://github.com/aboutcode-org/vulnerablecode/issues/1696


vulnerabilities/models.py
```diff
@@ -1581,3 +1581,32 @@ class Exploit(models.Model):
     @property
     def get_known_ransomware_campaign_use_type(self):
         return "Known" if self.known_ransomware_campaign_use else "Unknown"
+
+
+class CodeChange(models.Model):
+    commits = models.JSONField(blank=True, default=list)
```
pombredanne, line 1587: Can you add docstring for the model and help text for each field?

vulnerabilities/models.py
```diff
@@ -1581,3 +1581,32 @@ class Exploit(models.Model):
...
+    patch = models.TextField(blank=True, null=True)
+    notes = models.TextField(blank=True, null=True)
+    references = models.JSONField(blank=True, default=list)
+    status_reviewed = models.BooleanField(default=False)
+    base_version = models.ForeignKey(
+        "Package",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="base_version_changes",
```
pombredanne, line 1599: base_version_changes or base_version_codechange?

vulnerabilities/models.py
```diff
@@ -1581,3 +1581,32 @@ class Exploit(models.Model):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        abstract = True
+
+
+class CodeFix(CodeChange):
+    vulnerabilities = models.ManyToManyField("Vulnerability", related_name="codefixes", blank=True)
+    applies_to_versions = models.ManyToManyField("Package", related_name="fixes", blank=True)
```
pombredanne, line 1612: Can you elaborate on what applies_to_versions means ?
Also I wonder if it would not better to reuse the "AffectedByPackageRelatedVulnerability" relation?

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+        return (cls.collect_and_store_fix_commits,)
+
+    def collect_and_store_fix_commits(self):
+        references = VulnerabilityReference.objects.prefetch_related("vulnerabilities").distinct()
+
+        self.log(f"Processing {references.count():,d} references to collect fix commits.")
+
+        created_fix_count = 0
+        progress = LoopProgress(total_iterations=references.count(), logger=self.log)
+        for reference in progress.iter(references.paginated(per_page=500)):
```
pombredanne, line 38: How can we avoid to reprocess references that were already processed?

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+            for vulnerability in reference.vulnerabilities.all():
+                package_urls = self.extract_package_urls(reference)
+                commit_id = self.extract_commit_id(reference.url)
+
+                if commit_id and package_urls:
+                    for purl in package_urls:
+                        normalized_purl = normalize_purl(purl)
+                        package = self.get_or_create_package(normalized_purl)
```
pombredanne, line 46: Do we really want to create new packages here? I would have thought we are only adding new CodeFix?

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+                        codefix = self.create_codefix_entry(
+                            vulnerability=vulnerability,
+                            package=package,
+                            commit_id=commit_id,
```
pombredanne, line 50: Can craft a VCS URL instance instead? If we have a commit we need to have a base repo too, but with a vcs_url, we can combine this is one.

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+                    "references": [reference],
+                },
+            )
+            if created:
+                codefix.vulnerabilities.add(vulnerability)
+                codefix.save()
+            return codefix
+        except Exception as e:
+            self.log(f"Error creating CodeFix entry: {e}")
+            return None
```
pombredanne, line 112:
```suggestion
            return
```

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+            parts = reference.url.split("/")
+            if len(parts) >= 5:
+                namespace = parts[-3]
+                name = parts[-2]
+                commit = parts[-1]
+                if commit:
+                    urls.append(f"pkg:github/{namespace}/{name}@{commit}")
+        return urls
+
+    def extract_commit_id(self, url):
```
pombredanne, line 74: Can you extract that in a separate top level function so we can test all different variations?

vulnerabilities/pipelines/collect_commits.py
```diff
@@ -0,0 +1,112 @@
...
+                            package=package,
+                            commit_id=commit_id,
+                            reference=reference.url,
+                        )
+                        if codefix:
+                            created_fix_count += 1
+
+        self.log(f"Successfully created {created_fix_count:,d} CodeFix entries.")
+
+    def extract_package_urls(self, reference):
```
pombredanne, line 58: ah, I see you are creating a PURL... I wonder more simply if we could instead create a vcs url as in https://github.com/aboutcode-org/scancode-toolkit/blob/c40476ab1dcf91fd92f4b2eb5ee64ca4a6aa1002/src/packagedcode/utils.py#L42 ?

pombredanne (commented):
Thanks! Here are a few comments for your consideration.
IMHO I am not sure we want a PURL for a commit, we should consider using vcs URLs rather than a plain commit id... 

I am open to something else that we can design, but if we do that we need a way to get back to the commit.

We are going to have cases that are NOT on GitHub and therefore are just plain git coordinates.

TG1999:
Screenshot from 2025-01-08 20-21-06
 
Screenshot from 2025-01-08 20-21-28
API support is also added!

keshav-space (approved):
Thanks @TG1999, LGTM!

PR #1706, opened 2025-01-02
Add fixed by package in V2 API

TG1999:
Reference: https://github.com/aboutcode-org/vulnerablecode/issues/1665

### DIFF
--- a/vulnerabilities/api_v2.py
+++ b/vulnerabilities/api_v2.py
@@ -195,7 +195,21 @@ class PackageV2Serializer(serializers.ModelSerializer):
         ]
 
     def get_affected_by_vulnerabilities(self, obj):
-        return [vuln.vulnerability_id for vuln in obj.affected_by_vulnerabilities.all()]
+        """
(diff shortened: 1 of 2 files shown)


vulnerabilities/api_v2.py
```diff
@@ -195,7 +195,21 @@ class Meta:
         ]
 
     def get_affected_by_vulnerabilities(self, obj):
-        return [vuln.vulnerability_id for vuln in obj.affected_by_vulnerabilities.all()]
+        """
+        Return a dictionary with vulnerabilities as keys and their details, including fixed_by_packages.
+        """
+        vulnerabilities = obj.affected_by_vulnerabilities.prefetch_related("fixed_by_packages")
```
keshav-space, line 201: Shouldn't we prefetch `fixed_by_packages` in the viewset itself? I don't think it's a good idea to trigger a fresh query in the serializer, as this would undo query optimization already done in the viewset.

keshav-space (changes requested):
Thanks @TG1999, minor nit for your consideration.

Issue #586, opened 2021-12-19
Track Apache Log4j advisories 


pombredanne:
See https://logging.apache.org/log4j/2.x/security.html
It graduated to be tracked on its own.

TG1999:
It's in CycloneDX json format https://logging.apache.org/log4j/2.x/security.html#vulnerabilities

PR #1744, opened 2025-01-15
Add Apache Log4j Advisories

NucleonGodX:
This pull request addresses issue #586  by adding an importer for Apache Log4j advisories
image

### DIFF
--- a/vulnerabilities/pipelines/apache_log4j_importer.py
+++ b/vulnerabilities/pipelines/apache_log4j_importer.py
@@ -19,11 +19,14 @@ from defusedxml import ElementTree as SafeElementTree
 from packageurl import PackageURL
 from univers.versions import MavenVersion
 
+from vulnerabilities import severity_systems
 from vulnerabilities.importer import AdvisoryData
 from vulnerabilities.importer import AffectedPackage
 from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
 from vulnerabilities.pipelines import VulnerableCodeBaseImporterPipeline
 from vulnerabilities.utils import fetch_response
+from vulnerabilities.utils import get_cwe_id
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


TG1999:
@NucleonGodX thanks for your contributions, can you please use pipeline structure. https://github.com/aboutcode-org/vulnerablecode/blob/main/vulnerabilities/pipelines/nvd_importer.py check this for example

vulnerabilities/importers/apache_log4j.py
```diff
@@ -0,0 +1,252 @@
+import logging
```
pombredanne, line 1: Use the same header as in other files

vulnerabilities/importers/apache_log4j.py
```diff
@@ -0,0 +1,252 @@
...
+import xml.etree.ElementTree as ET
```
pombredanne, line 2: Consider using the cyclonedx python library to parse this, and not parsing XML yourself

vulnerabilities/importers/apache_log4j.py
```diff
@@ -0,0 +1,252 @@
...
+        except ValueError as e:
+            logger.error(f"Error getting versions in range: {e}")
+            return []
+
+    def to_advisory(self, xml_content):
+        """
+        Parse the XML content and create AdvisoryData objects.
+        """
+        advisories = []
+        version_mapping = [
```
pombredanne, line 71: Why use mapping? are you positively sure that these are forever all the known versions of log4j? Also this would not be a mapping, rather just a set of some versions?

vulnerabilities/importers/apache_log4j.py
```diff
@@ -0,0 +1,252 @@
...
+            "2.14.0",
+            "2.14.1",
+            "2.15.0",
+            "2.15.1",
+            "2.16.0",
+            "2.17.0",
+            "2.17.1",
+        ]
+
+        try:
```
pombredanne, line 126: Why a try/except here?

vulnerabilities/importers/apache_log4j.py
```diff
@@ -0,0 +1,252 @@
...
+            root = ET.fromstring(xml_content)
```
pombredanne, line 127: Use the cyclonedx parser.

vulnerabilities/improvers/valid_versions.py
```diff
@@ -481,3 +482,8 @@ class GithubOSVImprover(ValidVersionImprover):
 class CurlImprover(ValidVersionImprover):
     importer = CurlImporter
     ignorable_versions = []
+
+
+class Log4jImprover(ValidVersionImprover):
+    importer = ApacheLog4jImporter
+    ignorable_versions = []
```
pombredanne, line 489: We likely need some improver to get and expand the list of known versions

pombredanne (changes requested):
Thanks! In addition to use the new pipelines, here are a few comments for your consideration. We also need some tests.

vulnerabilities/importers/apache_log4j.py
NucleonGodX: I added them from https://www.cvedetails.com/version-list/45/37215/1/Apache-Log4j.html?order=0.

vulnerabilities/pipelines/apache_log4j_importer.py
```diff
@@ -0,0 +1,320 @@
...
+            for version_range in vuln_target.versions:
+                if version_range.version and not version_range.range:
+                    self._process_single_version(version_range, fixed_versions, version_groups)
+                elif version_range.range:
+                    self._process_version_range(version_range, fixed_versions, version_groups)
+
+        affected_packages = []
+        for fixed_version, versions in version_groups.items():
+            unique_versions = sorted(set(versions), key=lambda x: MavenVersion(x))
+            version_range_str = f"vers:maven/{('|'.join(unique_versions))}"
```
keshav-space, line 232: Why not use `MavenVersionRange` instead?

NucleonGodX: oh yeah, I'll try to use that.

Issue #517, opened 2021-07-28
Add tests for Docker


Hritik14:
We're shipping Docker containers which differ with our current test environment (for eg, test environment runs on `ubuntu-latest` but the Docker container runs on PYTHON image which is based on Debian). 
Due to divergence like these, we need to have separate tests for Dockerfile.

vuld0:
Hi, I'm interested in working on this issue. Is it still open and is it okay if I take a stab at it? If so, could you please provide some guidance on how I should proceed? Thanks!

aditi25srivastava:
Hi! I would like to work on this issue. Please assign it to me.

Hritik14:
@aditi25srivastava being assigned is not a necessity to work on an issue in general :)

aditi25srivastava:
Thank you for the clarification!
I’ll start working on the issue and open a PR soon.
Please let me know if there are any specific requirements you'd like me to follow.

Drag0nSlay:
Hi,
I have created a PR for this issue: https://github.com/aboutcode-org/vulnerablecode/pull/2115
The PR adds Docker-based CI tests along with the necessary test fixtures.

shivamshrma09:
Hi @Hritik14

want to fix
**#517  (Docker tests)** 

Know Django, local Docker running. 
PR ready in 2 day

PR #1746, opened 2025-01-15
Added tests for docker

unibik:
issue #517 Added separate tests for docker .  
Now both test environment and docker containers runs on ubuntu-latest

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,7 @@ coreapi==2.3.3
 coreschema==0.0.4
 cryptography==43.0.1
 crispy-bootstrap4==2024.1
+cvss
 cwe2==3.0.0
 dateparser==1.1.1
 decorator==5.1.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -30,6 +31,7 @@ distro==1.7.0
 Django==4.2.17
 django-crispy-forms==2.3
 django-environ==0.11.2
+django-extensions==3.2.3
```
TG1999, line 34: Why this ?

TG1999:
@unibik please explain the changes in requirements.txt

unibik:
Hello, @TG1999 While testing the Docker-based CI pipeline locally using the **act tool**, I encountered a **ModuleNotFoundError** for some dependencies like cvss, django-extensions, and htmllistparse. These modules were used in the **codebase** but weren’t listed in the **requirements.txt** file. 
As the docker tests use the command **docker compose build --no-cache**, which will create a new image which doesn't include the modules present in the host or owners machine. So I explicitly added those module.
Thank you for reviewing my PR

PR #1764, opened 2025-01-27
Add pipeline to compute Advisory ToDos 

keshav-space:
- Resolves https://github.com/aboutcode-org/vulnerablecode/issues/1699
- Blocker https://github.com/aboutcode-org/vulnerablecode/issues/1583, also at minimum, we need to get rid of aliases JSON field in the advisory and instead use an M2M relation.


TG1999:
https://github.com/aboutcode-org/vulnerablecode/pull/1863 Related

PR #1766, opened 2025-01-29
Add new content ID function

TG1999:
Reference: https://github.com/aboutcode-org/vulnerablecode/issues/1583


vulnerabilities/utils.py
```diff
@@ -536,3 +537,43 @@ def normalize_purl(purl: Union[PackageURL, str]):
...
+        return sorted(lst) if lst else []
+
+    def normalize_dict(obj):
+        """Ensure dictionary keys are ordered."""
+        return json.loads(json.dumps(obj, sort_keys=True)) if obj else {}
+
+    # Normalize fields
+    normalized_data = {
+        "summary": normalize_text(advisory_data.summary),
+        "affected_packages": normalize_list(advisory_data.affected_packages),
```
keshav-space, line 567: There is no certainty that this will work since we don't have a proper implementation for handling comparisons between `AffectedPackage`.

For example, this will not be able to normalize the list of affected_packages below:
```python
affected_packages = [
    AffectedPackage(
        package=PackageURL(
            type="alpine",
            namespace=None,
            name="linux-lts",
            version=None,
            qualifiers={
                "arch": "aarch64",
                "distroversion": "v3.20",
                "reponame": "main",
            },
            subpath=None,
        ),
        affected_version_range=None,
        fixed_version="6.6.13-r1",
    ),
    AffectedPackage(
        package=PackageURL(
            type="alpine",
            namespace=None,
            name="linux-lts",
            version=None,
            qualifiers={"arch": "armhf", "distroversion": "v3.20", "reponame": "main"},
            subpath=None,
        ),
        affected_version_range=None,
        fixed_version="6.6.13-r1",
    ),
]

```

vulnerabilities/utils.py
```diff
@@ -536,3 +537,43 @@ def normalize_purl(purl: Union[PackageURL, str]):
...
+        "references": normalize_list(advisory_data.references),
+        "weaknesses": normalize_list(advisory_data.weaknesses),
```
keshav-space, line 569: Similar to the above `affected_packages` list, this also won't work.

vulnerabilities/utils.py
```diff
@@ -536,3 +537,43 @@ def normalize_purl(purl: Union[PackageURL, str]):
...
+    }
+
+    if include_metadata:
+        normalized_data["created_by"] = advisory_data.created_by
```
keshav-space, line 573: `created_by` is a model field, not an attribute of AdvisoryData.

keshav-space (changes requested):
Thanks @TG1999, see some suggestions below.

vulnerabilities/models.py
```diff
@@ -1369,6 +1370,11 @@ def save(self, *args, **kwargs):
         self.unique_content_id = checksum.hexdigest()
         super().save(*args, **kwargs)
 
+    def save(self, *args, **kwargs):
+        advisory_data = self.to_advisory_data()
+        self.unique_content_id = compute_content_id(advisory_data, include_metadata=False)[:31]
```
keshav-space, line 1375: Why trim the hash to 32 characters when we're increasing the max_length for `unique_content_id` to 64?

```suggestion
        self.unique_content_id = compute_content_id(advisory_data, include_metadata=False)
```

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Find advisories with the same content and keep only the latest one.
+        """
+        # Get all advisories that have duplicates based on content ID
+        duplicate_content_ids = (
+            Advisory.objects.values("unique_content_id")
+            .annotate(count=Count("id"))
+            .filter(count__gt=1)
+            .values_list("unique_content_id", flat=True)
+        )
```
keshav-space, line 41: There are potentially millions of duplicate advisories, we should paginate them instead of loading them all at once into memory.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.log(
+            f"Found {len(duplicate_content_ids)} content IDs with duplicates", level=logging.INFO
+        )
+
+        for content_id in duplicate_content_ids:
```
keshav-space, line 47: We should use LoopProgress.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+                    f"{advisories.count() - 1} duplicates for content ID {content_id}",
+                    level=logging.INFO,
+                )
+
+    def update_content_ids(self):
+        """
+        Update content IDs for all advisories that don't have one.
+        """
+        advisories = Advisory.objects.filter(
+            Q(unique_content_id="") | Q(unique_content_id__isnull=True)
```
keshav-space, line 69: Since we have changed how we compute content id and also we're now using md5 insted sha256, we need to recompute content id for all advisories irrespective of whether or not the field is empty.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+        self.log(f"Found {advisories.count()} advisories without content ID", level=logging.INFO)
+
+        for advisory in advisories:
+            advisory.unique_content_id = compute_content_id(advisory)
+            advisory.save()
```
keshav-space, line 76: Calling save individually on millions of advisory is not right. We should insted do bulk update.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+
+            if self.log:
+                self.log(f"Updated content ID for advisory {advisory.id}", level=logging.DEBUG)
```
keshav-space, line 79: Maybe we can do something simpler like this.

```suggestion
        """
        Recompute content id and remove advisories with the same content and keep only the latest one.
        """

        advisories = Advisory.objects.all().order_by("-id").paginated()
        advisories_count = Advisory.objects.all().count()

        self.log(f"Computing new content id for {advisories_count} and removing duplicates.")

        batch_size = 10000
        deleted_advisory_count = 0
        updated_advisory_count = 0
        duplicate_advisory_id = []
        updated_advisory = []
        content_ids = set()
        progress = LoopProgress(
            total_iterations=advisories_count,
            logger=self.log,
            progress_step=1,
        )
        for advisory in progress.iter(advisories):
            content_id = compute_content_id(advisory)
            if content_id in content_ids:
                duplicate_advisory_id.append(advisory.id)
            else:
                advisory.unique_content_id = content_id
                updated_advisory.append(advisory)

            if len(duplicate_advisory_id) > batch_size:
                deleted_advisory_count += delete_advisories(
                    advisory_ids=duplicate_advisory_id,
                    logger=self.log,
                )

            if len(updated_advisory) > batch_size:
                updated_advisory_count += bulk_update_advisory(
                    items=updated_advisory,
                    fields=["unique_content_id"],
                    logger=self.log,
                )

        deleted_advisory_count += delete_advisories(
            advisory_ids=duplicate_advisory_id,
            logger=self.log,
        )

        updated_advisory_count += bulk_update_advisory(
            items=updated_advisory,
            fields=["unique_content_id"],
            logger=self.log,
        )

        self.log(f"Removed {deleted_advisory_count} duplicates advisories.")
        self.log(f"Updated content id for {deleted_advisory_count} advisories.")


def bulk_update_advisory(items, fields, logger):
    item_count = 0
    if items:
        try:
            Advisory.objects.bulk_update(objs=items, fields=fields)
            item_count += len(items)
        except Exception as e:
            logger(f"Error updating Advisory: {e}")
        items.clear()
    return item_count


def delete_advisories(advisory_ids, logger):
    item_count = 0
    if advisory_ids:
        try:
            Advisory.objects.filter(id__in=advisory_ids).delete()
            item_count += len(advisory_ids)
        except Exception as e:
            logger(f"Error deleting Advisory: {e}")
        advisory_ids.clear()
    return item_count
```

keshav-space (changes requested):
Thanks @TG1999, some nits for your consideration.

vulnerabilities/importer.py
```diff
@@ -55,15 +57,23 @@ class VulnerabilitySeverity:
...
+        return data
+
+    def __lt__(self, other):
+        if not isinstance(other, VulnerabilitySeverity):
+            return NotImplemented
+        return self._cmp_key() < other._cmp_key()
+
+    # TODO: Add cache
+    def _cmp_key(self):
+        return (str(self.system), self.value, self.scoring_elements, self.published_at)
```
pombredanne, line 76: This is less costly than getting a string on the whole object.
```suggestion
        return (self.system.identifier, self.value, self.scoring_elements, self.published_at)
```

vulnerabilities/importer.py
```diff
@@ -170,6 +183,15 @@ def get_fixed_purl(self):
...
         return update_purl_version(purl=self.package, version=str(self.fixed_version))
 
+    def __lt__(self, other):
+        if not isinstance(other, AffectedPackage):
+            return NotImplemented
+        return self._cmp_key() < other._cmp_key()
+
+    # TODO: Add cache
+    def _cmp_key(self):
+        return (str(self.package), str(self.affected_version_range), str(self.fixed_version))
```
pombredanne, line 193: If there is no fixed_version or no affected_version_range, their value would be `None` and the string value `"None"` which is not what we want IMHO, but rather `""`
```suggestion
        return (str(self.package), str(self.affected_version_range or ""), str(self.fixed_version or ""))
```

vulnerabilities/models.py
```diff
@@ -1315,7 +1316,7 @@ class Advisory(models.Model):
     """
 
     unique_content_id = models.CharField(
-        max_length=32,
+        max_length=64,
```
pombredanne, line 1319: We need a help text and including an explanation of why we use 64 chars, e.h. sha256 as hex.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,79 @@
...
+            # Get all advisories with this content ID
+            advisories = Advisory.objects.filter(unique_content_id=content_id)
+
+            # Find the latest advisory
+            latest = advisories.latest("date_imported")
+
+            # Delete all except the latest
+            advisories.exclude(id=latest.id).delete()
+
+            if self.log:
```
pombredanne, line 57: Why this test?

pombredanne (changes requested):
Here is some more feedback.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def remove_duplicates(self):
+        """
+        Find advisories with the same content and keep only the latest one.
+        """
+
+        duplicated_advisories = groupby(
+            Advisory.objects.order_by("unique_content_id").all().paginated(),
+            key=lambda x: x.unique_content_id,
+        )
```
keshav-space, line 42: I honestly doubt that this will work smoothly in production, ordering 118 million advisories using `unique_content_id`, which is an unindexed field is not good idea.

https://github.com/aboutcode-org/vulnerablecode/pull/1766#discussion_r1946776376 would be much more practical approach since `id` is an autogenerated primary key field and hence it's already indexed, and we can use this to select the latest or oldest advisory.

vulnerabilities/pipelines/remove_duplicate_advisories.py
```diff
@@ -0,0 +1,83 @@
...
+            total_iterations=Advisory.objects.count(),
+            progress_step=1,
+            logger=self.log,
+        )
+
+        for advisory in progress.iter(Advisory.objects.all().paginated()):
+            advisory.unique_content_id = compute_content_id(advisory)
+            advisories.append(advisory)
+
+        Advisory.objects.bulk_update(advisories, ["unique_content_id"], batch_size=1000)
```
keshav-space, line 83: Keeping all that `advisories` in memory is going to be very expensive!
Assuming 1 Advisory object ≈ 1000 bytes.
That would mean for 118 million advisories: (118 * 10^6) * 10^3 bytes ≈ 118 GB of memory!.

We should not keep all these advisories in memory. Instead we should bulk update as soon as we reach the batch size.

vulnerabilities/utils.py
```diff
@@ -546,3 +547,56 @@ def get_purl_version_class(purl):
...
+    :param include_metadata: Boolean indicating whether to include `created_by` and `url`
+    :return: SHA-256 hash digest as content_id
+    """
+
+    # Normalize fields
+    from vulnerabilities.importer import AdvisoryData
+    from vulnerabilities.models import Advisory
+
+    if isinstance(advisory_data, Advisory):
+        normalized_data = {
```
keshav-space, line 576: Content id should also include `aliases` field.

Let's look at this data source https://github.com/nodejs/security-wg/blob/75c78bbbd2ef86d289c16818bb487a70e315dc43/vuln/npm/7.json. This single advisory contains 2 CVEs, and for this we create 2 different advisory records. If we do not consider aliases while computing the content ID, then we will delete the one advisory in our dedupe pipeline.

keshav-space (changes requested):
Thanks @TG1999, some feedbacks.

vulnerabilities/pipelines/remove_duplicate_advisories.py
TG1999: db_index is in models for that! And since we believe there are lots of dupes. I believe index to work fine

keshav-space: We're adding this `db_index` now. I am not sure how long the index creation itself is going to take.

Issue #513, opened 2021-07-27
Add Docker labels


Hritik14:
Decide on Docker labels we will use to be consistent with ScancodeIO and other tools we ship.

S-i-y-a-g:
Hello @Hritik14, @pombredanne , I’d like to work on this issue if it’s still open.
I’ll first go through the existing Docker labels used in the project and related tools, and then propose a consistent set of labels. Please let me know if I can take this up. Thanks!

Mahaboobunnisa123:
hey @S-i-y-a-g, already a PR created for this issue. I think we need to go for other issues to not to create duplicate work. Happy collaborating!

PR #1769, opened 2025-02-01
Add dockerfile labels #513

unibik:
Added labels in Dockerfile


unibik:
hello @Hritik14 , Is that information enough for labels. If yes , then can you please review it once...Thank you

PR #1776, opened 2025-02-10
Add initial support for the VulnerableCode agent

ziadhany:
The VulnerableCode agent currently focuses on one main task: extracting the correct version range from the vulnerability summary.

image

Screenshot from 2025-02-10 07-07-49

Screenshot from 2025-02-10 06-57-58

### DIFF
--- a/agent/admin.py
+++ b/agent/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
--- a/agent/apps.py
+++ b/agent/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class AgentConfig(AppConfig):
+    default_auto_field = "django.db.models.BigAutoField"
+    name = "agent"
--- a/agent/forms.py
+++ b/agent/forms.py
@@ -1,0 +1,10 @@
+from django import forms
+
+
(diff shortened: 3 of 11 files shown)


ziadhany:
@pombredanne, this is an initial base for the AI summary improver:

Right now, we have two prompts—one to extract the purl and another to get the affected_versions and fixed_versions—without using RAG.

I think I should also feed the model with `agent/purl_db/PURL.rst` so it can generate more accurate results. I have already implemented the basics of this step.

However, I encountered a small issue related to testing and evaluating our improver because the model sometimes returns a different output each time.

How should we approach testing it?

There’s just a little work left, and I think this improver will be ready soon.

Input 
Summary:
```
Off-by-one error in the apr_brigade_vprintf function in Apache APR-util before 1.3.5
              on big-endian platforms allows remote attackers to obtain sensitive information or cause a
              denial of service (application crash) via crafted input.
```

Output:

```
purl: pkg:apache/apr-util@<1.3.5
{
    "affected_versions": ["< 1.3.5"],
    "fixed_versions": [">= 1.3.5"]
}

```

keshav-space:
Thanks @ziadhany, This is experimental, let's use https://github.com/aboutcode-org/vulnerablecode-ai-experiments for VulnerableCode agent.

Issue #1576, opened 2024-08-30
Update the Dockerfile and release script to use Python 3.12


keshav-space:
Also run CI on Python 3.12

Issue #1782, opened 2025-02-19
Python version 3.12 and 3.13 do not support "ImpImporter"


Dedsec0098:
The Newer version of Python do not support "ImpImporter" while make the installation. Running command "make test" will give thi following  error message :

 register_finder(pkgutil.ImpImporter, find_on_path)
                    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?

The newer versions has removed "ImpImporter" from its "pkgutil" module.
I would like to suggest to mention about it in the  [Installation docs](https://vulnerablecode.readthedocs.io/en/latest/installation.html) to depreciate the version to 3.11 so that all the commands can run.

Dedsec0098:
@pombredanne If I am not wrong and this issue is there for everyone then I can update the docs requirements section with the appropriate details.

PR #1790, opened 2025-03-01
Add support for Python3.12

kunalsz:
In reference to the issue #1576 

## Changes made
- Python3.12 has been added to the Dockerfile,Makefile and CI Workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11","3.12"]
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


Rishi-source:
Hi @kunalsz ,I think Directly adding support for python 3.12 would not do the job coz some libraries or their version are not compatible in python 3.12 as mentioned by @Dedsec0098 in one of his issue related to failing tests #1782 which means that some tests would be failing in 3.12. so you have to take a look on all of such issues which could come when workflow file runs on each pull request otherwise currently your code is cancelling the job as you can see in the image below.
jobs for python 3.9,3.10.3.11 are also giving error 

```
Run make dev envfile
-> Bootstrap the virtualenv with PYTHON_EXE=python3.12
make: python3.12: Command not found
make: *** [Makefile:51: virtualenv] Error [12](https://github.com/kunalsz/vulnerablecode/actions/runs/13607460788/job/38041005157#step:5:13)7

```
PYTHON_EXE should not be changed imo as I am working on python 3.13 still I use python3 while running any python command on cli.
Screenshot 2025-03-02 at 9 21 56 AM

Issue #1555, opened 2024-08-16
Do not use Semver for Apache Tomcat Importer


TG1999:
./manage.py import vulnerabilities.importers.apache_tomcat.ApacheTomcatImporter
Importing data using vulnerabilities.importers.apache_tomcat.ApacheTomcatImporter
'M5' is not a valid SemverVersion InvalidVersion("'M5' is not a valid <class 'univers.versions.SemverVersion'>")
'M1' is not a valid SemverVersion InvalidVersion("'M1' is not a valid <class 'univers.versions.SemverVersion'>")
'RC1' is not a valid SemverVersion InvalidVersion("'RC1' is not a valid <class 'univers.versions.SemverVersion'>")
Successfully imported data using vulnerabilities.importers.apache_tomcat.ApacheTomcatImporter

kunalsz:
Hi @TG1999 , Maven versioning is better suitable for handling Apache Tomcat. Hence,I will be removing the Semver versioning.

PR #1791, opened 2025-03-01
Fixed Apache Tomcat Importer by not using Semver

kunalsz:
In reference to the issue #1555 

## Changes made:
- Usage of `Semver` versioning is removed which was creating problems
- Only `Maven` versioning is being used

## Apache Tomcat Versioning Overview
Apache Tomcat versions follow a Maven-like versioning scheme, which can include:
- Milestone versions (e.g., 9.0.0.M1)
- Release candidates (e.g., 8.0.0.RC1)
- Regular numeric versions (e.g., 8.5.32)

This is not Semver-compliant, but Maven versioning handles all of these cases

@TG1999 looking forward to your insights !

### DIFF
--- a/vulnerabilities/importers/apache_tomcat.py
+++ b/vulnerabilities/importers/apache_tomcat.py
@@ -16,10 +16,8 @@ import requests
 from bs4 import BeautifulSoup
 from packageurl import PackageURL
 from univers.version_constraint import VersionConstraint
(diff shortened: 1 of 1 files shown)


keshav-space (changes requested):
@kunalsz IMO this is not the right approach instead we should fixed the version_class in upstream univers `ApacheVersionRange`.

Issue #1515, opened 2024-07-20
Add Apache Camel advisories


pombredanne:
See https://camel.apache.org/security/

PR #1792, opened 2025-03-02
Added Apache Camel Pipeline

kunalsz:
In reference to the issue #1515 

## Changes Made:
- Apache Camel Advisory Pipeline added
- Test written

### DIFF
--- a/vulnerabilities/pipelines/apache_camel_importer.py
+++ b/vulnerabilities/pipelines/apache_camel_importer.py
@@ -1,0 +1,222 @@
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# VulnerableCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/aboutcode-org/vulnerablecode for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+import logging
+import re
+from datetime import datetime
+from datetime import timezone
+from typing import Iterable
+from typing import Tuple
+
+import requests
+from bs4 import BeautifulSoup
+from packageurl import PackageURL
+from univers.version_constraint import VersionConstraint
+from univers.version_range import MavenVersionRange
+from univers.version_range import VersionRange
+from univers.versions import GenericVersion
+from univers.versions import MavenVersion
+
+from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackage
+from vulnerabilities.importer import Reference
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.pipelines import VulnerableCodeBaseImporterPipeline
+from vulnerabilities.severity_systems import GENERIC
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+from vulnerabilities.utils import fetch_response
+from vulnerabilities.utils import get_item
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class ApacheCamelImporterPipeline(VulnerableCodeBaseImporterPipeline):
+    """Collect Advisories from Apache Camel"""
+
+    pipeline_id = "apache_camel_importer"
+    spdx_license_expression = "Apache-2.0"
+    license_url = "https://www.apache.org/licenses/LICENSE-2.0"
+    root_url = "https://camel.apache.org/security/"
+    importer_name = "Apache Camel Importter"
+
+    def __init__(self):
+        super().__init__()
+        self.raw_data = None
+
+    @classmethod
+    def steps(cls):
+        return (
+            cls.fetch_html_response,
+            cls.collect_and_store_advisories,
+            cls.import_new_advisories,
+        )
+
+    # fetch the html content and saves in raw_data
+    def fetch_html_response(self):
+        try:
+            response = fetch_response(self.root_url).content
+            self.raw_data = BeautifulSoup(response, "html.parser")
+        except:
+            logger.error(f"Failed to fetch URL {self.root_url}")
+
+    # num of advisories
+    def advisories_count(self) -> int:
+        return fetch_count_advisories(self.raw_data)
+
+    # parse the response data
+    def collect_advisories(self) -> Iterable[AdvisoryData]:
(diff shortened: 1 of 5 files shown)


vulnerabilities/pipelines/apache_camel_importer.py
```diff
@@ -0,0 +1,343 @@
...
+    """Collect Advisories from Apache Camel"""
+
+    pipeline_id = "apache_camel_importer"
+    spdx_license_expression = "Apache-2.0"
+    license_url = "https://www.apache.org/licenses/LICENSE-2.0"
+    root_url = "https://camel.apache.org/security/"
+    importer_name = "Apache Camel Importter"
+
+    def __init__(self):
+        super().__init__()
```
keshav-space, line 45: Any reason why we need this? I'm curious, is there something peculiar about the Apache Camel importer that requires it here, but not in the OpenSSL importer here https://github.com/aboutcode-org/vulnerablecode/blob/e2f03f59c276b5ca9554153f2b06da49ead0a889/vulnerabilities/pipelines/openssl_importer.py

vulnerabilities/pipelines/apache_camel_importer.py
```diff
@@ -0,0 +1,343 @@
...
+    """
+
+    soup = fetch_html_response(url)
+    table = soup.find("tbody")
+    advisory_len = len(table.find_all("tr"))
+
+    return advisory_len
+
+
+def fetch_advisory_data(url):
```
keshav-space, line 123: It would be much cleaner and easier to parse advisory from text data. See example https://camel.apache.org/security/CVE-2023-34442.txt.asc

vulnerabilities/pipelines/apache_camel_importer.py
```diff
@@ -0,0 +1,343 @@
...
+        'camel'
+    """
+
+    alias = get_item(raw_data, "Reference")
+
+    version_pattern = re.compile(r"\b\d+\.\d+\.\d+\b")
+    fixed_version_out = get_item(raw_data, "Fixed")
+    fixed_versions = []
+    for fixed_version in version_pattern.findall(fixed_version_out):
+        fixed_versions.append(MavenVersion(fixed_version))
```
keshav-space, line 249: Strange, you’re collecting the fixed versions but not including them in the final `AdvisoryData`.

vulnerabilities/pipelines/apache_camel_importer.py
```diff
@@ -0,0 +1,343 @@
...
+
+    affected_packages = []
+    affected_package_string = get_item(raw_data, "Affected")
+    affected_package = parse_apache_camel_versions(affected_package_string)
+    affected_packages.append(
+        AffectedPackage(
+            package=PackageURL(
+                type="maven",
+                namespace="org.apache.camel",
+                name="camel",
```
keshav-space, line 259: Component is not always `camel`. For example, in this advisory https://camel.apache.org/security/CVE-2023-34442.html component is `camel-jira`.

vulnerabilities/pipelines/apache_camel_importer.py
```diff
@@ -0,0 +1,343 @@
...
+def fetch_date_published(cve):
+    """Fetches Date of a CVE"""
+
+    url = f"https://cveawg.mitre.org/api/cve/{cve}"
+    response = fetch_response(url).content
+    response = json.loads(response)
+    return response["cveMetadata"]["datePublished"]
+
+
+def parse_apache_camel_versions(version_string):
```
keshav-space, line 301: This parse function doesn't work correctly. For example, given the affected package statement `2.15.0 up to 2.15.4, 2.16.0` function will produces `vers:maven/2.15.0|2.15.4|2.16.0` which is not correct version range.

keshav-space (changes requested):
Thanks @kunalsz, see some suggestions and feedback.

kunalsz:
@keshav-space Thanks for the review, I'll make the necessary changes and update you.

vulnerabilities/pipelines/apache_camel_importer.py
kunalsz: Currently the advisories are fetched by just parsing the html table. Which keeps things cleaner as I have to just parse the fixed rows and columns. Parsing the text data would involve using regex to find the correct line hence making things more tedious

vulnerabilities/pipelines/apache_camel_importer.py
kunalsz: @keshav-space  Yes,you are right about that. There are various components for `camel` but they are not specified in an easier to parse fashion atleast on the advisory page/text file atleast. For example

- In https://camel.apache.org/security/CVE-2020-11973.html
`Apache Camel Netty` is the component, but it is hard to parse instead there's a link available on the advisory page https://issues.apache.org/jira/browse/CAMEL-14477 which exclusively specifies the component as `camel-netty`.

Same goes for other advisories also, the component is more exclusively defined on https://issues.apache.org/jira/browse/CAMEL-{ID}. So, should I parse the component from there ?

vulnerabilities/pipelines/apache_camel_importer.py
kunalsz: I'll update the parsing logic

vulnerabilities/pipelines/apache_camel_importer.py
kunalsz: I must have neglected it by mistake, I'll include that too

kunalsz:
@keshav-space I have update the version parsing logic and it works fine. But I have a doubt on how to incorporate `fixed_version` in the `affected_package`

For example:
- For the affected version string `2.22.x, 2.23.x, 2.24.x, 2.25.x, 3.0.0 up to 3.1.0` the fixed version string is `3.2.0`
- affected string `3.0.0 up to 3.14.5, and 3.15.0 up to 3.18.3, and 3.19.0.` fixed string : `3.14.6, 3.18.4` . Here what should be the fixed version for `3.19.0` ? Or should I leave it empty ?
- Similary in `2.22.x, 2.23.x, 2.24.x, 2.25.0 and 2.25.1, 3.0.0 up to 3.3.0` for which fixed string is `2.25.2, 3.4.0`. 
- And lastly the latest affected strings are like `Apache Camel 4.10.0 before 4.10.3. Apache Camel 4.8.0 before 4.8.6.` for which fixed strings are given in reverse making it tedious to get the fixed version `4.8.6 and 4.10.3`

Looking forward to your insights !

PR #1816, opened 2025-03-17
Fix : Improved Severity table of the Essentials screen

### LINKED ISSUE #1732: Improve display of the Severity table of the Essentials screen
The Severity table on the Essentials screen is becoming very hard to read with so much data repeated at a row level.
One idea is to add a separate column for each System - Score combination and/or creating System subsections.
Using https://public.vulnerablecode.io/vulnerabilities/VCID-29te-f3et-aaac as an example:
- There are many pairs of numeric and text scores from the same origin (URL). Since the text score is effectively an alias in these cases it would be better to see the cvss.31 and generic_textual values in separate columns on a single row for that origin. This is an extremely common pattern
- With the increase in cvss versions and other systems it might also makes sense to have separate sections for cvss vs. epss especially because these scores are not comparable.

For this particular VCID there are also 15 instances (14 duplicates) for the epss Score

Dedsec0098:
Fix #1732 

**Changes Made :**

1) CVSS Table : Added proper grouping by source/reference

2) Splitted into version/score/textual severity columns

3) EPSS Table: Now shows all unique entries (not just first occurrence)

@mjherzog  I have made the changes according to the issue and If you have any suggestions please let me know I will make the changes required.

Screenshot 2025-03-17 at 6 09 57 PM

### DIFF
--- a/vulnerabilities/templates/vulnerability_details.html
+++ b/vulnerabilities/templates/vulnerability_details.html
@@ -197,64 +197,103 @@ VulnerableCode Vulnerability Details - {{ vulnerability.vulnerability_id }}
 
             <!-- New Severities Tab -->
             <div class="tab-div content" data-content="severities">
-                <table class="table is-bordered is-striped is-narrow is-hoverable is-fullwidth gray-header-border">
-                    <tr>
-                        <th style="width: 160px;"> System </th>
-                        <th style="width: 100px;"> Score </th>
-                        <th> Found at </th>
-                    </tr>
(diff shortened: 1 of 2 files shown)


pombredanne:
This is not actionable. Please take the time to test your code and stop submitting code that does not run and is AI-generated, wasting precious review time from the maintainers.
If you are able to submit code that is minimally tested and does run and is not AI generated, come back, but only then.
I am going to close all your other PRs for the same reason

PR #1833, opened 2025-03-28
Add pipeline for adding advisory ID and tests

TG1999:


vulnerabilities/pipelines/add_advisory_id.py
```diff
@@ -0,0 +1,55 @@
...
+
+        batch_size = 500
+
+        progress = LoopProgress(total_iterations=advisories.count(), logger=self.log)
+
+        for advisory in progress.iter(advisories.iterator(chunk_size=batch_size)):
+            importer_name = advisory.created_by
+            aliases = Alias.objects.filter(advisories=advisory).values_list("alias", flat=True)
+            advisory_id = IMPORTERS_REGISTRY[importer_name].get_advisory_id(aliases=aliases)
+            if advisory_id is None:
```
TG1999, line 43: Use `if not advisory_id` instead of 
```suggestion
            if not advisory_id:
```

TG1999: log this thing

vulnerabilities/pipelines/add_advisory_id.py
```diff
@@ -0,0 +1,55 @@
...
+                continue
+            advisory.advisory_id = advisory_id
```
TG1999, line 45: Disassociate the advisory_id as an alias from the advisory

Issue #1798, opened 2025-03-07
Get rid of the custom `paginated()` and use the built-in `iterator()`


keshav-space:
The custom `BaseQuerySet.paginated()` performs extremely poorly in terms of compute time and memory usage compared to the built-in `iterator()`. 

### Using custom `paginated()`

Even after 20 minutes, the iteration is not completed, and the ETA keeps getting worse and worse.

```python
Iterating advisory data using custom BaseQuerySet.paginated()
INFO 2025-03-07 08:56:41.500 Pipeline [IterateAdvisory] starting
INFO 2025-03-07 08:56:41.500 Step [iterate_advisories] starting
INFO 2025-03-07 08:56:41.796 Iterating over 8073782 Advisory objects.
INFO 2025-03-07 08:57:50.106 Progress: 10% (807379/8073782) ETA: 615 seconds (10.2 minutes)
INFO 2025-03-07 09:00:09.226 Progress: 20% (1614757/8073782) ETA: 830 seconds (13.8 minutes)
INFO 2025-03-07 09:03:34.920 Progress: 30% (2422135/8073782) ETA: 964 seconds (16.1 minutes)
INFO 2025-03-07 09:09:04.444 Progress: 40% (3229513/8073782) ETA: 1114 seconds (18.6 minutes)
INFO 2025-03-07 09:16:32.571 Progress: 50% (4036891/8073782) ETA: 1191 seconds (19.9 minutes)
^CCommandError: Keyboard interrupt received. Stopping...
```

### Using built-in `iterator()`

It takes around 4 minutes to iterate over 8 million advisories.
```python
Iterating advisory data using built-in .iterator()
INFO 2025-03-07 10:02:43.483 Pipeline [IterateAdvisory] starting
INFO 2025-03-07 10:02:43.483 Step [iterate_advisories] starting
INFO 2025-03-07 10:02:44.042 Iterating over 8073782 Advisory objects.
INFO 2025-03-07 10:05:00.181 Progress: 10% (807379/8073782) ETA: 1225 seconds (20.4 minutes)
INFO 2025-03-07 10:05:19.790 Progress: 20% (1614757/8073782) ETA: 623 seconds (10.4 minutes)
INFO 2025-03-07 10:05:37.013 Progress: 30% (2422135/8073782) ETA: 404 seconds (6.7 minutes)
[...]

PR #1845, opened 2025-04-02
Replace paginated with iterator

kunalsz:
In reference to the issue #1798 

- Replaced the custom paginated by using the inbuilt `.iterator()` method
- Also removed the `paginated` method from the `BaseQuerySet`

@keshav-space Looking forward to your insights

### DIFF
--- a/vulnerabilities/improvers/default.py
+++ b/vulnerabilities/improvers/default.py
@@ -44,9 +44,9 @@ class DefaultImprover(Improver):
             return (
                 Advisory.objects.filter(Q(created_by=self.importer.qualified_name))
                 .order_by("-date_collected")
-                .paginated()
+                .iterator()
             )
-        return Advisory.objects.all().order_by("-date_collected").paginated()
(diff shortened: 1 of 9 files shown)


pombredanne:
@kunalsz Thanks! Did you make any quick measurement before and after to validate that this is more efficient in ram/cpu/number of queries? Would there be a small unit test that could check this?

kunalsz:
Hi @pombredanne , I don't know how to test this out like @keshav-space did in the respective issue, if he could share the code/test file which he used to compare both the functions then I'll happily add the test results too.

keshav-space:
> Hi @pombredanne , I don't know how to test this out like @keshav-space did in the respective issue, if he could share the code/test file which he used to compare both the functions then I'll happily add the test results too.

@kunalsz a very simple way to evaluate the performance would be to start VulnerableCode in Docker and run a large pipeline, like the GitHub importer pipeline, and collect the logs. Do this process twice, once with the old paginated and then with the built-in iterator and compare the logs. Make sure to reset the database between the two runs.

Issue #1903, opened 2025-06-09
Modify the GitLab importer to support package-first mode


michaelehab:
As a part of the ongoing GSoC 2025 project titled "On-demand live evaluation of packages and Integration with VulnTotal", this issue should set the base for using the new API endpoint.
It should modify the gitlab importer to support the package-first mode.
After the implementation of this issue, when a user calls the new API endpoint with Gitlab, the advisories database should only have those related to the input PackageURL.

PR #1910, opened 2025-06-15
Modify Gitlab Importer to support package-first mode

michaelehab:
Solves #1903 
- Extract Gitlab API handling code from vulntotal gitlab datasource into a utils file.
- Update vulntotal gitlab datasource to use the utils file.
- Update vulntotal gitlab datasource tests to reflect the changes.
- Modify Gitlab Importer to support package-first mode when a PURL is provided in the constructor.
- Update Gitlab Importer tests to test package-first mode.

### DIFF
--- a/vulnerabilities/pipelines/gitlab_importer.py
+++ b/vulnerabilities/pipelines/gitlab_importer.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from typing import Iterable
 from typing import List
 from typing import Tuple
+from urllib.parse import urljoin
 
 import pytz
 import saneyaml
@@ -31,6 +32,9 @@ from vulnerabilities.pipelines import VulnerableCodeBaseImporterPipeline
 from vulnerabilities.utils import build_description
 from vulnerabilities.utils import get_advisory_url
 from vulnerabilities.utils import get_cwe_id
+from vulntotal.datasources.gitlab import get_casesensitive_slug
+from vulntotal.datasources.gitlab_api import fetch_gitlab_advisories_for_purl
+from vulntotal.datasources.gitlab_api import get_estimated_advisories_count
 
 
 class GitLabImporterPipeline(VulnerableCodeBaseImporterPipeline):
@@ -42,9 +46,16 @@ class GitLabImporterPipeline(VulnerableCodeBaseImporterPipeline):
     license_url = "https://gitlab.com/gitlab-org/advisories-community/-/blob/main/LICENSE"
     importer_name = "GitLab Importer"
     repo_url = "git+https://gitlab.com/gitlab-org/advisories-community/"
+    is_batch_run = True
 
     @classmethod
     def steps(cls):
+        if not cls.is_batch_run:
+            return (
+                cls.collect_and_store_advisories,
+                cls.import_new_advisories,
+            )
+
         return (
             cls.clone,
             cls.collect_and_store_advisories,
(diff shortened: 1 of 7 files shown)


TG1999:
Make changes for this as well thanks! https://github.com/aboutcode-org/vulnerablecode/blob/main/vulnerabilities/pipelines/v2_importers/gitlab_importer.py

michaelehab:
@TG1999 I modified the V2 importer as well

vulnerabilities/pipelines/gitlab_importer.py
```diff
@@ -31,6 +32,9 @@
 from vulnerabilities.utils import build_description
 from vulnerabilities.utils import get_advisory_url
 from vulnerabilities.utils import get_cwe_id
+from vulntotal.datasources.gitlab import get_casesensitive_slug
```
TG1999: @keshav-space what do you think is it a good idea to import vulntotal functions in vcio ? Or shall we create separate functions here ?

TG1999: @michaelehab , also spdx license is missing in Gitlab vulntotal datasource. We need to have that before using it in our VCIO importers. https://github.com/aboutcode-org/vulnerablecode/blob/main/vulntotal/datasources/gitlab.py#L30

keshav-space: > what do you think is it a good idea to import vulntotal functions in vcio ?

We can import it from VulnTotal for now. Later on we can extract these out in common utility.

michaelehab: > @michaelehab , also spdx license is missing in Gitlab vulntotal datasource. We need to have that before using it in our VCIO importers. https://github.com/aboutcode-org/vulnerablecode/blob/main/vulntotal/datasources/gitlab.py#L30

I believe we discussed this in an earlier meeting where we said that the package-first API endpoint won't be enabled by default and users have to enable it locally which is like using vulntotal when it comes to license, that's even why I reused the vulntotal functions in VCIO's package-first mode.

vulnerabilities/pipelines/v2_importers/gitlab_live_importer.py
keshav-space, line 1: IMHO this should be inside `vulnerabilities/pipelines/v2_importers/gitlab_importer.py`

keshav-space (changes requested):
Thanks @michaelehab, pipeline steps are looking good. few nits for your consideration.

vulnerabilities/pipelines/v2_importers/gitlab_live_importer.py
```diff
@@ -0,0 +1,242 @@
...
+    advisory: dict,
+    purl_type_by_gitlab_scheme,
+    gitlab_scheme_by_purl_type,
+    logger,
+    purl=None,
+    advisory_url=None,
+):
+    """
+    Convert a GitLab advisory dict to AdvisoryDataV2.
+    """
```
keshav-space: This has lots of duplicate with https://github.com/aboutcode-org/vulnerablecode/blob/dcb0511c73283654ab8a4ca340b71d6d9c5a16b9/vulnerabilities/pipelines/v2_importers/gitlab_importer.py#L176-L304 we should re use the existing code and if needed we can break down `parse_gitlab_advisory` smaller function.

vulnerabilities/pipelines/v2_importers/gitlab_live_importer.py
michaelehab: image
image

Based on the proposed architecture, we agreed that we'll have another parallel registry for live importers, each new live importer pipeline class will be added to this registry which the API endpoint uses to pick the relevant live importers to run.

vulnerabilities/pipelines/v2_importers/gitlab_live_importer.py
michaelehab: I've adressed this in my latest commit by extracting the shared logic into a common file so we can avoid duplicate code, thanks for pointing out.

vulnerabilities/pipelines/v2_importers/gitlab_live_importer.py
michaelehab: I guess it will be better if we merge #1969 first, it adds the live evaluation API which directly utilizes the LIVE_IMPORTERS_REGISTRY.

PR #1913, opened 2025-06-15
Modify Rust importer to support package-first mode

michaelehab:
Solves #1911 
* Update Rust importer to only load and process advisories relevant to the purl passed in the constructor
* Update Rust importer tests to include testing the package-first mode

### DIFF
--- a/vulnerabilities/importers/rust.py
+++ b/vulnerabilities/importers/rust.py
@@ -8,7 +8,9 @@
 #
 
 import asyncio
+import logging
 from itertools import chain
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Set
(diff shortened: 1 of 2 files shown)


vulnerabilities/importers/rust.py
```diff
@@ -27,8 +29,19 @@
...
 from vulnerabilities.utils import nearest_patched_package
 
+logger = logging.getLogger(__name__)
+
 
 class RustImporter(Importer):
+    def __init__(self, purl=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.purl = purl
+        if self.purl:
```
TG1999, line 39:
```suggestion
        if self.purl and self.purl.type!="cargo"
```

TG1999:
We are not running Rust importers either in V1 nor in V2 as of now, I think we need to first make importer functional before doing this.

pombredanne:
@michaelehab can you join the community calls and chat rooms so we can discuss this?

TG1999:
@michaelehab it will be helpful for us mentors and community members to design and help you if you can regularly join us in VCIO community calls that happens on Tuesday 8:00 AM PST https://meet.jit.si/VulnerableCode

michaelehab:
@TG1999 @pombredanne Sure, I will start joining this call too regularly so we can discuss this more. 
Regarding this PR, I worked on it to show how the Rust importer can be potentially supported like other importers. I wanted to mark it as a draft PR, but I'm not sure I have the access to do this.

PR #1923, opened 2025-06-29
Modify Curl importer to support package-first mode

michaelehab:
Solves #1918

* Update Curl importer to filter and process advisories relevant to the PURL passed in the constructor
* Update Curl importer tests to include testing the package-first mode

### DIFF
--- a/vulnerabilities/importers/curl.py
+++ b/vulnerabilities/importers/curl.py
@@ -39,6 +39,15 @@ class CurlImporter(Importer):
     importer_name = "Curl Importer"
     api_url = "https://curl.se/docs/vuln.json"
 
+    def __init__(self, purl=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.purl = purl
+        if self.purl:
+            if self.purl.type != "generic" or self.purl.name != "curl":
+                print(
+                    f"Warning: This importer handles curl package vulnerabilities. Current PURL: {self.purl!s}"
+                )
+
     def fetch(self) -> Iterable[Mapping]:
         response = fetch_response(self.api_url)
         return response.json()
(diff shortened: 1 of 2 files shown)


vulnerabilities/importers/curl.py
```diff
@@ -48,11 +57,42 @@ def advisory_data(self) -> Iterable[AdvisoryData]:
...
+
+            if self.purl.version and affected_package.affected_version_range:
+                try:
+                    purl_version = SemverVersion(self.purl.version)
+
+                    if purl_version not in affected_package.affected_version_range:
+                        continue
+                except Exception as e:
+                    logger.error(f"Error checking version {self.purl.version}: {e}")
+                    continue
```
ziadhany: Better catch specific exceptions `InvalidVersion`

vulnerabilities/importers/curl.py
```diff
@@ -39,6 +39,15 @@ class CurlImporter(Importer):
...
     api_url = "https://curl.se/docs/vuln.json"
 
+    def __init__(self, purl=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.purl = purl
+        if self.purl:
+            if self.purl.type != "generic" or self.purl.name != "curl":
+                print(
+                    f"Warning: This importer handles curl package vulnerabilities. Current PURL: {self.purl!s}"
+                )
```
ziadhany:
```suggestion
        super().__init__(*args, **kwargs)
        self.purl = purl
        if self.purl and (self.purl.type != "generic" or self.purl.name != "curl"):
            print(
                f"Warning: This importer handles curl package vulnerabilities. Current PURL: {self.purl!s}"
            )

```

ziadhany (changes requested):

PR #1935, opened 2025-07-04
Add Elixir Security Live V2 Importer

michaelehab:
Solves #1933

* Update Elixir Security importer to filter and process advisories relevant to the purl passed in the constructor
* Update Elixir Security v2 importer to filter and process advisories relevant to the purl passed in the constructor
* Update Elixir Security importer tests to include testing package-first mode

### DIFF
--- a/vulnerabilities/pipelines/v2_importers/elixir_security_importer.py
+++ b/vulnerabilities/pipelines/v2_importers/elixir_security_importer.py
@@ -58,32 +58,26 @@ class ElixirSecurityImporterPipeline(VulnerableCodeBaseImporterPipelineV2):
         return count
 
     def collect_advisories(self) -> Iterable[AdvisoryDataV2]:
-        try:
-            base_path = Path(self.vcs_response.dest_dir)
-            vuln = base_path / "packages"
-            for file in vuln.glob("**/*.yml"):
-                yield from self.process_file(file, base_path)
-        finally:
-            if self.vcs_response:
-                self.vcs_response.delete()
+        base_path = Path(self.vcs_response.dest_dir)
+        vuln = base_path / "packages"
+        for file in vuln.glob("**/*.yml"):
+            relative_path = str(file.relative_to(base_path)).strip("/")
+            path_segments = str(file).split("/")
+            # use the last two segments as the advisory ID
+            advisory_id = "/".join(path_segments[-2:]).replace(".yml", "")
+            advisory_url = f"https://github.com/dependabot/elixir-security-advisories/blob/master/{relative_path}"
+
+            yaml_file = load_yaml(str(file))
+            yield from self.build_advisory_from_text(
+                advisory_id=advisory_id, advisory_url=advisory_url, yaml_file=yaml_file
+            )
 
     def on_failure(self):
         self.clean_downloads()
 
-    def process_file(self, file, base_path) -> Iterable[AdvisoryDataV2]:
-        relative_path = str(file.relative_to(base_path)).strip("/")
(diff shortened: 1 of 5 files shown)


michaelehab:
@TG1999 I merged the recent fixes here and did another functional test for the v2 importer.
I'd appreciate your feedback and review, thanks.

vulnerabilities/importers/elixir_security.py
```diff
@@ -30,7 +34,22 @@ class ElixirSecurityImporter(Importer):
     spdx_license_expression = "CC0-1.0"
     importer_name = "Elixir Security Importer"
 
+    def __init__(self, purl=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.purl = purl
+        if self.purl:
```
TG1999: This pattern is being used at multiple importers, we shall extract it out as a function

michaelehab: I agree
We can modify the constructor of the base class instead of modifying each individual importer's constructor. 
But I believe in this case we won't show the warning messages if the purl is not right for the importer.

TG1999: We can always pass the message or the type or even the log as a parameter

michaelehab: I like the idea of passing the supported types, that would make it generic with less duplicate code
The warning message is the same, so it will adapt if we pass supported types

vulnerabilities/pipelines/v2_importers/elixir_security_importer.py
```diff
@@ -37,25 +41,65 @@ class ElixirSecurityImporterPipeline(VulnerableCodeBaseImporterPipelineV2):
...
+    def __init__(self, *args, purl=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.purl = purl
+        if self.purl:
+            ElixirSecurityImporterPipeline.is_batch_run = False
+            if self.purl.type != "hex":
+                self.log(
+                    f"Warning: PURL type {self.purl.type} is not 'hex', may not match any advisories"
+                )
+
```
keshav-space: Do not add `purl` as a primary argument to Pipeline. Primary arguments to the pipeline are exclusively reserved for managing pipeline execution. Instead, you can pass `purl` to the pipeline like this: `ElixirSecurityImporterPipeline(purl="pkg:hex/coherence")` and then access it inside a pipeline step using `self.inputs["purl"]`.

keshav-space (changes requested):
@michaelehab there is no need to override `__init__` in pipeline.

vulnerabilities/pipelines/v2_importers/elixir_security_importer.py
michaelehab: I see that the constructor has arguments for managing pipeline execution (in BasePipelineRun class).
How do I access the purl argument without overriding the constructor or modifying BasePipelineRun constructor?
The way I approached this was to call the parent constructor with the arguments so that the pipeline functions properly, and then take the PURL which is important in the importer-level.

michaelehab: And based on @TG1999's comment, we can unify the purl handling in package-first mode by overriding the VulnerableCodeBaseImporterPipeline constructor for example to do something like this.
```
def __init__(self, *args, **kwargs):
    self.purl = kwargs.get("purl")
    self.supported_types = kwargs.get("supported_types)
```
Then doing the package-first checks and warnings which is unified for all importers.

PR #1941, opened 2025-07-12
Add NPM Live Importer V2

michaelehab:
Solves #1936 
- Update NPM importer to filter and process advisories relevant to the purl passed in the constructor
- Update NPM v2 importer to filter and process advisories relevant to the purl passed in the constructor
- Update NPM importer tests to test package-first mode
- Update NPM v2 importer tests to test package-first mode

### DIFF
--- a/vulnerabilities/importers/__init__.py
+++ b/vulnerabilities/importers/__init__.py
@@ -62,6 +62,7 @@ from vulnerabilities.pipelines.v2_importers import mattermost_importer as matter
 from vulnerabilities.pipelines.v2_importers import mozilla_importer as mozilla_importer_v2
 from vulnerabilities.pipelines.v2_importers import nginx_importer as nginx_importer_v2
 from vulnerabilities.pipelines.v2_importers import npm_importer as npm_importer_v2
+from vulnerabilities.pipelines.v2_importers import npm_live_importer as npm_live_importer_v2
 from vulnerabilities.pipelines.v2_importers import nvd_importer as nvd_importer_v2
 from vulnerabilities.pipelines.v2_importers import openssl_importer as openssl_importer_v2
 from vulnerabilities.pipelines.v2_importers import oss_fuzz as oss_fuzz_v2
@@ -196,3 +197,9 @@ TODO_EXCLUDED_PIPELINES = [
     for key, value in IMPORTERS_REGISTRY.items()
     if issubclass(value, VulnerableCodeBaseImporterPipelineV2) and value.exclude_from_package_todo
 ]
+
+LIVE_IMPORTERS_REGISTRY = create_registry(
+    [
+        npm_live_importer_v2.NpmLiveImporterPipeline,
+    ]
+)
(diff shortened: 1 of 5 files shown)


vulnerabilities/pipelines/npm_importer.py
```diff
@@ -39,14 +44,24 @@ class NpmImporterPipeline(VulnerableCodeBaseImporterPipeline):
...
+        self.purl = purl
+        if self.purl:
+            NpmImporterPipeline.is_batch_run = False
+            if self.purl.type != "npm":
+                print(f"Warning: This importer handles NPM packages. Current PURL: {self.purl!s}")
+
     @classmethod
     def steps(cls):
-        return (
+        return [
```
TG1999: Why this change ?

vulnerabilities/pipelines/npm_importer.py
```diff
@@ -58,9 +73,26 @@ def advisories_count(self):
...
+                    if data.get("module_name") == package_name:
+                        affected_package = self.get_affected_package(data, package_name)
+                        if not self.purl.version or self._version_is_affected(affected_package):
+                            filtered_files.append(advisory_file)
+                except Exception as e:
+                    self.log(f"Error processing advisory file {advisory_file}: {str(e)}")
+            advisory_files = filtered_files
+
+        for advisory in list(advisory_files):
+            for result in self.to_advisory_data(advisory):
```
TG1999: Why we are not using yield from here ?

vulnerabilities/pipelines/npm_importer.py
```diff
@@ -58,9 +73,26 @@ def advisories_count(self):
 
     def collect_advisories(self) -> Iterable[AdvisoryData]:
         vuln_directory = Path(self.vcs_response.dest_dir) / "vuln" / "npm"
-
-        for advisory in vuln_directory.glob("*.json"):
-            yield from self.to_advisory_data(advisory)
+        advisory_files = list(vuln_directory.glob("*.json"))
+
+        if not self.is_batch_run:
+            package_name = self.purl.name
```
TG1999: Why we not actually check in our DB if the purl exists, before even checking all files ?

vulnerabilities/pipelines/npm_importer.py
michaelehab: It should be a tuple as well to maintain code consistency. I will change this.

vulnerabilities/pipelines/npm_importer.py
michaelehab: How does that guarantee that we won't find new advisories if we have the purl in DB? 
As far as I understand, in order to find out whether we have new advisories, we have to check all files.

vulnerabilities/pipelines/npm_importer.py
michaelehab: Do you mean?
```
for advisory in list(advisory_files):
   yield from self.to_advisory_data(advisory)
```

PR #1970, opened 2025-07-29
Add ImpactedPackage model to track affected and fixed packages

keshav-space:
Resolves:

- [x] https://github.com/aboutcode-org/vulnerablecode/issues/1950
- [x] https://github.com/aboutcode-org/vulnerablecode/issues/1960
- [x] https://github.com/aboutcode-org/vulnerablecode/issues/1959

### DIFF
--- a/vulnerabilities/pipes/advisory.py
+++ b/vulnerabilities/pipes/advisory.py
@@ -134,6 +134,7 @@ def insert_advisory(advisory: AdvisoryData, pipeline_id: str, logger: Callable =
     return advisory_obj
 
 
+@transaction.atomic
 def insert_advisory_v2(
     advisory: AdvisoryData,
     pipeline_id: str,
--- a/vulnerabilities/tests/pipes/test_vulnerablecode_importer_pipeline_v2.py
+++ b/vulnerabilities/tests/pipes/test_vulnerablecode_importer_pipeline_v2.py
@@ -10,11 +10,17 @@
 import logging
 from datetime import datetime
 from datetime import timedelta
+from unittest.mock import patch
 
 import pytest
+from packageurl import PackageURL
+from univers.version_range import VersionRange
 
 from vulnerabilities.importer import AdvisoryData
+from vulnerabilities.importer import AffectedPackageV2
 from vulnerabilities.models import AdvisoryV2
+from vulnerabilities.models import ImpactedPackage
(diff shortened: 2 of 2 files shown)


vulnerabilities/templates/advisory_package_details.html
```diff
@@ -6,83 +6,62 @@
 {% load url_filters %}
 
 {% block title %}
-VulnerableCode Advisory Package Details - {{ advisory.advisory_id }}
```
TG1999, line 9: Why this change ?

vulnerabilities/pipes/advisory.py
```diff
@@ -192,6 +188,29 @@ def insert_advisory_v2(
                 level=logging.ERROR,
             )
 
+    if created:
```
TG1999, line 192: How do we ensure that if the loop breaks before creating all ImpactedPackages in the list, then also we are able to create all ImpactedPackages when the importer restarts.

vulnerabilities/pipelines/__init__.py
```diff
@@ -271,13 +260,15 @@ class VulnerableCodeBaseImporterPipelineV2(VulnerableCodePipeline):
...
     repo_url = None
-    advisory_confidence = MAX_CONFIDENCE
     ignorable_versions = []
-    unfurl_version_ranges = False
 
     @classmethod
     def steps(cls):
-        return (cls.collect_and_store_advisories,)
+        return (
+            # Add step for downloading/cloning resource as required.
```
TG1999, line 268: If these comments are TODOs, please mark them as todos

vulnerabilities/templates/advisory_package_details.html
keshav-space: Primary context object for this view is advisoryv2 object.

vulnerabilities/pipes/advisory.py
keshav-space: We now use atomic transactions to ensure that partial advisory is never imported.

vulnerabilities/pipelines/__init__.py
keshav-space: No, these are not TODOs. These are just instructions for whoever implements pipeline using this base, should consider adding these steps if applicable to their use case.

TG1999 (approved):
LGTM!

PR #1971, opened 2025-08-01
Add v2 pipeline for importing Red Hat advisories

### LINKED ISSUE #1954: Redhat

keshav-space:
- Resolves #1954

### DIFF
--- a/vulnerabilities/importer.py
+++ b/vulnerabilities/importer.py
@@ -458,6 +458,22 @@ class AdvisoryData:
         return summary
 
     def to_dict(self):
+        is_adv_v2 = (
+            self.severities
+            or self.references_v2
+            or (self.affected_packages and isinstance(self.affected_packages[0], AffectedPackageV2))
+        )
+        if is_adv_v2:
+            return {
+                "aliases": self.aliases,
+                "summary": self.summary,
+                "affected_packages": [pkg.to_dict() for pkg in self.affected_packages],
+                "references_v2": [ref.to_dict() for ref in self.references_v2],
+                "severities": [sev.to_dict() for sev in self.severities],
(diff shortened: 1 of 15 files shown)


vulnerabilities/importer.py
```diff
@@ -458,6 +460,22 @@ def clean_summary(self, summary):
         return summary
 
     def to_dict(self):
+        is_adv_v2 = (
+            self.severities
```
TG1999: What if references_v2, severities, affected_packages are empty

vulnerabilities/importer.py
```diff
@@ -458,6 +460,22 @@ def clean_summary(self, summary):
...
+            or self.references_v2
+            or (self.affected_packages and isinstance(self.affected_packages[0], AffectedPackageV2))
```
TG1999, line 467: Also do we need advisory_id in json as well ?

vulnerabilities/importer.py
keshav-space: In such case there is no way to differentiate between v1 and v2 and hence it shall be treated as v1 ATM to prevent breaking v1 importers.
Edit: If `advisory_id` is required field for v2 advisory we can include that too here to determine if its an v2 advisory, let me update this.

vulnerabilities/importer.py
keshav-space: Yes we can, but i suggest we do it in separate pr since that may require adjusting tests for our existing v2 pipelines.

keshav-space: We also have `advisory_id` now https://github.com/aboutcode-org/vulnerablecode/pull/1971/commits/737d94a75f6c9825a2693ba3a07e961ab6e9380e

Issue #1383, opened 2023-12-28
Review and add CVE fixes  dataset


pombredanne:
See https://zenodo.org/records/7029359 and https://github.com/secureIT-project/CVEfix

armijnhemel:
The latter URL should be: https://github.com/secureIT-project/CVEfixes

ziadhany:
This is done:
- https://github.com/aboutcode-org/vulnerablecode/pull/1992

For the CVEfixes dataset, we collect fix commits in a similar way to what is implemented here:

https://github.com/secureIT-project/CVEfixes/blob/main/Code/collect_commits.py#L76

Issue #1129, opened 2023-02-14
Extract unpublished vulnerabilities from commit histories and trackers


pombredanne:
We should extract unpublished vulnerabilities from commit histories and issue trackers
- [ ] Parse issues and trackers such as github issues
- [ ] Parse CHANGELOGs See https://github.com/nexB/vulnerablecode/issues/233 and https://github.com/pyupio/changelogs/
- [ ] Parse Git commit messages (and to include fix commits #207 )

These are valuable information and we can search for CVE and security-related keywords and track these in a curation queue. And eventually submit these as NVD CVEs.

pombredanne:
We now can parse Git commit messages for vulnerabilities.
- The main PR is https://github.com/aboutcode-org/vulnerablecode/pull/1992

This is an initial script to parse Git commit messages that can be easily integrated with our model. The script takes a Git repository as input, parses all commits, and returns the CVEs along with their corresponding fixed commits.

We now have a "collect_fix_commits" importer, and we import fix commits from at least these repositories: 
- https://github.com/aboutcode-org/vulnerablecode/pull/1992#issuecomment-3828666714

PR #1992, opened 2025-08-30
Add support for parsing Git commit messages

ziadhany:
I created an initial script to parse Git commit messages that can be easily integrated with our model. The script takes a Git repository as input, parses all commits, and returns the CVEs along with their corresponding fixed commits.

Issues:
- #1383
- #2000
- #1129 

results:
[vulnerabilities_packagecommitpatch2.zip](https://github.com/user-attachments/files/25210286/vulnerabilities_packagecommitpatch2.zip)


vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+    def steps(cls):
+        return (
+            cls.clone,
+            cls.collect_and_store_advisories,
+            cls.clean_downloads,
+        )
+
+    def clone(self):
+        """Clone the repository."""
+        self.repo_url = "https://github.com/torvalds/linux"
```
ziadhany: This part should not be static

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+            "git",
+            "clone",
+            "--bare",
+            "--filter=blob:none",
+            "--no-checkout",
+            self.repo_url,
+            repo_path,
+        ]
+        subprocess.run(cmd, check=True)
+        self.repo = Repo(repo_path)
```
keshav-space: Since we already using gitpython let's use the same for cloning too.

```suggestion
        self.repo = Repo.clone_from(
                url=self.repo_url,
                to_path=repo_path,
                bare=True,
                no_checkout=True,
                multi_options=["--filter=blob:none"]
        )
```

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+        return grouped_commits
+
+    def collect_advisories(self):
+        """
+        Generate AdvisoryData objects for each vulnerability ID grouped with its related commits.
+        """
+        self.log("Generating AdvisoryData objects from grouped commits.")
+        grouped_commits = self.collect_fix_commits()
+        for vuln_id, commits in grouped_commits.items():
+            references = [ReferenceV2(url=f"{self.repo_url}/commit/{cid}") for cid, _ in commits]
```
keshav-space: Where are we storing proper fix commit? just keeping it in reference is not sufficient IMO.

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+
+            summary_lines = [f"- {cid}: {msg}" for cid, msg in commits]
+            summary = f"Commits fixing {vuln_id}:\n" + "\n".join(summary_lines)
+            yield AdvisoryData(
+                advisory_id=vuln_id,
```
keshav-space, line 139: This will be problematic since we intend to collect fixed commits from multiple different repo here. Suppose we get a fix commit for `CVE-000-000` in two different repo, we will end up with a conflict while inserting the advisory, as we use `advisory_id` prefixed with the `pipeline_id` to create unique AVID. In this case we will end up with same AVID for fix commits imported from two different git repos.

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+                aliases=[vuln_id],
+                summary=summary,
+                references_v2=references,
+                url=self.repo_url,
+            )
+
+    def clean_downloads(self):
+        """Cleanup any temporary repository data."""
+        self.log("Cleaning up local repository resources.")
+        if os.path.isdir(self.repo.working_tree_dir):
```
keshav-space: `working_tree_dir` is not what we need here this will return `None` as we are working on bare repository, see https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.working_tree_dir.
We should instead use `working_dir`.

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
```diff
@@ -0,0 +1,117 @@
...
+            shutil.rmtree(path=self.repo.working_tree_dir)
```
keshav-space: Also, let's make sure that we have local clone before proceeding to remove.

```suggestion
        if hasattr(self, 'repo') and self.repo.working_dir:
            shutil.rmtree(path=self.repo.working_dir)
```

keshav-space (changes requested):
Thanks @ziadhany, see some suggestions.

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
ziadhany: @keshav-space I was relying on this pipeline [ CollectFixCommitsPipeline ](https://github.com/aboutcode-org/vulnerablecode/blob/main/vulnerabilities/pipelines/collect_commits.py#L26) to create fix commits , the issue is that `CommitFixV2` is currently tied to both affected_package and advisory.  IMO, storing fix commits as references can be useful.

vulnerabilities/pipelines/v2_importers/collect_repo_fix_commits.py
ziadhany: @keshav-space Not sure what the best solution for this is, but based on my understanding, we can make the pipeline_id dynamic

* `django_fix_commit`
* `django_restframework_fix_commit`

For example:

```yaml
avid: "e.g., django_fix_commit/PYSEC-2020-2233"
avid: "e.g., django_restframework_fix_commit/PYSEC-2020-2233"
```

This will generate a different `avid` for each Git repository.
I’m not sure if this completely solves the problem, though.

ziadhany:
This is the list of repositories I think we should collect and parse Git messages from:

- [x] Linux Kernel: https://github.com/torvalds/linux
- [x] BusyBox: https://github.com/mirror/busybox.git
- [x] Nginx: https://github.com/nginx/nginx.git
- [x] Apache Tomcat: https://github.com/apache/tomcat.git
- [x] MySQL Server: https://github.com/mysql/mysql-server.git
- [x] PostgreSQL: https://github.com/postgres/postgres.git
- [x] MongoDB: https://github.com/mongodb/mongo.git
- [x] Redis: https://github.com/redis/redis.git
- [x] SQLite: https://github.com/sqlite/sqlite.git
- [x] PHP: https://github.com/php/php-src.git
- [x] Python (CPython): https://github.com/python/cpython.git
- [x] Ruby: https://github.com/ruby/ruby.git
- [x] Go (Golang): https://github.com/golang/go.git
- [x] Node.js: https://github.com/nodejs/node.git
- [x] Rust: https://github.com/rust-lang/rust.git
- [x] OpenJDK (Java): https://github.com/openjdk/jdk.git
- [x] Swift: https://github.com/swiftlang/swift.git
- [x] Django: https://github.com/django/django.git
- [x] Rails: https://github.com/rails/rails.git
- [x] Laravel: https://github.com/laravel/framework.git
- [x] Spring Framework: https://github.com/spring-projects/spring-framework.git
- [x] React: https://github.com/facebook/react
- [x] Angular: https://github.com/angular/angular
- [x] WordPress: https://github.com/WordPress/WordPress.git
- [x] Docker (Moby): https://github.com/moby/moby.git
- [x] Kubernetes: https://github.com/kubernetes/kubernetes.git
- [x] QEMU: https://gitlab.com/qemu-project/qemu.git
- [x] Xen Project: https://github.com/xen-project/xen.git
- [x] VirtualBox: https://github.com/mirror/vbox.git
- [x] Containerd: https://github.com/containerd/containerd.git
- [x] Ansible: https://github.com/ansible/ansible.git
- [x] Terraform: https://github.com/hashicorp/terraform.git
- [x] Wireshark: https://gitlab.com/wireshark/wireshark.git
- [x] Tcpdump: https://github.com/the-tcpdump-group/tcpdump.git
- [x] Git: https://github.com/git/git.git
- [x] Jenkins: https://github.com/jenkinsci/jenkins.git
- [x] GitLab (Community Edition): https://gitlab.com/gitlab-org/gitlab-foss.git

Ignored:
- OpenSSL: https://github.com/openssl/openssl.git

vulnerabilities/pipelines/__init__.py
```diff
@@ -321,3 +329,109 @@ def collect_and_store_advisories(self):
...
+                AffectedPackageV2(
+                    package=purl,
+                    fixed_by_commit_patches=[
+                        PackageCommitPatchData(vcs_url=self.repo_url, commit_hash=commit_hash)
+                        for commit_hash in commit_hash_set
+                    ],
+                )
+            ]
+
+            yield AdvisoryData(
```
keshav-space: Would it makes sense also keep all these commit messages in summary?

vulnerabilities/pipelines/__init__.py
```diff
@@ -321,3 +329,109 @@ def collect_and_store_advisories(self):
...
+
+    def clean_downloads(self):
+        """Cleanup any temporary repository data."""
+        self.log("Cleaning up local repository resources.")
+        if hasattr(self, "repo") and self.repo.working_dir:
+            shutil.rmtree(path=self.repo.working_dir)
+
+    def on_failure(self):
+        """Ensure cleanup is always performed on failure."""
+        self.clean_downloads()
```
keshav-space: This not part of base VCIO pipeline, lets move this inside `pipes/` directory.

vulnerabilities/pipelines/v2_importers/collect_fix_commits.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class CollectSwiftFixCommitsPipeline(CollectVCSFixCommitPipeline):
+    pipeline_id = "collect_swift_fix_commits"
+    repo_url = "https://github.com/swiftlang/swift.git"
+
+
+class CollectOpensslFixCommitsPipeline(CollectVCSFixCommitPipeline):
+    pipeline_id = "collect_openssl_fix_commits"
+    repo_url = "https://github.com/openssl/openssl.git"
```
keshav-space: Let's remove this, we are already collecting fixed commits in OpenSSL v2 pipeline.

keshav-space (changes requested):
@ziadhany thanks, see some feedback below.

vulnerabilities/tests/pipelines/v2_importers/test_collect_fix_commit.py
```diff
@@ -0,0 +1,120 @@
...
+        grouped_commits = json.loads(input_file.read_text(encoding="utf-8"))
+
+        pipeline = CollectVCSFixCommitPipeline()
+        pipeline.repo_url = "https://github.com/test/repo"
+        pipeline.log = MagicMock()
+        pipeline.collect_fix_commits = MagicMock(return_value=grouped_commits)
+
+        result = [adv.to_dict() for adv in pipeline.collect_advisories()]
+
+        util_tests.check_results_against_json(result, expected_file, True)
```
ziadhany:
```suggestion
        util_tests.check_results_against_json(result, expected_file)
```

vulnerabilities/pipes/vcs_collector_utils.py
```diff
@@ -0,0 +1,133 @@
...
+                AffectedPackageV2(
+                    package=purl,
+                    fixed_by_commit_patches=[
+                        PackageCommitPatchData(vcs_url=self.repo_url, commit_hash=commit_hash)
+                        for commit_hash in commit_hash_set
+                    ],
+                )
+            ]
+
+            yield AdvisoryData(
```
keshav-space: @ziadhany let's also add commit url for these commits in reference at least for known vcs like github.

keshav-space (approved):
Thanks @ziadhany, LGTM! Do we have an estimate of how long it takes to run `CollectLinuxFixCommitsPipeline`?

ziadhany:
> Thanks @ziadhany, LGTM! Do we have an estimate of how long it takes to run `CollectLinuxFixCommitsPipeline`?

The Linux kernel pipeline takes around 5 minutes if the network call is ignored.

pombredanne:
Refinement will be tracked in:
- https://github.com/aboutcode-org/vulnerablecode/issues/2319

PR #1995, opened 2025-09-09
Add pipeline to unfurl affected VERS range in V2 impacts

### LINKED ISSUE #1967: Create pipeline to unfurl version ranges
This is a must-have to cut a release.

keshav-space:
- Resolves #1967


vulnerabilities/models.py
```diff
@@ -2927,17 +2929,19 @@ class ImpactedPackage(models.Model):
...
     base_purl = models.CharField(
         max_length=500,
-        blank=True,
+        blank=False,
         help_text="Version less PURL related to impacted range.",
     )
 
     affecting_vers = models.TextField(
         blank=True,
+        null=True,
```
TG1999, line 2938: IMO, we should have a check that at least one of `affecting_vers` or `fixed_vers` should exist while creating/saving an object

vulnerabilities/models.py
```diff
@@ -3065,6 +3069,41 @@ def get_or_create_from_purl(self, purl: Union[PackageURL, str]):
 
         return package, is_created
 
+    def bulk_get_or_create_from_purls(self, purls: List[Union[PackageURL, str]]):
```
TG1999, line 3072: Shall we have a boolean to mark if an impactd package is unfurled, so we do not unfurl that again

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
```diff
@@ -0,0 +1,130 @@
...
+                )
+        except Exception as e:
+            logger(
+                f"Error while checking {version!s} in {affecting_version_range!s}: {e!r} \n {traceback_format_exc()}",
+                level=logging.ERROR,
+            )
+    return affected_purls
+
+
+def get_purl_versions(purl, cached_versions={}):
```
TG1999: This is dangerous because the dict persists across calls between executions.

What about:

```
def get_purl_versions(purl, cached_versions=None):
    if cached_versions is None:
        cached_versions = {}
```

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
```diff
@@ -0,0 +1,130 @@
...
+
+        processed_impacted_packages_count = 0
+        processed_affected_packages_count = 0
+        cached_versions = {}
+        self.log(f"Unfurl affected vers range for {impacted_packages_count:,d} ImpactedPackage.")
+        progress = LoopProgress(total_iterations=impacted_packages_count, logger=self.log)
+        for impact in progress.iter(impacted_packages):
+            purl = PackageURL.from_string(impact.base_purl)
+            if not impact.affecting_vers or not any(
+                c in impact.affecting_vers for c in ("<", ">", "!")
```
TG1999, line 46: Won't this ignore:
vers like this?
=1.0.0 or ^1.2.0

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
```diff
@@ -0,0 +1,130 @@
...
+        self.log(f"{processed_affected_packages_count:,d} new Impact-Package relation created.")
+
+
+def get_affected_purls(versions, affecting_vers, base_purl, logger):
+    affecting_version_range = VersionRange.from_string(affecting_vers)
+    version_class = affecting_version_range.version_class
+
+    try:
+        versions = [version_class(v) for v in versions]
+    except Exception as e:
```
TG1999, line 82: Let's return an empty list by default, since the calling side expects something to iterate upon

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
```diff
@@ -0,0 +1,130 @@
...
+
+def bulk_create_with_m2m(purls, impact, relation, logger):
+    """Bulk create PackageV2 and also bulk populate M2M Impact and Package relationships."""
+    if not purls:
+        return 0
+
+    affected_packages_v2 = PackageV2.objects.bulk_get_or_create_from_purls(purls=purls)
+
+    relations = [
+        relation(impactedpackage=impact, packagev2=package) for package in affected_packages_v2
```
TG1999, line 121: Would there be a chance that duped relations may be created from this ?

vulnerabilities/models.py
keshav-space: We’re being bit flexible in what we accept to make sure we at least store the base purl, even if we don't get affecting or fixed vers. And if there are no affecting or fixed vers, we will not attempt to unfurl no harm done.

vulnerabilities/models.py
keshav-space: We should not optimize for this right now, as sometimes during run we may not get complete list of versions due to timeout or network issue, and we don't want to skip processing them in our next pass.

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
keshav-space: That's what we want, we want to persist versions during execution. Whenever we start a new execution, we flush this by calling `get_purl_versions` with `cached_versions = {}`. 
https://github.com/aboutcode-org/vulnerablecode/blob/f55ad7b8fa448dafd82bbee56af2313ecf468812/vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py#L40

Nevertheless, I will change this to make it more straightforward so that `cached_versions` is no longer optional.
```suggestion
def get_purl_versions(purl, cached_versions):
```

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
keshav-space: No, this is a vers string, and VERS spec only allows `>`, `<`, and `!` for comparison.

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
keshav-space: If no purls are returned, we will skip processing the current impact right away and go no further.
https://github.com/aboutcode-org/vulnerablecode/blob/f55ad7b8fa448dafd82bbee56af2313ecf468812/vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py#L61-L62

vulnerabilities/pipelines/v2_improvers/unfurl_version_range.py
keshav-space: No duplicate relation can be created, we use `ignore_conflicts=True` while bulk creating to ignore already existing relations.

PR #1988, opened 2025-08-25
Collect existing fix commits for aosp_dataset

ziadhany:
- issue: #1697

PR #2017, opened 2025-11-03
Add support for introduced and fixed commits in AdvisoryData

ziadhany:
- Introduce `introduced_by_commits` and `fixed_by_commits` fields in our advisory
- Update `from_dict` and `to_dict` methods
- Create a `CodePatchData` importer class
- #2022
- #1988


TG1999:
@ziadhany add description in the PR please!

vulnerabilities/importer.py
```diff
@@ -444,6 +496,8 @@ class AdvisoryData:
     date_published: Optional[datetime.datetime] = None
     weaknesses: List[int] = dataclasses.field(default_factory=list)
     severities: List[VulnerabilitySeverity] = dataclasses.field(default_factory=list)
+    fixed_by_commits: List[CodeCommitData] = dataclasses.field(default_factory=list)
```
TG1999: Why are we making changes in V1 and V2 AdvisoryData, when we are using V1, let's get rid of V2 ?

vulnerabilities/importers/osv.py
```diff
@@ -392,11 +411,49 @@ def get_fixed_versions(fixed_range, raw_id, supported_ecosystem) -> List[Version
                 fixed_versions.append(SemverVersion(version))
             except InvalidVersion:
                 logger.error(f"Invalid SemverVersion: {version!r} for OSV id: {raw_id!r}")
+
+        if fixed_range_type == "GIT":
```
TG1999: Why this? And what's the get_code_commit function ?

vulnerabilities/importer.py
```diff
@@ -194,6 +194,58 @@ def from_url(cls, url):
...
+        if not self.vcs_url:
+            raise ValueError("Commit must have a non-empty vcs_url.")
+
+    def __lt__(self, other):
+        if not isinstance(other, CodeCommitData):
+            return NotImplemented
+        return self._cmp_key() < other._cmp_key()
+
+    # TODO: Add cache
+    def _cmp_key(self):
```
TG1999, line 226: commit_hash and VCS URL, shall be the keys only I guess, @keshav-space your thoughts on this ?

vulnerabilities/importers/osv.py
ziadhany: In the OSV schema, a commit be treated as the package version.
https://ossf.github.io/osv-schema/#python-vulnerability
```
"ranges": [ {
    "type": "GIT",
    "repo": "https://github.com/owner/repo",
    "events": [
      { "introduced": "X" },
      { "fixed": "Y" },
    ]
} ]
```
When importing this data into the new model, we currently log it as `Unsupported fixed version type`.
Instead, we should skip these entries silently to avoid unnecessary log noise, since they are already handled in the get_code_commit function.

vulnerabilities/importer.py
ziadhany: Are you sure we should remove V2?
We’re not currently using V2 in our code, but we’re planning to migrate to it, since we use `is_adv_v2` in our AdvisoryData to check whether an advisory supports V2.

I can change that, no problem , just wanted to make sure 😅 .

TG1999: Makes sense!

vulnerabilities/importers/osv.py
TG1999: Why to skip and not create code commits out of these ?

vulnerabilities/importer.py
TG1999: We should have all the fields in keys for comparison. Looks good!

vulnerabilities/importers/osv.py
ziadhany: Because the `get_fixed_versions` function returns only versions, I thought the name might be misleading.

No problem  I’ll update it and rename `get_fixed_versions` to `get_fixed_versions_and_commits`

TG1999:
@ziadhany mostly looks good! Please run the importer once and paste the logs here. Thanks!

I want to see if we are missing on any data in OSV format. And how does the AdvisoryData and ImpactedPackages looks with the new CommitData. Thanks!

vulnerabilities/importers/osv.py
```diff
@@ -350,53 +360,81 @@ def get_fixed_version_range(versions, ecosystem):
...
-            logger.error(f"Unsupported fixed version type: {version!r} for OSV id: {raw_id!r}")
+                logger.error(f"Invalid SemverVersion: {fixed!r} for OSV id: {raw_id!r}")
+
+        elif fixed_range_type == "GIT" and (fixed or introduced):
+            repo = ranges.get("repo")
+            if not repo:
+                logger.error(f"Missing 'repo' field in ranges: {ranges} (OSV id: {raw_id!r})")
+                continue
+
+            # Git uses this magic hash for the empty tree
```
TG1999: Add some reference, URLs for readers to refer for

ziadhany:
@TG1999 This is the log output for the following importers:

- pysec_importer_v2
- pypa_importer_v2
- oss_fuzz_importer_v2

[importers_logs.zip](https://github.com/user-attachments/files/23420135/importers_logs.zip)

the database query result :
`vulnerabilities_advisoryv2` Total rows:  10274
`vulnerabilities_impactedpackage_fixed_by_commits` Total rows:  4013 
`vulnerabilities_impactedpackage_affecting_commits` Total rows: 3623
`vulnerabilities_codecommit` Total rows: 3791

TG1999:
@ziadhany 

```
Invalid VersionRange  for affected_pkg: {'package': {'name': 'apache-commons-io', 'ecosystem': 'OSS-Fuzz', 'purl': 'pkg:generic/apache-commons-io'}, 'ranges': [{'type': 'GIT', 'repo': 'https://github.com/apache/commons-io.git', 'events': [{'introduced': '72b1f88fb722def136ce87c9b2bfdd3c9126bb3d'}, {'fixed': 'd3e5bd6de8bc96abbadccea8b934dc038a32e90c'}]}], 'versions': ['commons-io-2.14.0-RC1', 'rel/commons-io-2.14.0'], 'ecosystem_specific': {'severity': 'LOW'}, 'database_specific': {'introduced_range': 'c511d15294d1a406a177368804014313948e2601:06fde31494c279ad940149e1a3d4944040c73c0d', 'fixed_range': '247c8e7d85a8df293011c7e9c94fd50bb2986fb7:d3e5bd6de8bc96abbadccea8b934dc038a32e90c'}} for OSV id: 'OSV-2023-962': error:InvalidVersion("'commons-io-2.14.0-RC1' is not a valid <class 'univers.versions.SemverVersion'>")
Invalid VersionRange  for affected_pkg: {'package': {'name': 'apache-commons-io', 'ecosystem': 'OSS-Fuzz', 'purl': 'pkg:generic/apache-commons-io'}, 'ranges': [{'type': 'GIT', 'repo': 'https://github.com/apache/commons-io.git', 'events': [{'introduced': '72b1f88fb722def136ce87c9b2bfdd3c9126bb3d'}, {'fixed': 'd3e5bd6de8bc96abbadccea8b934dc038a32e90c'}]}], 'versions': ['commons-io-2.14.0-RC1', 'rel/commons-io-2.14.0'], 'ecosystem_specific': {'severity': 'LOW'}, 'database_specific': {'introduced_range': 'c511d15294d1a406a177368804014313948e2601:06fde31494c279ad940149e1a3d4944040c73c0d', 'fixed_range': '247c8e7d85a8df293011c7e9c94fd50bb2986fb7:d3e5bd6de8bc96abbadccea8b934dc038a32e90c'}} for OSV id: 'OSV-2023-618': error:InvalidVersion("'commons-io-2.14.0-RC1' is not a valid <class 'univers.versions.SemverVersion'>")
```

Why are we getting in this logs? The commit data should have been created for this

TG1999:
See all `Invalid VersionRange` errors. Why these are coming?

```
{'package': {'name': 'apache-commons-codec', 'ecosystem': 'OSS-Fuzz', 'purl': 'pkg:generic/apache-commons-codec'}, 'ranges': [{'type': 'GIT', 'repo': 'https://gitbox.apache.org/repos/asf/commons-codec.git', 'events': [{'introduced': '44e4c4d778c3ab87db09c00e9d1c3260fd42dad5'}, {'fixed': '3bf874e2141dc08550c0b330c7a7006f358bb0f0'}]}], 'versions': ['commons-codec-1.16.1-RC1', 'rel/commons-codec-1.16.1'], 'ecosystem_specific': {'severity': 'LOW'}, 'database_specific': {'fixed_range': '72c40fe6f62410bcaa019dbf2cb570ee4e49b70e:3bf874e2141dc08550c0b330c7a7006f358bb0f0'}} for OSV id: 'OSV-2023-1195': error:InvalidVersion("'commons-codec-1.16.1-RC1' is not a valid <class 'univers.versions.SemverVersion'>")
```

when we have introduced and fixed events to create code commit data.

ziadhany:
I updated the script to handle unsupported packages (especially for OSS-Fuzz). CodeCommit is no longer ignored even if the package is unsupported, and logs are now more meaningful.

This is the updated logs:
[importers_v2.zip](https://github.com/user-attachments/files/23479370/importers_v2.zip)


the database query result :
vulnerabilities_advisoryv2 Total rows: 17041
vulnerabilities_impactedpackage_fixed_by_commits Total rows: 7343
vulnerabilities_impactedpackage_affecting_commits Total rows: 6553
vulnerabilities_codecommit Total rows: 6553

 Issues related:
- pysec_importer_v2 / pypa_importer_v2:
    - https://github.com/aboutcode-org/univers/issues/174
    - https://github.com/aboutcode-org/vulnerablecode/issues/2019
 - oss_fuzz_importer_v2
   - ` Unsupported package type: None in OSV: 'OSV-2021-1227'` This means the package type is unknown (e.g., generic, etc.), and there is no PURL associated with it.
   - `Invalid VersionRange  for affected_pkg` It depends on whether this is a valid version, for example, a semver version or not. 
  example:
`>   SemverVersion('commons-io-2.14.0-RC1')`
`>   univers.versions.InvalidVersion: 'commons-io-2.14.0-RC1' is not a valid <class 'univers.versions.SemverVersion'>`

TG1999:
`ERROR 2025-11-11 13:34:49.213781 UTC Unsupported PyPI advisory data file: GHSA-227r-w5j2-6243.json`

This log does not tell me a lot, what's the data. Why this is unsupported.

TG1999:
`Invalid VersionRange for affected_pkg: ['0.8', '0.9', '0.9.3', '0.9.4', '0.9.5', '0.9.6', '0.9.7', '0.9.8', '0.9.9', '2.0.1', '2.0.1rc1', '2.0.1rc2-git', '2.0.1rc3', '2.0.1rc4', '2.0.2', '2.0.3', '2.0.4', '2.0.5', '2.0b4', '2.0b5', '2.0b6', '2.0b7', '2.0b8', '2.0b9', '3.0.0', '3.0.0b1', '3.0.0b2', '3.0.1', '3.0.2', '3.0.3', '3.0.4', '3.0.5', '3.1', '3.2', '3.2.1', '3.2.2', '3.2.3', '3.2.4', '3.2.5', '3.3', '3.4', '3.4.1', '3.4.2', '3.4.3', '3.4.4', '3.4.5', '3.5', '3.5b1', '3.6', '3.6.1', '3.6.2', '3.6.3', '3.6.4'] for OSV id: 'PYSEC-2021-859': error:InvalidVersion("'2.0.1rc2-git' is not a valid <class 'univers.versions.PypiVersion'>")`

One of the list might not be a valid version, but all others are valid, are we ingesting them or skipping whole list if we can't ingest one.

ziadhany:
> `ERROR 2025-11-11 13:34:49.213781 UTC Unsupported PyPI advisory data file: GHSA-227r-w5j2-6243.json`
> 
> This log does not tell me a lot, what's the data. Why this is unsupported.

@TG1999 We are ignoring GHSA files since we target only PYSEC files.
https://github.com/aboutcode-org/vulnerablecode/blob/main/vulnerabilities/pipelines/v2_importers/pysec_importer.py#L54

TG1999:
> > `ERROR 2025-11-11 13:34:49.213781 UTC Unsupported PyPI advisory data file: GHSA-227r-w5j2-6243.json`
> 
> > 
> 
> > This log does not tell me a lot, what's the data. Why this is unsupported.
> 
> 
> 
> @TG1999 We are ignoring GHSA files since we target only PYSEC files.
> 
> https://github.com/aboutcode-org/vulnerablecode/blob/main/vulnerabilities/pipelines/v2_importers/pysec_importer.py#L54

Then add that to the log as well :)

ziadhany:
> `Invalid VersionRange for affected_pkg: ['0.8', '0.9', '0.9.3', '0.9.4', '0.9.5', '0.9.6', '0.9.7', '0.9.8', '0.9.9', '2.0.1', '2.0.1rc1', '2.0.1rc2-git', '2.0.1rc3', '2.0.1rc4', '2.0.2', '2.0.3', '2.0.4', '2.0.5', '2.0b4', '2.0b5', '2.0b6', '2.0b7', '2.0b8', '2.0b9', '3.0.0', '3.0.0b1', '3.0.0b2', '3.0.1', '3.0.2', '3.0.3', '3.0.4', '3.0.5', '3.1', '3.2', '3.2.1', '3.2.2', '3.2.3', '3.2.4', '3.2.5', '3.3', '3.4', '3.4.1', '3.4.2', '3.4.3', '3.4.4', '3.4.5', '3.5', '3.5b1', '3.6', '3.6.1', '3.6.2', '3.6.3', '3.6.4'] for OSV id: 'PYSEC-2021-859': error:InvalidVersion("'2.0.1rc2-git' is not a valid <class 'univers.versions.PypiVersion'>")`
> 
> One of the list might not be a valid version, but all others are valid, are we ingesting them or skipping whole list if we can't ingest one.

We are skipping this since the version range would likely be inconsistent if we processed it.
I also created a related issue in univers: 
- https://github.com/aboutcode-org/univers/issues/174

I can changes this if needed.

TG1999:
> > `Invalid VersionRange for affected_pkg: ['0.8', '0.9', '0.9.3', '0.9.4', '0.9.5', '0.9.6', '0.9.7', '0.9.8', '0.9.9', '2.0.1', '2.0.1rc1', '2.0.1rc2-git', '2.0.1rc3', '2.0.1rc4', '2.0.2', '2.0.3', '2.0.4', '2.0.5', '2.0b4', '2.0b5', '2.0b6', '2.0b7', '2.0b8', '2.0b9', '3.0.0', '3.0.0b1', '3.0.0b2', '3.0.1', '3.0.2', '3.0.3', '3.0.4', '3.0.5', '3.1', '3.2', '3.2.1', '3.2.2', '3.2.3', '3.2.4', '3.2.5', '3.3', '3.4', '3.4.1', '3.4.2', '3.4.3', '3.4.4', '3.4.5', '3.5', '3.5b1', '3.6', '3.6.1', '3.6.2', '3.6.3', '3.6.4'] for OSV id: 'PYSEC-2021-859': error:InvalidVersion("'2.0.1rc2-git' is not a valid <class 'univers.versions.PypiVersion'>")`
> 
> > 
> 
> > One of the list might not be a valid version, but all others are valid, are we ingesting them or skipping whole list if we can't ingest one.
> 
> 
> 
> We are skipping this since the version range would likely be inconsistent if we processed it.
> 
> I also created a related issue in univers: 
> 
> - https://github.com/aboutcode-org/univers/issues/174
> 
> 
> 
> I can changes this if needed.

@keshav-space @pombredanne thoughts on this one ?

TG1999:
For PYSEC data we would be using github version range, coz the versions are Semver. And if a version is not parsable that version should be skipped. Not the entire range. Also we should introduce a flag for advisories that were not completely parsed. So in future if our parsing techniques gets better we can delete the incomplete parsed advisory with a new one.

vulnerabilities/importer.py
```diff
@@ -194,6 +194,64 @@ def from_url(cls, url):
...
+            "commit_message": self.commit_message,
+            "commit_date": self.commit_date,
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict):
+        """Create a Commit instance from a dictionary."""
+        commit_date = data.get("commit_date")
+        return cls(
+            commit_hash=str(data.get("commit_hash", "")),
```
keshav-space: `commit_hash` is always string, right?

vulnerabilities/importer.py
```diff
@@ -444,6 +502,8 @@ class AdvisoryData:
     date_published: Optional[datetime.datetime] = None
     weaknesses: List[int] = dataclasses.field(default_factory=list)
     severities: List[VulnerabilitySeverity] = dataclasses.field(default_factory=list)
+    fixed_by_commits: List[CodeCommitData] = dataclasses.field(default_factory=list)
+    affected_by_commits: List[CodeCommitData] = dataclasses.field(default_factory=list)
```
keshav-space: This is not correct. `affected_by_commits` and `fixed_by_commit` should be part of `AffectedPackage`, not directly part of `AdvisoryData`. In our django model we keep the affected and fixed commits in `ImpactedPackage` and `AffectedPackage` is a low level representation of that same `ImpactedPackage`.

That's the only way to properly capture which version/commit range is fixed by which version/commit. An advisory can have multiple such sets of affected and fixed versions/commits.

ziadhany: @keshav-space Yes, I think it should be part of AffectedPackage, but this would force us to create an ImpactedPackage relation before we can store fix commits. For example, if we have a pipeline that loops over 1,000 Git repositories to extract fixed commits, we might not be able to get ImpactedPackages (because there’s no purl provided for AffectedPackage).

Therefore, I think `fixed_by_commits` and `affected_by_commits` should be standalone if we don’t have ImpactedPackage/AffectedPackage.

Not sure but I can reimplement this if needed, I thought we have a previous discussion on this @TG1999

keshav-space: @ziadhany  

> For example, if we have a pipeline that loops over 1,000 Git repositories to extract fixed commits, we might not be able to get ImpactedPackages (because there’s no purl provided for AffectedPackage).

Don't you have vcs url for this commit? we can create purl from vcs url. Can you share some example.

ziadhany: Let's say we have a Swift package:

@keshav-space 
* **Swift:** [https://github.com/LiYanan2004/SFSymbolKit](https://github.com/LiYanan2004/SFSymbolKit)

The possible purls for this package could be:

* `pkg:swift/github.com/LiYanan2004/SFSymbolKit`
* `pkg:github/LiYanan2004/SFSymbolKit`

Now, consider another example - perhaps a GitLab or Android repository:
[https://android.googlesource.com/platform/external/freetype/+/f720f0dbcf012d6c984dbbefa0875ef9840458c6](https://android.googlesource.com/platform/external/freetype/+/f720f0dbcf012d6c984dbbefa0875ef9840458c6)
I think it can be a generic purl type.

We can have multiple purls, but if no purl is provided by the data source, we might need to allow some tolerance in how we store the CodeCommit with ImpactedPackage.

keshav-space: @ziadhany in both the example you shared i see no reason why we can not create `AffectedPackage`. we have all the details to create an `AffectedPackage` and associate commit with it.

And for https://github.com/LiYanan2004/SFSymbolKit at the very least we should create AffectedPackage with a github purl `pkg:github/LiYanan2004/SFSymbolKit`. And ideally we should create 2 AffectedPackage one for swift and one for github purl, and both of these can have the commit associated with it.

TG1999: Then how will we store vcs URL, and commit hashes ? Since current affectedpackage data class does not support any of this ?

vulnerabilities/importer.py
TG1999: @ziadhany there is no point of creating commits with empty strings. If commit hash is not present then let's not create a commit at all.

vulnerabilities/importer.py
keshav-space: That’s the change this PR should make i.e. to add the `affected_by_commits` and `fixed_by_commits` fields to `AffectedPackage` in line with our change to the `ImpactedPackage` model here #2007 .

ziadhany: There are two ways we import the data: we need to know which is the best way 
```
yield AdvisoryData(
    aliases=[vuln_id],
    affected_packages=[affected_packages],
    references=sorted(references),
    fixed_by_commits=[CommitV2(commit_hash="abcd", vcs_url="github/...")]
    affected_commits=[CommitV2(commit_hash="abcd", vcs_url="github/...")]
    date_published=date_published,
    url=self.data_url,
)
```

OR 

```
fixed_by_commits = [CommitV2(commit_hash="abcd", vcs_url="github/...")]
affected_commits = [CommitV2(commit_hash="abcd", vcs_url="github/...")]

affected_packages = [
	AffectedPackageV2(
		package=purl,
		fixed_by_commits=fixed_commits,
		affected_by_commits=affected_commits,
    )
]

yield AdvisoryData(
    aliases=[vuln_id],
    affected_packages=[affected_packages],
    references=sorted(references),
    date_published=date_published,
    url=self.data_url,
)
```

keshav-space: we're going with the second one.

vulnerabilities/importer.py
TG1999: This is still not fixed. Please check this! Thanks!

ziadhany:
This is the log output for the following importers:

- pysec_importer_v2
- pypa_importer_v2
- oss_fuzz_importer_v2
- github_osv_importer_v2


[importers.zip](https://github.com/user-attachments/files/23585420/importers.zip)

TG1999:
`Failed to extract fixed commits: ValueError('Commit must be a valid a commit_hash.')` We need to know the hash here. Do log the hash as well.

TG1999:
Unsupported severity type: {'type': 'CVSS_V4', 'score': 'CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:N/VI:H/VA:N/SC:N/SI:N/SA:N'} for OSV id: 'PYSEC-2024-154'

Why this score is not supported ?

ziadhany:
> Unsupported severity type: {'type': 'CVSS_V4', 'score': 'CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:N/VI:H/VA:N/SC:N/SI:N/SA:N'} for OSV id: 'PYSEC-2024-154'
> 
> Why this score is not supported ?

I have a PR for this 
- #1974

vulnerabilities/importer.py
```diff
@@ -344,18 +406,25 @@ class AffectedPackageV2:
...
-        if not (self.affected_version_range or self.fixed_version_range):
+        if not (
+            self.affected_version_range
+            or self.fixed_version_range
+            or self.affected_by_commits
+            or self.fixed_by_commits
+        ):
             raise ValueError(
                 f"Affected Package {self.package!r} should have either fixed version range or an "
                 "affected version range."
```
keshav-space: Let's also update this error message.

vulnerabilities/importer.py
```diff
@@ -385,6 +456,12 @@ def to_dict(self):
             "package": purl_to_dict(self.package),
             "affected_version_range": affected_version_range,
             "fixed_version_range": fixed_version_range,
+            "affected_by_commits": [
+                affected_by_commit.to_dict() for affected_by_commit in self.affected_by_commits
+            ],
+            "fixed_by_commits": [
+                fixed_by_commit.to_dict() for fixed_by_commit in self.fixed_by_commits
+            ],
```
keshav-space: Much readable.

```suggestion
            "affected_by_commits": [ commit.to_dict() for commit in self.affected_by_commits ],
            "fixed_by_commits": [ commit.to_dict() for commit in self.fixed_by_commits ],
```

vulnerabilities/importer.py
```diff
@@ -417,6 +496,13 @@ def from_dict(cls, affected_pkg: dict):
             package=package,
             affected_version_range=affected_version_range,
             fixed_version_range=fixed_version_range,
+            affected_by_commits=[
+                CodeCommitData.from_dict(affected_by_commit)
+                for affected_by_commit in affected_by_commits
+            ],
+            fixed_by_commits=[
+                CodeCommitData.from_dict(fixed_by_commit) for fixed_by_commit in fixed_by_commits
+            ],
```
keshav-space:
```suggestion
            affected_by_commits=[ CodeCommitData.from_dict(commit) for commit in affected_by_commits ],
            fixed_by_commits=[ CodeCommitData.from_dict(commit) for commit in fixed_by_commits ],
```

vulnerabilities/importers/osv.py
```diff
@@ -188,24 +200,26 @@ def parse_advisory_data_v2(
     )
 
 
-def extract_fixed_versions(fixed_range) -> Iterable[str]:
+def extract_introduced_and_fixed(ranges) -> Tuple[List[str], List[str]]:
```
keshav-space: This don't look right, you're yielding a list of tuples and not a tuple of lists.

vulnerabilities/importers/osv.py
```diff
@@ -188,24 +200,26 @@ def parse_advisory_data_v2(
...
+    [('0', None), (None, '1.6.0')]
 
-    >>> list(extract_fixed_versions(
-    ... {"type": "ECOSYSTEM","events":[{"introduced": "0"},
-    ... {"fixed": "1.0.0"},{"fixed": "9.0.0"}]}))
-    ['1.0.0', '9.0.0']
+    >>> list(extract_introduced_and_fixed(
+    ... {"type": "GIT", "events": [{"introduced": "abc123"},
+    ... {"fixed": "def456"}]}))
+    [('abc123', None), (None, 'def456')]
```
keshav-space: Shouldn't it just return [('abc123', 'def456')]?

keshav-space (changes requested):
@ziadhany Thanks, See comments below. Parsing of introduced version in OSV not working as expected. See the example below:
```python
In [1]: from vulnerabilities.importers import osv

In [2]: raw_data = {
   ...: "schema_version": "1.4.0",
   ...: "id": "GHSA-2hjr-vmf3-xwvp",
   ...: "aliases": [
   ...:   "CVE-2023-49921"
   ...: ],
   ...: "summary": "summary",
   ...: "details": "details",
   ...: "severity": [],
   ...:   "affected": [
   ...:     {
   ...:       "package": { "ecosystem": "Maven", "name": "org.elasticsearch:elasticsearch" },
   ...:       "ranges": [
   ...:         {
   ...:           "type": "ECOSYSTEM",
   ...:           "events": [
   ...:             {"introduced": "0"},
   ...:             {"fixed": "7.17.16"}
   ...:           ]
   ...:         }
   ...:       ]
   ...:     },
   ...:     {
   ...:       "package": { "ecosystem": "Maven", "name": "org.elasticsearch:elasticsearch" },
   ...:       "ranges": [
   ...:         { "type": "ECOSYSTEM",
   ...:         "events": [
   ...:             { "introduced": "8.0.0" },
   ...:             { "fixed": "8.11.2" }
   ...:           ]
   ...:         }
   ...:       ]
   ...:     }
   ...:   ],
   ...:   "references": []
   ...: }

In [3]: advisory_url ="https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2024/07/G
   ...: HSA-2hjr-vmf3-xwvp/GHSA-2hjr-vmf3-xwvp.json"

In [4]: supported_ecosystems= ["maven"]

In [5]: advisory_text=""

In [6]: adv=osv.parse_advisory_data_v2(raw_data,supported_ecosystems,advisory_url,advisory_text)

In [7]: len(adv.affected_packages)
Out[7]: 2

In [8]: af1=adv.affected_packages[0]

In [9]: af2=adv.affected_packages[1]

In [10]: str(af1.affected_version_range)
Out[10]: 'None'

In [11]: str(af1.fixed_version_range)
Out[11]: 'vers:maven/7.17.16'

In [12]:

In [12]: str(af2.affected_version_range)
Out[12]: 'None'

In [13]: str(af2.fixed_version_range)
Out[13]: 'vers:maven/8.11.2'

```
In the above example af2.affected_version_range shouldn't be none we are completely missing `8.0.0`.  

We should also update osv tests to include case where we test parsed advisory data for introduced and fixed version.

vulnerabilities/importers/osv.py
ziadhany: but how are we going to distinguish between introduced and fixed commits?
`(affected_by_commits, fixed_by_commits)`

keshav-space: I don't like how we're returning these fixed/affected version/commit it's clearly not obvious we can either return dict with proper label or use intermediate dataclass.

ziadhany:
> In the above example af2.affected_version_range shouldn't be none we are completely missing `8.0.0`.
> 
> We should also update osv tests to include case where we test parsed advisory data for introduced and fixed version.

We were ignoring the introduced field (not sure why), but we compute the affected_version_range from the `versions` field,
 ex: https://ossf.github.io/osv-schema/#ruby-vulnerability

We should create an issue for that and implement a fallback.

keshav-space:
@ziadhany 

> We should create an issue for that and implement a fallback.  

Fallback is meant to handle unforeseen cases, what we have here is a bug in the core OSV parsing logic, which needs to be fixed.

We should not make changes to the broken parsing logic in this PR, as it will only pile up our technical debt. We should break this PR into two, this PR was only meant to add support for affected and fixed commits in our Advisory data and the corresponding change in our insert advisory function, in line with the model changes introduced in #2007. The second PR should fix the OSV parsing and collect commits.

vulnerabilities/importers/osv.py
```diff
@@ -10,21 +10,26 @@
...
 import dateparser
 from cvss.exceptions import CVSS3MalformedError
 from cvss.exceptions import CVSS4MalformedError
 from packageurl import PackageURL
+from packageurl.contrib.url2purl import url2purl
 from univers.version_range import RANGE_CLASS_BY_SCHEMES
 from univers.versions import InvalidVersion
 from univers.versions import SemverVersion
 from univers.versions import Version
+from websockets.version import commit
```
pombredanne: Why do we use websockets?

ziadhany: I’m not sure why the IDE imported this automatically (it’s not used in the code).

TG1999:
@ziadhany we either need a test pipeline or an actual code commit collection pipeline to see how we are doing. Also there is nothing in current code that talks about SUPPORTED_CODE_COMMIT types for now.

ziadhany:
@TG1999 Okay, I will add the `aosp_dataset` importer (https://github.com/quarkslab/aosp_dataset) since it has some diversity in vcs_urls.

ziadhany:
url2purl currently fails to parse commit hashes correctly:
- https://github.com/package-url/packageurl-python/issues/210

I’ve opened a PR with a fix:
- https://github.com/package-url/packageurl-python/pull/211

vulnerabilities/tests/test_importer.py
```diff
@@ -17,3 +22,172 @@ def test_all_importers_have_unique_name():
...
+            "a1b2c3d4",
+            "diff content",
+            "https://example.com/user/unknown/commits/a1b2c3d4",
+            (
+                None,
+                PatchData(
+                    patch_text="diff content",
+                    patch_url="https://example.com/user/unknown/commits/a1b2c3d4",
+                    vcs_url="https://example.com/user/unknown",
+                    commit_hash="a1b2c3d4",
```
ziadhany: I think we should have `vcs_url` and `commit_hash` fields to the Patch model.

vulnerabilities/tests/test_importer.py
```diff
@@ -17,3 +22,172 @@ def test_all_importers_have_unique_name():
...
+        #     None,
+        #     None,
+        #     (
+        #         None,
+        #         PatchData(
+        #             patch_text=None,
+        #             patch_url=None,
+        #         ),
+        #     ),
+        # ),
```
ziadhany: @TG1999 @keshav-space, I am not sure if we should handle this case.

vulnerabilities/tests/test_importer.py
TG1999: Why to have these fields? If we have these 2 fields it's already a package code patch

vulnerabilities/importer.py
```diff
@@ -194,6 +198,115 @@ def from_url(cls, url):
         return cls(url=url)
 
 
+@dataclasses.dataclass(eq=True)
+@functools.total_ordering
+class PackageCommitPatchData:
+    vcs_url: str
+    commit_hash: str
+    patch_text: Optional[str] = None
+    patch_url: Optional[str] = None
```
TG1999: Why this ? 

https://github.com/aboutcode-org/vulnerablecode/issues/2022#issuecomment-3576671729

When we decided on these models

vulnerabilities/models.py
```diff
@@ -2744,6 +2745,78 @@ def url(self):
...
+        help_text="SHA512 checksum of the patch content.",
+    )
+
+    objects = BaseQuerySet.as_manager()
+
+    class Meta:
+        unique_together = ["patch_checksum", "patch_url"]
+
+
+class PackageCommitPatch(models.Model):
```
TG1999, line 2794: Models are not same with what we discussed earlier.

vulnerabilities/tests/test_importer.py
```diff
@@ -17,3 +22,172 @@ def test_all_importers_have_unique_name():
...
+            (
+                None,
+                PatchData(
+                    vcs_url="https://unsupported.example.com/repo",
+                    commit_hash="112233",
+                    patch_text=None,
+                    patch_url=None,
+                ),
+            ),
+        ),
```
ziadhany, line 62: @TG1999 How are we going to handle this case then? if we don't have vcs_url and commit_hash in the Patch model

TG1999: Patch does not even have vcs URL, it will be stored in patch_url

ziadhany: let’s say we have an OSV example with an unsupported package. Provide only the vcs_url and commit_hash.

how we are gona construct a patch_url ? and if we are just storing vcs_url in patch_url we are missing some data ( commit_hash )

https://github.com/google/oss-fuzz-vulns/blob/4df1da5f533aa777b9511e2ebbbbcdcc354421b2/vulns/cras/OSV-2023-506.yaml#L25

TG1999: Have you read this https://github.com/aboutcode-org/vulnerablecode/issues/2022#issuecomment-3576762471 ?

ziadhany: Yes, I’ve read this multiple times, but I’m a bit confused about it.

TG1999: What are the confusing parts, which are not clear ?

TG1999: > let’s say we have an OSV example with an unsupported package. Provide only the vcs_url and commit_hash.
> 
> 
> 
> how we are gona construct a patch_url ? and if we are just storing vcs_url in patch_url we are missing some data ( commit_hash )
> 
> 
> 
> https://github.com/google/oss-fuzz-vulns/blob/4df1da5f533aa777b9511e2ebbbbcdcc354421b2/vulns/cras/OSV-2023-506.yaml#L25

What does https://github.com/aboutcode-org/vulnerablecode/issues/2022#issuecomment-3576762471 says about this ?

ziadhany: When we have a VCS URL & Commit hash & not code patch
```
Case 2: We can not parse a package
- For the types that we don't support, we will not create PackageCommitPatch for them today, instead store them as Patch with patch URL for now as a fallback. We will also log these kind of URLs. We will have an improver that takes Patch with the advisories associated and tries to create valid PackageCommitPatches. Later when improve our parsing capabilities in url2purl for various VCS URL types, this improver will handle the rest
```

simply how we are gona get patch URL if we don't have a it ?

TG1999: We will store VCS URL as patch URL, and use improvers later,that is what we planned and discussed

ziadhany: Okay, but we’re missing storing the commit_hash.

TG1999: Please read the above sentence, if we can't parse it today. We will use improvers to create commit hash when we improve our code parsing capabilities

ziadhany: Ok, I will update it

vulnerabilities/tests/test_importer.py
TG1999: @ziadhany I am sorry, I did not get you earlier. If you have a VCS URL and also a commit hash and you can't parse it into a package today. Store it as a reference so you do not loose out on commit hash and associate it with the advisory. Thanks!

ziadhany: @TG1999 No worries ,you don’t need to be sorry. We’re trying to come up with the best design. I was thinking about it, and we’re going to lose the commit hash.

ziadhany: why not we store `commit_hash` as optional field in a patch model instead of reference

TG1999: Let's not do that please

vulnerabilities/tests/pipes/test_vulnerablecode_importer_pipeline_v2.py
```diff
@@ -92,4 +125,5 @@ def test_advisory_import_atomicity(dummy_importer):
     dummy_importer.collect_and_store_advisories()
     assert AdvisoryV2.objects.count() == 1
     assert ImpactedPackage.objects.count() == 2
```
TG1999, line 131: Why ImpactedPackage was not created for the PackageCommitPatchData ?

vulnerabilities/pipes/advisory.py
```diff
@@ -96,6 +107,101 @@ def get_or_create_advisory_weaknesses(weaknesses: List[str]) -> List[AdvisoryWea
...
+
+    all_patches = Patch.objects.filter(query)
+    return list(all_patches)
+
+
+VCS_URLS_SUPPORTED_TYPES = {"github", "bitbucket", "gitlab"}
+
+
+def classify_patch_source(url, commit_hash, patch_text):
+    """Classify a patch as a PackageCommitPatchData or PatchData using provided args."""
```
TG1999: Docstring should tell what this function is returning

TG1999:
```
[
{url: https://github.com/abc/def, commit_hash: None, patch_text: None}, 
{url: https://github.com/abc/def, commit_hash: None, patch_text: "+1-2"},
{url: https://github.com/abc/def, commit_hash: "1213",patch_text: None},
{url: https://github.com/abc/def, commit_hash: "1213",patch_text: "+1-2"}, 
{url: "https://github.com/abc/def/commit/12323", commit_hash: None, patch_text: None},
{url: "https://github.com/abc/def/commit/12323", commit_hash: None, patch_text: "+1-2"},
{url: "https://github.com/abc/def/commit/12323", commit_hash: 12323, patch_text: None},
{url: "https://github.com/abc/def/commit/12323", commit_hash: 12323, patch_text:"+1-2"},
{url: "https://unknown.com/abc/def", commit_hash: None, patch_text: None},
{url: "https://unknown.com/abc/def", commit_hash: None, patch_text: "+1-2"},
{url: "https://unknown.com/abc/def", commit_hash: "113324", patch_text: None},
{url: "https://unknown.com/abc/def", commit_hash: "113324", patch_text:"+1-2"},
{url:"https://unknown.com/abc/def/123434", commit_hash:None, patch_text: None},
{url:"https://unknown.com/abc/def/123434", commit_hash:None, patch_text: "+1-2"}, 
{url:"https://unknown.com/abc/def/123434", commit_hash:"123434", patch_text: None}
{url:"https://unknown.com/abc/def/123434", commit_hash:"123434", patch_text: "+1-2"}
]

```

@ziadhany excellent job! One thing I want is to use these test fixtures as mock advisories, I want to see how are we forming ImpactedPackages (base_purl, packagecommitpatches (intorduced/fixed)),  ReferencesV2, Pactch. I want to see something end to end. How are we forming/storing advisories and also how are we relating advisories to references, impacted_packages and patches.

Thanks!

vulnerabilities/tests/pipes/test_vulnerablecode_importer_pipeline_v2.py
ziadhany: I think AffectedPackageV2 is the only component responsible for creating the ImpactedPackage, since it can accurately establish the relationship between the purl and (affected_version_range, fixed_version_range, introduced_by_commit_patches, fixed_by_commit_patches), even when the purl is not matching vcs_url extracted purl .

**The problem is that PackageCommitPatchData cannot determine whether it represents introduced_by_commit_patches or fixed_by_commit_patches.**

If PackageCommitPatchData were responsible for creating ImpactedPackage, we would also lose the ability to correctly associate the affected_version_range and fixed_version_range.

I think the importer pipeline should handle this ( assign the correct purl to the AffectedPackageV2 that have a  (affected_version_range, fixed_version_range, introduced_by_commit_patches, fixed_by_commit_patches) , and we can have two options:

- Validate strictly and raise an error ( if the PackageCommitPatchData extracted purl not matching the AffectedPackageV2 purl )
- Allow some tolerance assuming the pipeline provides a correct purl relying on `classify_patch_source` function which is not used in this test

ziadhany:
@TG1999 I added an end-to-end test. Can you please have a look?

vulnerabilities/importer.py
```diff
@@ -473,6 +600,7 @@ def to_dict(self):
                 "summary": self.summary,
                 "affected_packages": [pkg.to_dict() for pkg in self.affected_packages],
                 "references_v2": [ref.to_dict() for ref in self.references_v2],
+                "patches": [patch.to_dict() for patch in self.patches],
```
keshav-space, line 603: We should also handle patch in AdviosryData.from_dict()

vulnerabilities/importer.py
```diff
@@ -533,6 +661,7 @@ class AdvisoryDataV2:
     summary: Optional[str] = ""
     affected_packages: List[AffectedPackage] = dataclasses.field(default_factory=list)
     references: List[ReferenceV2] = dataclasses.field(default_factory=list)
+    patches: List[PatchData] = dataclasses.field(default_factory=list)
```
keshav-space: @TG1999 We're not using `AdvisoryDataV2` anywhere, IMO we should get rid of dead code if it's no longer needed.

vulnerabilities/importer.py
```diff
@@ -557,6 +686,7 @@ def to_dict(self):
             "summary": self.summary,
             "affected_packages": [pkg.to_dict() for pkg in self.affected_packages],
             "references": [ref.to_dict() for ref in self.references],
+            "patches": [ref.to_dict() for ref in self.patches],
```
keshav-space:
```suggestion
            "patches": [patch.to_dict() for patch in self.patches],
```

vulnerabilities/importer.py
```diff
@@ -574,6 +704,7 @@ def from_dict(cls, advisory_data):
                 if pkg is not None
             ],
             "references": [Reference.from_dict(ref) for ref in advisory_data["references"]],
+            "patches": [PatchData.from_dict(ref) for ref in advisory_data["patches"]],
```
keshav-space:
```suggestion
            "patches": [PatchData.from_dict(patch) for patch in advisory_data["patches"]],
```

vulnerabilities/models.py
```diff
@@ -2744,6 +2748,67 @@ def url(self):
...
+        help_text="SHA512 checksum of the patch content.",
+    )
+
+    def save(self, *args, **kwargs):
+        if self.patch_text:
+            self.patch_checksum = compute_patch_checksum(self.patch_text)
+        super().save(*args, **kwargs)
+
+    class Meta:
+        unique_together = ["patch_checksum", "patch_url"]
```
keshav-space, line 2782: Since both `patch_url` and `patch_text` could be blank we should put some constraint here to make sure we have at least one field to create a valid Patch obj.

vulnerabilities/pipes/advisory.py
```diff
@@ -96,6 +107,124 @@ def get_or_create_advisory_weaknesses(weaknesses: List[str]) -> List[AdvisoryWea
...
+        if (
+            commit_obj.patch_checksum != input_data.patch_checksum
+            or commit_obj.patch_text != input_data.patch_text
+        ):
+            commit_obj.patch_checksum = input_data.patch_checksum
+            commit_obj.patch_text = input_data.patch_text
+            to_update.append(commit_obj)
+
+    if to_update:
+        PackageCommitPatch.objects.bulk_update(to_update, fields=["patch_checksum", "patch_text"])
```
keshav-space, line 147: If for a given pair of `vcs_url` and `commit_hash` we get different `patch_text`, then it breaks our assumption that `vcs_url` and `commit_hash` alone are sufficient to track a commit patch. Whenever we encounter this scenario this should fail loudly and that would mean we need to make `vcs_url`, `commit_hash`, and `patch_text` unique together. We should never silently overwrite the existing `patch_text` if the incoming `patch_text` is different. The only case where we should overwrite `patch_text` is when the existing `patch_text` field is empty.

vulnerabilities/pipes/advisory.py
```diff
@@ -96,6 +107,124 @@ def get_or_create_advisory_weaknesses(weaknesses: List[str]) -> List[AdvisoryWea
...
+
+    if not commit_hash and not purl.version:
+        return None, PatchData(patch_url=url, patch_text=patch_text or None)
+
+    base_purl = get_core_purl(purl)
+    base_purl_str = base_purl.to_string()
+    base_url = purl2url(base_purl_str)
+    return base_purl, PackageCommitPatchData(
+        vcs_url=base_url, commit_hash=purl.version or commit_hash, patch_text=patch_text or None
+    )
```
keshav-space: This is not handling the case where we don't have `purl.version` but we do have `commit_hash`?
Example:
vcs_url = "https://github.com/aboutcode-org/vulnerablecode"
commit_hash = "520531d341790022c0c9297c30fc93ff0e94266c"

vulnerabilities/importer.py
TG1999: @keshav-space agreed!

ziadhany: @TG1999 @keshav-space You mean we should get rid of `AdvisoryData`, not `AdvisoryDataV2`?

AdvisoryData is the old version, am I right?

vulnerabilities/pipes/advisory.py
```diff
@@ -96,6 +107,124 @@ def get_or_create_advisory_weaknesses(weaknesses: List[str]) -> List[AdvisoryWea
...
+    if not url:
+        if not patch_text:
+            return
+
+        return None, PatchData(patch_text=patch_text)
+
+    purl = url2purl(url)
+    if not purl or (purl.type not in VCS_URLS_SUPPORTED_TYPES):
+        if commit_hash:
+            return None, ReferenceV2(
```
TG1999: Why are we returning here? Are not we missing on patch_text like this ?

TG1999: We discussed that we will create both reference and Patch here, not just one of them

vulnerabilities/pipes/advisory.py
ziadhany: @keshav-space we have a test for this `vulnerabilities/tests/test_importer.py` L:33:41

https://github.com/aboutcode-org/vulnerablecode/pull/2017/files/9b8cd079df0b4065f5779e5244233ca01592df2f#diff-30adc9b685adde4729b483a5efb2c0a27134d3f3ee2b16db3c19923b88c96a0b

keshav-space: Okay, I missed that we're using `purl.version or commit_hash` while passing args. There is no harm in having few extra variables if it helps with code readability.
We can have something like this:
```python
package_commit_hash = purl.version or commit_hash
```

ziadhany: OK, I’ll update it and make it more readable.

vulnerabilities/importer.py
keshav-space: @ziadhany  We are not using `AdvisoryDataV2` anywhere, for both v1 and v2 we're using `AdvisoryData`.

ziadhany: Ok, I will drop `AdvisoryDataV2`

vulnerabilities/pipes/advisory.py
ziadhany: I updated the function to return both the reference and the patch

ziadhany:
@keshav-space @TG1999 I’ve added the requested changes. Please have a look and let me know if I should update anything.

AospImporterPipeline Logs:
```
Importing data using aosp_dataset_fix_commits
INFO 2025-12-11 09:52:54.675302 UTC Pipeline [AospImporterPipeline] starting
INFO 2025-12-11 09:52:54.675437 UTC Step [clone] starting
INFO 2025-12-11 09:52:54.675488 UTC Cloning `git+https://github.com/quarkslab/aosp_dataset`
INFO 2025-12-11 09:52:57.279282 UTC Step [clone] completed in 3 seconds
INFO 2025-12-11 09:52:57.279391 UTC Step [collect_and_store_advisories] starting
INFO 2025-12-11 09:52:57.294661 UTC Collecting 1,994 advisories
INFO 2025-12-11 09:52:57.294748 UTC Processing aosp_dataset fix commits.
INFO 2025-12-11 09:52:58.749586 UTC Progress: 10% (200/1994) ETA: 13 seconds
INFO 2025-12-11 09:53:00.234658 UTC Progress: 20% (399/1994) ETA: 12 seconds
INFO 2025-12-11 09:53:01.706514 UTC Progress: 30% (599/1994) ETA: 10 seconds
INFO 2025-12-11 09:53:02.664337 UTC Progress: 40% (798/1994) ETA: 8 seconds
INFO 2025-12-11 09:53:04.016003 UTC Progress: 50% (997/1994) ETA: 7 seconds
INFO 2025-12-11 09:53:05.483900 UTC Progress: 60% (1197/1994) ETA: 5 seconds
INFO 2025-12-11 09:53:06.949434 UTC Progress: 70% (1396/1994) ETA: 4 seconds
INFO 2025-12-11 09:53:08.425321 UTC Progress: 80% (1596/1994) ETA: 3 seconds
INFO 2025-12-11 09:53:09.726271 UTC Progress: 90% (1795/1994) ETA: 1 seconds
INFO 2025-12-11 09:53:11.224522 UTC Successfully collected 1,992 advisories
INFO 2025-12-11 09:53:11.224660 UTC Step [collect_and_store_advisories] completed in 14 seconds
INFO 2025-12-11 09:53:11.224715 UTC Step [clean_downloads] starting
INFO 2025-12-11 09:53:11.224761 UTC Removing cloned repository
INFO 2025-12-11 09:53:11.254953 UTC Step [clean_downloads] completed in 0 seconds
INFO 2025-12-11 09:53:11.255088 UTC Pipeline completed in 17 seconds

Process finished with exit code 0```

File: vulnerabilities/models.py [hunk 28]
```diff
@@ -2744,6 +2748,73 @@ def url(self):
...
+    patch_checksum = models.CharField(
+        max_length=128,
+        blank=True,
+        null=True,
+        help_text="SHA512 checksum of the patch content.",
+    )
+
+    def clean(self):
+        if not self.patch_url and not self.patch_text:
+            raise ValidationError("Either patch_url or patch_text must be provided.")
```
keshav-space: This is not the right approach. Here we're relying on `save()` for data integrity, but model-level validation including save() are not applied during bulk operations. Instead of model level validation we should add the constraint at the database level.

vulnerabilities/models.py
```diff
@@ -2744,6 +2748,73 @@ def url(self):
...
+
+    def save(self, *args, **kwargs):
+        # https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.clean
+        self.full_clean()
```
keshav-space: Also this is not needed for database level constraint

ziadhany: you mean check constraint like this https://docs.djangoproject.com/en/5.2/ref/models/constraints/#checkconstraint

    class Meta:
        constraints = [
            models.CheckConstraint(check=(Q(patch_url__isnull=False) | Q(patch_text__isnull=False)),
                name="patch_url_or_patch_text",
            )
        ]

keshav-space: yes

vulnerabilities/models.py
ziadhany: Also, for `patch_checksum`, should we use a database trigger, or just rely on save() and the importer to handle bulk operations?

keshav-space: No need for db trigger ATM, we should supply the `patch_checksum` while doing bulk create.

vulnerabilities/models.py
```diff
@@ -2744,6 +2748,73 @@ def url(self):
...
+    def save(self, *args, **kwargs):
+        if self.patch_text:
+            self.patch_checksum = compute_patch_checksum(self.patch_text)
+        super().save(*args, **kwargs)
+
+    class Meta:
+        unique_together = ["patch_checksum", "patch_url"]
+        constraints = [
+            models.CheckConstraint(
+                check=(Q(patch_url__isnull=False) | Q(patch_text__isnull=False)),
```
keshav-space: Both cannot be null, and both cannot be blank.

vulnerabilities/models.py
```diff
@@ -2744,6 +2748,74 @@ def url(self):
...
+        if self.patch_text:
+            self.patch_checksum = compute_patch_checksum(self.patch_text)
+        super().save(*args, **kwargs)
+
+    class Meta:
+        unique_together = ["patch_checksum", "patch_url"]
+        constraints = [
+            models.CheckConstraint(
+                check=(Q(patch_url__isnull=False) | Q(patch_text__isnull=False))
+                & ~(Q(patch_url="") & Q(patch_text="")),
```
keshav-space: @ziadhany current constraint will not throw an error if one field null and other field is blank. For example if we try to do `Patch.objects.create(patch_url="", patch_text=None)`, this database entry will be created successfully, which should not happen.

We should have something like this as constraint:
```suggestion
                check=(
                    Q(patch_url__isnull=False) & ~Q(patch_url="") |
                    Q(patch_text__isnull=False) & ~Q(patch_text="")
                ),
```

ziadhany: You're right. I’ll update it and add a simple test for it.

TG1999:
@ziadhany LGTM! please rebase and adjust the migrations! great work :raised_hands:

ziadhany:
> @ziadhany LGTM! please rebase and adjust the migrations! great work 🙌

@TG1999 Done , Please merge 🚀

PR #2046, opened 2025-11-26
Add V2_importer to collect advisories from EUVD

Samk1710:
- Fixes: [https://github.com/aboutcode-org/vulnerablecode/issues/1873](https://github.com/aboutcode-org/vulnerablecode/issues/1873)

### Data Source

* **API**: `https://euvdservices.enisa.europa.eu/api/search`
* **Format**: JSON with pagination

Dependent on : 
- https://github.com/aboutcode-org/aboutcode-mirror-euvd/pull/1

Logs:

```text
INFO 2025-11-25 23:53:04.724015 UTC Progress: 100% (452459/452459)
INFO 2025-11-25 23:53:04.735236 UTC Successfully collected 452,441 advisories
INFO 2025-11-25 23:53:04.735444 UTC Step [collect_and_store_advisories] completed in 23064 seconds (6.4 hours)
```


vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,218 @@
...
+    def advisories_count(self) -> int:
+        return len(self.fetch_data())
+
+    def collect_advisories(self) -> Iterable[AdvisoryData]:
+        for raw_data in self.fetch_data():
+            try:
+                advisory = self.parse_advisory(raw_data)
+                if advisory:
+                    yield advisory
+            except Exception as e:
```
ziadhany: Please avoid using general exceptions.

vulnerabilities/tests/pipelines/v2_importers/test_euvd_importer_v2.py
```diff
@@ -0,0 +1,124 @@
...
+        first = advisories[0]
+        assert isinstance(first, AdvisoryData)
+        assert first.advisory_id == "EUVD-2025-197757"
+        assert "EUVD-2025-197757" in first.aliases
+        assert "CVE-2025-13284" in first.aliases
+        assert first.summary == "ThinPLUS vulnerability that allows remote code execution"
+        assert first.date_published is not None
+        assert len(first.severities) == 1
+        assert first.severities[0].system.identifier == "cvssv3.1"
+        assert first.severities[0].value == "9.8"
```
ziadhany: I think it would be easier if you test using `util_tests.check_results_against_json(result, expected_file)` and with an expected file.

vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @classmethod
+    def steps(cls):
+        return (cls.collect_and_store_advisories,)
+
+    def fetch_data(self):
+        # Return cached data if already fetched
+        if self._cached_data is not None:
+            logger.info(f"Using cached data: {len(self._cached_data)} items")
+            return self._cached_data
```
ziadhany, line 53: Why do we have _cached_data? It is because the API returns repeated data

vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,218 @@
...
+
+        headers = {"User-Agent": "VulnerableCode"}
+        all_items = []
+        page = 0
+        size = 100
+        max_retries = 100
+
+        logger.info(f"Fetching data from EUVD API: {self.url}")
+
+        while True:
```
ziadhany: We should avoid loops without a condition. Maybe looping over the total 452584 advisories is a good idea.

vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,218 @@
...
+            while retry_count < max_retries and not success:
+                try:
+                    params = {"size": size, "page": page}
+                    response = requests.get(self.url, headers=headers, params=params, timeout=30)
+
+                    if response.status_code != HTTPStatus.OK:
+                        logger.error(f"API returned status {response.status_code} for page {page}")
+                        retry_count += 1
+                        if retry_count < max_retries:
+                            sleep_time = min(10 * (2 ** min(retry_count - 1, 5)), 60)
```
ziadhany: Why this sleep_time? We run the importers multiple times. If one request fails, we can have just one retry.

( please avoid complex retry )

ziadhany (changes requested):
@Samk1710 Great start! Just a few small tweaks

Samk1710:
> @Samk1710 Great start! Just a few small tweaks

Thanks a lot @ziadhany for the review. Will make the changes as suggested.

vulnerabilities/pipelines/v2_importers/euvd_importer.py
Samk1710: `_cached_data` prevents a second full API fetch.
The base importer calls `fetch_data()` once to count advisories and again to iterate through them.
 Caching ensures both steps use the same dataset snapshot while avoiding duplicated network requests and API load.

vulnerabilities/pipelines/v2_importers/euvd_importer.py
Samk1710: Screenshot from 2025-11-25 20-25-53

I added the retry logic as I faced some API failures during importer run (in my case it was network failures), but as you mentioned that we run importers multiple times, it shouldn't be an issue. I will apply one retry on failure as suggested.

Samk1710:
Hey @ziadhany ,

I’ve pushed the requested updates. Summary of changes:

* Replaced broad exception handling with specific parsing-related exceptions.
* Updated the test to use `util_tests.check_results_against_json` along with expected JSON fixtures.
* Removed unbounded loops by using the API’s total count field to determine total number of pages.
* Simplified retry handling(Retains a fixed 3-second delay between retries to prevent rapid page skipping in case of network interruptions).

Let me know if you’d like any additional modifications. Thanks again for the feedback and guidance!

vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,244 @@
...
+
+class EUVDImporterPipeline(VulnerableCodeBaseImporterPipelineV2):
+    """
+    EUVD (EU Vulnerability Database) Importer Pipeline
+
+    This pipeline imports security advisories from the European Union Vulnerability Database (EUVD).
+    """
+
+    pipeline_id = "euvd_importer_v2"
+    spdx_license_expression = "LicenseRef-scancode-other-permissive"
```
AyanSinhaMahapatra:
```suggestion
    spdx_license_expression = "CC-BY-4.0"
```
This is CC-by-4.0 as per:
- https://www.enisa.europa.eu/sites/default/files/2024-12/enisa-ipr-policy-public-version.pdf
- https://www.enisa.europa.eu/about-enisa/legal-notice

Samk1710: Thanks a lot Ayan

Samk1710:
I have updated the License Expression and added sample test data from the EUVD API as suggested in today's call.

Samk1710:
Hey @pombredanne @ziadhany 
Since caching is not a feasible option, we could use the "total" field returned by the API to populate the total advisories count in advisories_count function.

"total": 452844

This would mean the importer would take 5-6 hours and fetch data(only once) in collect_advisories.

ziadhany:
@Samk1710 Yes, we can use the total field. Since we know the total number of advisories, we can iterate over the endpoint using either the date or the advisory count, if that’s available. 

the time isn’t a big issue to get all the available data in under a couple of hours.

Samk1710:
> @Samk1710 Yes, we can use the total field. Since we know the total number of advisories, we can iterate over the endpoint using either the date or the advisory count, if that’s available.
> 
> the time isn’t a big issue to get all the available data in under a couple of hours.

yes the date field is available in the API response and if time(5-6 hours) ain't a issue, this would be the simplest approach to avoid double fetching and caching will not be required. Shall I move forward with this approach?

keshav-space:
@Samk1710 here is the repo for the EUVD mirror https://github.com/aboutcode-org/aboutcode-mirror-euvd. For data collection script take a look at the pipeline here https://github.com/aboutcode-org/aboutcode-mirror-nuget-catalog/blob/main/sync_catalog.py. We want to do something similar for EUVD the script will be used in a workflow that will be almost identical to what we have here https://github.com/aboutcode-org/aboutcode-mirror-nuget-catalog/blob/main/.github/workflows/sync.yml

Samk1710:
> @Samk1710 here is the repo for the EUVD mirror https://github.com/aboutcode-org/aboutcode-mirror-euvd. For data collection script take a look at the pipeline here https://github.com/aboutcode-org/aboutcode-mirror-nuget-catalog/blob/main/sync_catalog.py. We want to do something similar for EUVD the script will be used in a workflow that will be almost identical to what we have here https://github.com/aboutcode-org/aboutcode-mirror-nuget-catalog/blob/main/.github/workflows/sync.yml

thanks a lot @keshav-space. will look into it

Samk1710:
Hey, I have added a pipeline to mirror the EUVD advisories

https://github.com/aboutcode-org/aboutcode-mirror-euvd/pull/1

Would look forward to a review so that I can continue this importer. Thanks a lot!

vulnerabilities/tests/pipelines/v2_importers/test_euvd_importer_v2.py
```diff
@@ -0,0 +1,78 @@
...
+        ]
+        mock_get.side_effect = mock_responses
+
+        pipeline = EUVDImporterPipeline()
+        advisories = [data.to_dict() for data in list(pipeline.collect_advisories())]
+
+        expected_file = TEST_DATA / "euvd-expected.json"
+        util_tests.check_results_against_json(advisories, expected_file)
+
+    def test_get_scoring_system(self):
```
pombredanne: Please create one test function for each

vulnerabilities/pipelines/v2_importers/euvd_importer.py
```diff
@@ -0,0 +1,244 @@
...
+        logger.info(f"Fetching data from EUVD API: {self.url}")
+
+        total_count = self._fetch_total_count(size, max_retries)
+        if total_count is None:
+            logger.error("Failed to fetch total count from API")
+            return all_items
+
+        total_pages = math.ceil(total_count / size)
+        logger.info(f"Total advisories: {total_count}, Total pages: {total_pages}")
+
```
ziadhany, line 68: @Samk1710 Please update the code. There’s no need to fetch the API, we can clone the mirror and start parsing it.

Samk1710: Working on it. Will push the updates once my mid-sem exams are over this week.

Samk1710:
The product field is a little messy. We might need to drop some(less than 1 %).

For eg:
- https://euvd.enisa.europa.eu/vulnerability/EUVD-2023-12722
- https://euvd.enisa.europa.eu/vulnerability/EUVD-2024-16938
- https://euvd.enisa.europa.eu/vulnerability/EUVD-2023-39960
- https://euvd.enisa.europa.eu/vulnerability/EUVD-2017-8416

PR #2048, opened 2025-11-26
Move ssvc_calculator to utils.py to avoid code duplication

suji7724:
Moved the 'ssvc_calculator' function from 
'v2_importers/vulnrichment_importer.py' to 
'vulnerabilities/utils.py' because it is used in multiple places.

Updated imports in the original file. All tests pass.


keshav-space:
This has already been addressed in https://github.com/aboutcode-org/vulnerablecode/pull/2050/

PR #2050, opened 2025-11-28
Collect SSVC trees

TG1999:
This PR resolves:
- https://github.com/aboutcode-org/vulnerablecode/issues/1963

It adds support to collect, store and display SSVC vulnerability scorings.

These are displayed in a tab in the UI:
Screenshot from 2025-12-15 22-48-22

... and also reported in API results both as a score vector and as textual values:

Screenshot from 2025-12-15 22-48-05

### DIFF
--- a/vulnerabilities/tests/pipelines/v2_improvers/test_collect_ssvc_trees.py
+++ b/vulnerabilities/tests/pipelines/v2_improvers/test_collect_ssvc_trees.py
@@ -7,6 +7,8 @@
 # See https://aboutcode.org for more information about nexB OSS projects.
 #
 
+from datetime import datetime
+
 import pytest
 
 from vulnerabilities.models import SSVC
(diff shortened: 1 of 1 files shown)


vulnerabilities/templates/advisory_detail.html
```diff
@@ -543,6 +450,45 @@
...
+                          <a href="{{ ssvc.advisory.get_absolute_url }}">
+                            {{ ssvc.advisory.avid }}
+                            <i class="fa fa-external-link fa_link_custom"></i>
+                          </a>
+                        </p>
+                          <details>
+                            <summary class="is-size-7 has-text-link" style="cursor: pointer;">
+                              View SSVC decision tree
+                            </summary>
+                            <pre>{{ ssvc.options|pprint }}</pre>
```
pombredanne, line 482: We should use saneyaml dump instead, not a pprint

PR #2065, opened 2025-12-19
Fix broken ScanCode documentation link in contributing.rst

VaaishnaviS:
This PR fixes the broken ScanCode documentation link (related to issue aboutcode-org/vulnerablecode #4645).
Signed-off-by: Vaishnavi S  <vaishnavibabblu1@gmail.com>

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -73,7 +73,7 @@ overlooked. We value any suggestions to improve
 
 .. tip::
     Our documentation is treated like code. Make sure to check our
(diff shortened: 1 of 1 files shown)


keshav-space:
Thanks @VaaishnaviS. This is already fixed, we now have up to date URL here https://github.com/aboutcode-org/vulnerablecode/blob/2dbbd3827759246af00d0112dac4affad31cf474/docs/source/contributing.rst?plain=1#L76 Also we don't need to use web.archive.org for our own documentation URLs.

PR #2114, opened 2026-01-17
Enforce LF line endings to fix Windows Docker setup

Ashishworks:
Fixes #2100

Windows clones may convert LF to CRLF, causing Docker/Linux to fail with python\r in shebang lines (e.g., manage.py).
This PR enforces LF line endings using .gitattributes and renormalizes files accordingly.

Tested:

docker compose run vulnerablecode ./manage.py import --list works on Windows with Docker.

PR #2124, opened 2026-01-24
Fix: Add missing SECRET_KEY and ALTCHA_HMAC_KEY to docker.env

paarthbhatt:
Using the docker compose setup fails locally because `SECRET_KEY` and `ALTCHA_HMAC_KEY` are missing from `docker.env`.
This PR adds default values ("secret") to `docker.env` to ensure the local docker environment starts correctly out of the box.

Note: Line ending fixes are being handled in separate PR #2114.


keshav-space:
@paarthbhatt did you even put in the basic effort to go through our [README](https://github.com/aboutcode-org/vulnerablecode/blob/main/README.rst)? There are clear instructions on how to run VulnerableCode in docker see https://github.com/aboutcode-org/vulnerablecode?tab=readme-ov-file#run-with-docker. You are supposed to run `make envfile` before running the docker commands, which takes care of these secret keys. Please do not waste the time of maintainers with sloppy PR.

PR #2210, opened 2026-03-14
Add LibreOffice security advisories importer

### LINKED ISSUE #1898: Collect data from https://www.libreoffice.org/about-us/security/advisories/

NucleiAv:
- Closes #1898

Adds a pipeline importer for LibreOffice security advisories.

Instead of scraping individual advisory pages with BeautifulSoup which is brittle and hardcoded, and breaks whenever the site layout or UI changes, I used a different and a better approach. The importer fetches the advisory listing page, extracts CVE IDs, then calls the CVE API at https://cveawg.mitre.org/api/cve/{cve_id} for each one. This works because every LibreOffice advisory page links to the CVE record on https://www.cve.org/CVERecord?id={cve_id} in its references section, and the cveawg API returns the full structured CVE 5.0 JSON with CVSS scores, CWE weaknesses, references, and publish dates.

### DIFF
--- a/vulnerabilities/tests/pipelines/v2_importers/test_libreoffice_importer.py
+++ b/vulnerabilities/tests/pipelines/v2_importers/test_libreoffice_importer.py
@@ -16,30 +16,22 @@ from vulnerabilities.pipelines.v2_importers.libreoffice_importer import parse_ad
 from vulnerabilities.pipelines.v2_importers.libreoffice_importer import parse_advisory_urls
 
 LISTING_HTML = """
-<p><a href="/about-us/security/advisories/cve-2025-1080/">CVE-2025-1080</a> Macro URL</p>
-<p><a href="/about-us/security/advisories/cve-2023-2255/">CVE-2023-2255</a> Macro URL</p>
-<p><a href="/about-us/security/advisories/cve-2023-4863/">CVE-2023-4863</a> Heap buffer overflow</p>
+<a href="/about-us/security/advisories/cve-2025-1080/">CVE-2025-1080</a>
+<a href="/about-us/security/advisories/cve-2023-2255/">CVE-2023-2255</a>
+<a href="/about-us/security/advisories/cve-2023-4863/">CVE-2023-4863</a>
 """
 
 ADVISORY_HTML = """\
 <html><body id="cve-2025-1080">
-<section id="content1"><div class="container"><article>
+<section id="content1">
 <div class="row col-sm-10 margin-20">
-<ul class="breadcrumb">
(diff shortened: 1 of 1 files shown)


vulnerabilities/pipelines/v2_importers/libreoffice_importer.py
```diff
@@ -0,0 +1,154 @@
...
+from vulnerabilities.importer import ReferenceV2
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.pipelines import VulnerableCodeBaseImporterPipelineV2
+from vulnerabilities.severity_systems import SCORING_SYSTEMS
+from vulnerabilities.utils import get_cwe_id
+
+logger = logging.getLogger(__name__)
+
+ADVISORIES_URL = "https://www.libreoffice.org/about-us/security/advisories/"
+CVE_API_URL = "https://cveawg.mitre.org/api/cve/{cve_id}"
```
ziadhany: @NucleiAv This is incorrect. You are using two data sources: `https://cveawg.mitre.org` and `https://www.libreoffice.org/about-us/security/advisories/`. We should only use `https://www.libreoffice.org/about-us/security/advisories/`.

If `https://www.libreoffice.org/about-us/security/advisories/` does not provide an API (feel free to do a deep search to confirm this), you should parse the HTML instead. Please take a look at other importers, such as the nginx importer: `nginx_importer_v2.NginxImporterPipeline`.

NucleiAv: @ziadhany sure! I will incorporate the changes, but I thought that HTML parsing wont work if the website layout changes. Moreover libreoffice website does not provide details like CVSS score, CVSS version(2.0, 3.x, 4.0), Severity or CWEs, etc. To populate those details I those to use the api approach. I will research again if libreoffice provides an api and if not, will modify the code to do html parsing.

(below, no details regarding CVSS, CWE, etc is mentioned in the website)
image

NucleiAv:
@ziadhany I researched but could not find any API. So, I have switched to HTML parsing using BeautifulSoup, following the same pattern as the nginx importer. The importer now fetches the listing page to extract advisory URLs, then fetches each individual advisory page and parses the available fields like CVE ID, title, announced date, description, and references. As informed above, a few fields are not available on LibreOffice's pages and will remain empty like CVSS versions, CVSS scores, severity ratings, CWE IDs, and affected version ranges. LibreOffice only lists the fixed version. If those are needed in future, NVD enrichment would be a separate step.

PR #2219, opened 2026-03-16
Add Eclipse Foundation security advisories importer

### LINKED ISSUE #1495: Collect data from https://www.eclipse.org/security/known.php and https://www.eclipse.org/security/

NucleiAv:
Closes #1495

HTML parsing the Eclipse website at https://www.eclipse.org/security/known.php would return nothing useful because the CVE-ID links in the table redirect to NVD and the project links return a 404. Instead I found that the website renders its advisory table entirely via JavaScript from a JSON API endpoint. By inspecting the JS bundle, I found https://api.eclipse.org/cve which returns all 197 advisories as a clean JSON array with no auth and no pagination required. So I choose API approach, since its clean and consistent. 

Each entry provides details like CVE ID, publish date, Eclipse project name, summary, CVSS score, and reference URLs (CVE Mitre link, Eclipse bugtracker ticket, GitHub CVE pull request where applicable).

API does not provide CVSS vector string or version (only a bare float score, stored using GENERIC scoring system), CWE weaknesses, and affected or fixed package versions. These fields are left empty.

### DIFF
--- a/vulnerabilities/pipelines/v2_importers/eclipse_importer.py
+++ b/vulnerabilities/pipelines/v2_importers/eclipse_importer.py
@@ -1,0 +1,104 @@
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# VulnerableCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/aboutcode-org/vulnerablecode for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+import json
+import logging
+from typing import Iterable
+
+import dateparser
+import requests
+
+from vulnerabilities.importer import AdvisoryDataV2
+from vulnerabilities.importer import ReferenceV2
+from vulnerabilities.importer import VulnerabilitySeverity
+from vulnerabilities.pipelines import VulnerableCodeBaseImporterPipelineV2
+from vulnerabilities.severity_systems import GENERIC
+
+logger = logging.getLogger(__name__)
+
+ECLIPSE_API_URL = "https://api.eclipse.org/cve"
+
+
+class EclipseImporterPipeline(VulnerableCodeBaseImporterPipelineV2):
+    """Collect Eclipse Foundation security advisories via the Eclipse CVE API."""
+
+    pipeline_id = "eclipse_importer"
+    spdx_license_expression = "LicenseRef-scancode-proprietary-license"
+    license_url = "https://www.eclipse.org/security/"
+    precedence = 200
+
+    @classmethod
+    def steps(cls):
+        return (
+            cls.fetch,
+            cls.collect_and_store_advisories,
+        )
+
+    def fetch(self):
+        self.log(f"Fetch `{ECLIPSE_API_URL}`")
+        resp = requests.get(ECLIPSE_API_URL, timeout=30)
+        resp.raise_for_status()
+        self.advisories_data = resp.json()
+
+    def advisories_count(self):
+        return len(self.advisories_data)
+
+    def collect_advisories(self) -> Iterable[AdvisoryDataV2]:
+        for entry in self.advisories_data:
+            advisory = parse_advisory(entry)
+            if advisory:
+                yield advisory
+
+
+def parse_advisory(entry: dict):
+    advisory_id = entry.get("id") or ""
+    if not advisory_id:
+        return None
+
+    date_published = None
+    raw_date = entry.get("date_published") or ""
+    if raw_date:
+        date_published = dateparser.parse(
+            raw_date,
+            settings={"TIMEZONE": "UTC", "RETURN_AS_TIMEZONE_AWARE": True, "TO_TIMEZONE": "UTC"},
+        )
+        if date_published is None:
+            logger.warning("Could not parse date %r for %s", raw_date, advisory_id)
+
+    summary_obj = entry.get("summary")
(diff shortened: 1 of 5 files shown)


vulnerabilities/pipelines/v2_importers/eclipse_importer.py
```diff
@@ -0,0 +1,104 @@
...
+    if raw_date:
+        date_published = dateparser.parse(
+            raw_date,
+            settings={"TIMEZONE": "UTC", "RETURN_AS_TIMEZONE_AWARE": True, "TO_TIMEZONE": "UTC"},
+        )
+        if date_published is None:
+            logger.warning("Could not parse date %r for %s", raw_date, advisory_id)
+
+    summary_obj = entry.get("summary")
+    summary = summary_obj.get("content") or "" if isinstance(summary_obj, dict) else ""
```
ziadhany: I think something like this is simpler.

```suggestion
    summary_obj = entry.get("summary", {})
    summary = summary_obj.get("content") or ""
```

vulnerabilities/pipelines/v2_importers/eclipse_importer.py
```diff
@@ -0,0 +1,104 @@
...
+        entry.get("live_link") or "",
+        entry.get("request_link") or "",
+        entry.get("cve_pull_request") or "",
+    ]:
+        if url:
+            references.append(ReferenceV2(url=url))
+
+    severities = []
+    cvss = entry.get("cvss")
+    if cvss is not None:
```
ziadhany:
```suggestion
    if cvss:
```

vulnerabilities/pipelines/v2_importers/eclipse_importer.py
```diff
@@ -0,0 +1,104 @@
...
+        for entry in self.advisories_data:
+            advisory = parse_advisory(entry)
+            if advisory:
+                yield advisory
+
+
+def parse_advisory(entry: dict):
+    advisory_id = entry.get("id") or ""
+    if not advisory_id:
+        return None
```
ziadhany: Do we have any examples of this? If yes, please log this, if no, please remove it.

```suggestion
```

vulnerabilities/tests/pipelines/v2_importers/test_eclipse_importer.py
```diff
@@ -0,0 +1,116 @@
...
+
+with open(TEST_DATA / "eclipse_api_sample.json") as f:
+    SAMPLE_DATA = json.load(f)
+
+ENTRY_WITH_CVSS = SAMPLE_DATA[0]
+ENTRY_WITHOUT_CVSS = SAMPLE_DATA[1]
+ENTRY_WITHOUT_SUMMARY = SAMPLE_DATA[2]
+
+
+class TestParseAdvisory(TestCase):
```
ziadhany: Please change this to test against the file instead of running `parse_advisory` for every attribute.

ziadhany (changes requested):
@NucleiAv Good start, see some comments there.

vulnerabilities/pipelines/v2_importers/eclipse_importer.py
NucleiAv: Used `entry.get("summary") or {}` instead of the suggested `entry.get("summary", {})` because the API actually returns `"summary": null` for some entries (CVE-2024-2212 in the sample data). When the key exists but its value is null, `.get("summary", {})` still returns None, which makes `None.get("content")` blow up. The `or {}` handles both the missing key and the null case cleanly.

vulnerabilities/pipelines/v2_importers/eclipse_importer.py
NucleiAv: Done!

vulnerabilities/pipelines/v2_importers/eclipse_importer.py
NucleiAv: I removed it since I did not find any instances in API data

vulnerabilities/tests/pipelines/v2_importers/test_eclipse_importer.py
NucleiAv: Replaced the whole class with a single `test_parse_advisories()` that runs all three sample entries and checks against an expected JSON file using `util_tests.check_results_against_json`. Also dropped `test_collect_advisories_skips_on_http_error` as it had assert not `hasattr(...) or True` which is always true and wasn't actually testing anything.

keshav-space:
I doubt there is any benefit in having this importer if there is not much information to import. There is not even a complete summary. We just have CVE and a link to the corresponding GHSA advisory, which we are already importing.

For instance, let us take a look at one of those entries in https://api.eclipse.org/cve.

```json
{
    "id": "CVE-2026-1605",
    "date_published": "2026-03-05",
    "project": "rt.jetty",
    "request_link": "https://gitlab.eclipse.org/security/cve-assignment/-/issues/79",
    "cve_pull_request": "",
    "status": "PUBLIC",
    "summary": {
    "content": "The Eclipse Jetty Server Artifact has a Gzip request memory leak ",
    "source": "https://api.github.com/advisories?cve_id=CVE-2026-1605"
    },
    "cvss": 7.5,
    "live_link": "https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2026-1605"
},
```
We do not have anything substantial here to import. Also we already have much better data for this vulnerability from GHSA, see https://public.vulnerablecode.io/advisories/github_osv_importer_v2/GHSA-xxh7-fcf3-rj7f.

PR #2226, opened 2026-03-19
Add Photon OS importer pipeline #1586

### LINKED ISSUE #1586: Datasource Request: Photon OS Advisories
Upstream Source: https://github.com/vmware/photon/wiki/Security-Advisories (CC BY-SA 4.0.)

Republished in OSV at https://github.com/captn3m0/photon-os-advisories, enriched using https://packages.vmware.com/photon/photon_cve_metadata/

Sample advisory URL: https://captnemo.in/photon-os-advisories/advisories/PHSA-2024-4.0-0685.json

Shivaay26:
Fixes #1586

## Summary
Add an importer pipeline for Photon OS security advisories using the
OSV-format feed republished at https://github.com/captn3m0/photon-os-advisories

## Changes
* Add `PhotonOSImporterPipeline` in `vulnerabilities/pipelines/photon_os_importer.py`
* Register pipeline in `vulnerabilities/importers/__init__.py`
* Add unit tests and test data in `vulnerabilities/tests/`

## Notes
* Uses PHSA ID as the only alias to avoid cross-version alias conflicts
  since the same CVE appears in separate advisories for each
  Photon OS version (3.0, 4.0, 5.0)
* CVEs stored as references instead of aliases to preserve the data
* Handles inconsistent ranges format (dict vs list) present in upstream data
* Deduplicates affected package entries — same purl appears multiple
  times per advisory with and without version data

### DIFF
--- a/vulnerabilities/tests/pipelines/test_photon_importer.py
+++ b/vulnerabilities/tests/pipelines/test_photon_importer.py
@@ -1,0 +1,69 @@
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# VulnerableCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/aboutcode-org/vulnerablecode for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+from pathlib import Path
+
+from vulnerabilities.pipelines.photon_os_importer import normalize_ranges
+from vulnerabilities.pipelines.photon_os_importer import parse_photon_advisory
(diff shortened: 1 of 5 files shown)


ziadhany:
@Shivaay26, we already have an OSV parser, so there's no need to implement one from scratch.

also , It looks like the data from this source is just auto-generated from other sources,
`https://github.com/captn3m0/photon-os-advisories/blob/main/generate.py`

I think we should handle Photon OS by following the implementation in this PR: `https://github.com/aboutcode-org/vulnerablecode/pull/2198`

I really appreciate the effort and thanks for giving this a try!.

Shivaay26:
Thanks a lot for the detailed comment.

And I didn't know that new pr already exists, so it was my fault.

PR #2378, opened 2026-07-15
fix: Added standard OCI image labels (org.opencontainers.image.source

### LINKED ISSUE #513: Add Docker labels

singlaamitesh:
Fixes #513

## Summary
Added standard OCI image labels (org.opencontainers.image.source, .description, .licenses) to the Dockerfile, matching the convention already used by ScancodeIO's Dockerfile for consistency across aboutcode-org tools as requested in issue #513. Verified the Dockerfile parses/builds cleanly through the LABEL instructions (the build later fails at an unrelated pre-existing pip dependency conflict in requirements.txt, unaffected by this change).

## Changes
- `Dockerfile`

## How this was tested
Ran `the affected tests` locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,10 @@
 
 FROM python:3.12
 
+LABEL org.opencontainers.image.source="https://github.com/aboutcode-org/vulnerablecode"
+LABEL org.opencontainers.image.description="VulnerableCode"
+LABEL org.opencontainers.image.licenses="Apache-2.0 AND CC-BY-SA-4.0"
+
 WORKDIR /app
 
 # Python settings: Force unbuffered stdout and stderr (i.e. they are flushed to terminal immediately)


pombredanne:
Thanks! Do you have a link to the upstream documentation for these labels?

singlaamitesh:
Good question! These are the standard OCI (Open Container Initiative) pre-defined annotation keys, documented here: https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

I used `org.opencontainers.image.source`, `.description`, and `.licenses`, matching the same set and convention that ScancodeIO already uses in its Dockerfile (https://github.com/aboutcode-org/scancode.io/blob/main/Dockerfile), so VulnerableCode's image labels stay consistent with our other tooling. Happy to add more of the pre-defined keys (e.g. `.version`, `.title`, `.vendor`) if you'd like broader coverage.

PR #2397, opened 2026-07-28
fix: return 404 for unknown advisory IDs

### LINKED ISSUE #2396: Advisory Page Returns 500 Instead of 404 for Invalid Advisory IDs
Invalid advisory URLs currently return a Server Error (500) instead of a 404 Page Not Found.

Example:

```
https://public.vulnerablecode.io/advisories/pysec/PYSEC-3000-0
```

Expected: 404 Page Not Found

Actual: Server Error (500)

w3lld1:
## Summary

- return no result when an advisory ID does not exist instead of propagating `DoesNotExist`
- add regression coverage for an unknown advisory ID

## Validation

- `python -m pytest -o addopts='' --nomigrations -q vulnerabilities/tests/test_same_avid_different_content_id.py` (7 passed using SQLite)
- `black --check vulnerabilities/models.py vulnerabilities/tests/test_same_avid_different_content_id.py`
- `isort --check-only vulnerabilities/models.py vulnerabilities/tests/test_same_avid_different_content_id.py`
- `python -m compileall -q vulnerabilities/models.py vulnerabilities/tests/test_same_avid_different_content_id.py`
- `git diff --check`

The full PostgreSQL-backed suite is deferred to CI because PostgreSQL is not available in my local environment.

Fixes #2396


ziadhany (approved):
LGTM. Thanks, @w3lld1

Issue #226, opened 2020-07-15
Check types in CI


sbs2001:
Since the codebase mostly uses `typing` the CI should enable a type checking.

Because wrong types are worse than no types at all. 

We could start by simply running a mypy check in the CI.

abhishak3:
I would like to work on this issue.

suraj-mandal:
is this issue open or closed? Possible to work on it?

aditi25srivastava:
Hi! I would like to work on this issue. Please assign it to me.

shivamshrma09:
Hi @sbs2001 

Want to fix 
**#226 (add mypy type checking to CI).** 

Know Python + Docker, local setup ready. 

Can you assign me

AravCS:
Hi, I am working on this issue with @tristonfgibson and @yaeger211202, and we are part of CodeDay for the summer. Can we work on this? Thanks!

PR #2398, opened 2026-07-29
Add mypy type checking to CI (#226) 

AravCS:
Hello me and my team, @yaeger211202 and @tristonfgibson worked on this as part of CodeDay Labs. Here is a summary of our changes: 

Added mypy type checking to CI. Consists of already-clean, high-value modules as the initial whitelist target. 

Rather than running mypy across the whole codebase (pre-existing errors), it uses a whitelist in mypy.ini, only the listed files are checked. This follows [mypy’s existing-codebase docs](https://mypy.readthedocs.io/en/stable/existing_code.html) by starting on a clean subset and expanding over time. Config uses mypy’s recommended foundation settings as well. 

Changes pass both locally and on Github Actions (run on a fork against current main). 

Changes:
- Mypy.ini (config + whitelist)
- Main.yml (added a Check types step running make mypy) 
- Makefile (added a make mypy target) 
- Setup.cfg / requirements-dev.txt (added mypy version 2.1.0)


ziadhany:
Hi @yaeger211202 and @tristonfgibson, I think it may be a good idea to use `Mypy`, but we should discuss it in more detail: what the impact on the CI runtime is, what dependencies it introduces, and whether it should be part of VulnerableCode or all projects.

I think the best way to discuss this in more detail is to create an issue in https://github.com/aboutcode-org/skeleton/. 

If it proves useful, we can integrate it into all of our projects.

AravCS:
Hi @ziadhany, thanks for the reply! As you suggested, I've opened an issue in skeleton: https://github.com/aboutcode-org/skeleton/issues/106

mypy.ini
```diff
@@ -0,0 +1,70 @@
...
+
+files =
+    # fetch and parse vulnerability data from upstream sources
+    vulnerabilities/importers/debian_oval.py,
+    vulnerabilities/importers/epss.py,
+    vulnerabilities/importers/project_kb_msr2019.py,
+    vulnerabilities/importers/redhat.py,
+    vulnerabilities/importers/suse_scores.py,
+    vulnerabilities/importers/ubuntu_usn.py,
+    vulnerabilities/importers/xen.py,
```
ziadhany: I think we should use an exclude list instead. Adding every file individually isn't a good option, especially because whenever we create a new file, it will require manually updating the mypy configuration.

Makefile
```diff
@@ -131,6 +131,10 @@ webtest:
 	@echo "-> Run web tests"
 	${ACTIVATE} ${PYTHON_EXE} -m pytest -vvs -m "webtest"
 
+mypy:
+	@echo "-> Run mypy type checking"
+	@${ACTIVATE} mypy
+
```
ziadhany, line 139: I think `mypy` command should also be part of `check` command

mypy.ini
```diff
@@ -0,0 +1,70 @@
...
+namespace_packages = True
+
+check_untyped_defs = True
+
+warn_unused_configs = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+
+strict_equality = True
+warn_return_any = True
```
ziadhany, line 14: There are a lot of options available. The main question is whether any of them work well with django without requiring a plugin.

ziadhany (changes requested):

mypy.ini
AravCS: done, switched to using an exclude list (mypy uses regex syntax for the exclude list), so now whenever there is a new file created, it is type checked automatically.

Makefile
AravCS: done, added mypy command to the check command as well, as suggested. Also, the make check is commented out currently in main.yml, so I left the standalone mypy command as well.

mypy.ini
AravCS: Hi, the config options alone can’t completely solve this, they control how strict mypy is, but can’t give it knowledge of Django’s ORM. django-stubs works for this and gives more useful type checking, but it requires a mypy plugin. 

At the moment, without django-stubs installed, mypy treats Django as untyped, so it primarily checks regular Python code it can understand, rather than type checking Django-specific APIs. With the ignore_missing_imports = True flag that we included, mypy suppresses errors about missing type information for imports, such as Django.

mypy.ini
```diff
@@ -0,0 +1,205 @@
...
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_apache_httpd_importer_pipeline_v2\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_apache_kafka_importer\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_apache_tomcat_importer_pipeline\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_collect_fix_commit\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_debian_importer\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_elixir_security_importer_v2\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_github_osv_importer_v2\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_gitlab_importer_v2\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_istio_importer_v2\.py$
+    | ^vulnerabilities/tests/pipelines/v2_importers/test_mattermost_importer_v2\.py$
```
ziadhany, line 152: I think it would be simpler to exclude the tests using one or two regex patterns instead of listing them all out individually

AravCS: Hi @ziadhany, I’m happy to make those changes, I just wanted to make sure of the following before. If I make the regex more concise like you mentioned, I would have to exclude the test directories more broadly which would cause us to lose coverage on the test files that are clean. 

Right now, roughly 96 files are clean, and 54 have errors. There doesn't seem to be a simple way to make it just one or two regex patterns while keeping the coverage, because the preexisting typing errors are spread out in the test directory. 

I’m happy to just make it concise though with a broader one line regex, let me know what you prefer.

AravCS:
Hey @ziadhany, just following up on this, did you get a chance to see my comment above about the test exclusions? I wanted to see which one you prefer before making any changes.

mypy.ini
ziadhany: @AravCS I got your point. We are excluding only the 54 test error files, which is fine, I think. No need to change that.

ziadhany:
@AravCS We should also check with @AyanSinhaMahapatra whether we are going to merge this as a separate PR in VulnerableCode, or if it will be part of the skeleton project and get applied later across all the projects we have, and see if there are any blockers.

see:
- https://github.com/aboutcode-org/skeleton/issues/106#issuecomment-5347860564

AravCS:
Hey @ziadhany, thanks for the update, I'll leave the test exclusions as they are then. I'll wait for @AyanSinhaMahapatra to take a look whenever he has time as you said.

PR #2400, opened 2026-07-30
Fix: return 404 instead of 500 for invalid advisory IDs

### LINKED ISSUE #2396: Advisory Page Returns 500 Instead of 404 for Invalid Advisory IDs

gorasatryanGH:
Fixes #2396

## Problem
`AdvisoryV2QuerySet.latest_for_avid()` used `self.get(...)`, which raises
`AdvisoryV2.DoesNotExist` when no matching row exists. The three views that
call it (`AdvisoryDetails`, `AdvisoryPackagesDetails`,
`AdvisoryPackageCommitPatchDetails`) all expect a falsy return value so they
can raise `Http404` themselves — so the exception propagated unhandled and
Django returned a 500 instead of a 404 for any invalid/unknown avid.

## Fix
Catch `DoesNotExist` in `latest_for_avid()` and return `None`, matching what
callers already expect. Query count is unchanged (still 1 query).

## Tests
- Added a queryset-level regression test asserting `latest_for_avid()`
  returns `None` for an unknown avid.
- Added HTTP-level regression tests for all three affected views, asserting
  a 404 response for an invalid avid.
- Verified both new tests fail with the original code (reproducing the
  reported 500) and pass with the fix.
- Ran the wider test suite touching `AdvisoryV2` (models, api_v2, advisory
  merge) — all green.


ziadhany:
Closing this PR as issue #2396 has been fixed by PR #2397.

Thanks!

